PR #3, opened 2015-06-04
Refactor fmt for log

phillbaker:
Swaps for for fatalf, fatalln, printf, etc.

cc @aybabtme @bryanl @macb

### DIFF
--- a/account.go
+++ b/account.go
@@ -1,8 +1,7 @@
 package main
 
 import (
-	"fmt"
-	"os"
+	"log"
 
 	"github.com/digitalocean/doctl/Godeps/_workspace/src/github.com/codegangsta/cli"
 	"github.com/digitalocean/doctl/Godeps/_workspace/src/github.com/digitalocean/godo"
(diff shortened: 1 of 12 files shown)


account.go
```diff
@@ -33,11 +33,10 @@ func accountShow(ctx *cli.Context) {
 	client := godo.NewClient(oauthClient)
 
 	account, _, err := client.Account.Get()
-
 	if err != nil {
-		fmt.Printf("%s\n", err)
-		os.Exit(1)
+		log.Fatalln(err)
```
aybabtme: you don't need the `ln` in `Fatalln`, newlines are inserted automatically

account.go
```diff
@@ -33,11 +33,10 @@ func accountShow(ctx *cli.Context) {
...
 	}
+
 	fmt.Printf(
```
aybabtme: this should be to stderr

action.go
```diff
@@ -38,8 +38,7 @@ var ActionCommand = cli.Command{
 
 func actionShow(ctx *cli.Context) {
 	if len(ctx.Args()) == 0 || len(ctx.Args()) > 1 {
-		fmt.Printf("Error: Must provide exactly one id of an Action to show.\n")
-		os.Exit(64)
+		log.Fatalln("Error: Must provide exactly one id of an Action to show.")
```
aybabtme: same here, don't use `ln`

action.go
```diff
@@ -79,8 +77,7 @@ func actionList(ctx *cli.Context) {
 	actionList, _, err := client.Actions.List(opt)
 
 	if err != nil {
-		fmt.Printf("Unable to list Actions: %s\n", err)
-		os.Exit(1)
+		log.Fatalf("Unable to list Actions: %s\n", err)
```
aybabtme: don't need the `\n`

action.go
```diff
@@ -53,8 +52,7 @@ func actionShow(ctx *cli.Context) {
 	action, _, err := client.Actions.Get(int(id))
 
 	if err != nil {
-		fmt.Printf("%s\n", err)
-		os.Exit(1)
+		log.Fatalln(err)
```
aybabtme: same

aybabtme:
Everywhere:

- [ ] replace the calls to `Funcln` by `Func`
- [ ] remove the trailing `\n` from format strings passed to `log.Funcf`
- [ ] remove all calls to `fmt.Func`
- [ ] carefully define what the output of `stdout` should be for calls, if any. it should be somewhat machine readable and not have ASCII drawings

phillbaker:
Updated.

droplet.go
```diff
@@ -156,7 +156,7 @@ func dropletCreate(ctx *cli.Context) {
 			if sshKey != nil && err == nil {
 				sshKeys = append(sshKeys, godo.DropletCreateSSHKey{ID: sshKey.ID})
 			} else {
-				fmt.Printf("Warning: Could not find key: %s.\n", keyName)
+				log.Fatalf("Warning: Could not find key: %s", keyName)
```
aybabtme: use `%q` instead of `%s` to print keys, so that empty strings are quotted

droplet.go
```diff
@@ -402,8 +383,7 @@ func dropletActionPowercycle(ctx *cli.Context) {
 
 func dropletActionShutdown(ctx *cli.Context) {
 	if ctx.Int("id") == 0 && len(ctx.Args()) != 1 {
-		fmt.Printf("Error: Must provide ID or name for Droplet to destroy.\n")
-		os.Exit(1)
+		log.Fatal("Error: Must provide ID or name for Droplet to issue shutdown.")
```
aybabtme: some messages are punctuated (dot), some aren't. would be good to have them all punctuated

aybabtme:
lgtm minus the two nitpicks

main_test.go
```diff
@@ -54,8 +54,8 @@ func TestGlobalFormatFlag(t *testing.T) {
 	args := []string{"doctl", "-k", "key", "-f", "invalid"}
 	err := app.Run(args)
 
-	expected := "invalid output format: invalid, available output options: json, yaml"
+	expected := "invalid output format: \"invalid\", available output options: json, yaml."
```
aybabtme: raw strings:
```go
`invalid output format: "invalid", available output options: json, yaml.`
```

aybabtme:
:shipit:

PR #6, opened 2015-06-05
Add an 'image' subcommand.

andrewsomething:


andrewsomething:
So as the title suggests, this adds an image subcommand. Though there is some more stuff going on here. The code base is a bit hard to test right now. This also does two things to make that easier.

 * Moves the creation of `godo.Client` out of the individual commands and into `buildApp()` in `main.go` This allows you to override client.BaseURL in one place when mocking the API server.
 
 * Makes `CLIOutput`'s `io.Writer` overridable. This allows us to capture command output when testing.

Let me know if that's the right approach or if there is something I'm missing.

zyegfryed:
What's missing from this PR to get merged?
Can we help on something?
Cheers,

bryanl:

v1 has multiple image commands:

```
~/D/d/p/ha-minio ❯❯❯ doctl compute image                                                                                                                                                ⏎
image commands

Usage:
  doctl compute image [command]

Available Commands:
  list              list images
  list-distribution list distribution images
  list-application  list application images
  list-user         list user images
  get               Get image
  update            Update image
  delete            Delete image

```

PR #8, opened 2015-06-29
Add many Droplet Actions command

zyegfryed:
* disable_backups
* restore
* rebuild
* rename
* change_kernel
* kernels
* enable_ipv6
* enable_private_networking
* snapshot
* upgrade

### DIFF
--- a/droplet.go
+++ b/droplet.go
@@ -64,6 +64,14 @@ var DropletCommand = cli.Command{
 		},
 		// Droplet Actions
 		{
+			Name:   "disable_backups",
+			Usage:  "[--id | <name>] Disable backups on a Droplet.",
+			Action: dropletActionDisableBackups,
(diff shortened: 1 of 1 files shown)


tlk:
Nice addition. Input to the PR:
- log.Fatal error messages contain "to resize" in dropletAction functions that do not resize
- dropletAction functions with lots of duplicate code (could be refactored in a separate commit)

zyegfryed:
Hi @tlk 
I've fixed the first bullet point. For the second one, there's already PR#6 that comes with a solution. So, let's wait a see whether it gets merged, otherwise, feel free to open an new PR to bring this change.

tlk:
Hi @zyegfryed That's great! Let's wait and see what happens with these PR's.

bryanl:

PR #10, opened 2015-08-08
Don't require API key if no args were passed

kitschysynq:
Typing `doctl` on its own should not produce an error when `doctl
--help` would not.


mdlayher:
:+1:

PR #11, opened 2015-08-20
Invalid type on --add-region parameter

gen0cide-:
This was automatically appending the region on the end of the droplet. The logic was broken due to a type mismatch between the arg definition and the check.


kitschysynq:
Beat me to it! :+1:

PR #12, opened 2015-09-14
added ssh support

thebyrd:
This changes the behavior of `doctl ssh` to ssh into a droplet instead of manipulating ssh keys. Those commands I moved to `doctl keys`.


bryanl:

PR #13, opened 2015-10-09
Update README.md

FreddieV4:
Add link to `LICENSE.txt`


bryanl:

PR #15, opened 2016-03-23
Merging doit into doctl

bryanl:
These commits effectively replace doctl with doit. doit will be renamed doctl.


aybabtme:
:shipit: :tada:

PR #23, opened 2016-03-30
Add installation instructions from community tutorial

brennen:
Copies in the installation section from the community tutorial, per a few users who have had some difficulty getting the binary into a `$PATH`-included location.


slobatch:
@brennen Is it worth pointing out that OS X users should use the Darwin binary?
(Not that I would ever make that mistake or anything. 😛 )

PR #35, opened 2016-04-04
Add explicit parent to SSH command

### LINKED ISSUE #30: --ssh-port option ignored
My droplets use a non-standard port for ssh. So I tried this:
`doctl compute ssh <my_droplet> --ssh-user <my_user> --ssh-port 1234`

But I get `ssh: connect to host <droplet_ip> port 22: Connection refused`
`Error: 255`

Software | Version
--- | ---
`doctl` | v1.0.0
OS | OSX El Capitan 10.11.4 (15E65)
Shell | iTerm 2 (2.1.4)

Btw, thank you all for this awesome tool!!

bryanl:
Add explicit parent to SSH command, so it can create configuration keys in the proper place. This solves the problem of ssh flags not working since they were be being searched for in `doit.ssh.*` instead of `compute.ssh.*`.

Fixes #30

### DIFF
--- a/pkg/runner/mocks/Runner.go
+++ b/pkg/runner/mocks/Runner.go
@@ -1,0 +1,21 @@
+package mocks
+
+import "github.com/stretchr/testify/mock"
+
+type Runner struct {
+	mock.Mock
+}
+
+// Run provides a mock function with given fields:
+func (_m *Runner) Run() error {
+	ret := _m.Called()
+
+	var r0 error
+	if rf, ok := ret.Get(0).(func() error); ok {
+		r0 = rf()
+	} else {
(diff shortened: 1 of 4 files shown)


pkg/runner/mocks/Runner.go
```diff
@@ -0,0 +1,21 @@
+package mocks
+
+import "github.com/stretchr/testify/mock"
+
+type Runner struct {
+	mock.Mock
+}
+
+// Run provides a mock function with given fields:
+func (_m *Runner) Run() error {
```
aybabtme: what's up with the underscore?

aybabtme:
:shipit: !

PR #38, opened 2016-04-05
Issues with listing droplet kernels

### LINKED ISSUE #33: Multiple Missing Things for 'change-kernel' Action
The `--format` option is missing for the `change-kernel` action using the following command:

```
doctl compute droplet-action change-kernel <droplet-id>
```

This option exists for other similar `droplet-action` such as `power-off`.

I was updating my mass kernel updater script to use `doctl` instead of `curl` and came across this issue.

EDIT:
It also appears the `--no-header` option is missing too.

EDIT 2:
It appears that it does not list all kernels too. My assumption would be the paging in the DigitalOcean API.

bryanl:
* Kernel lists had no formating or header options.
* Pagination skipped the second page of results.

Fixes #33

### DIFF
--- a/do/pagination_test.go
+++ b/do/pagination_test.go
@@ -1,4 +1,3 @@
-
 /*
 Copyright 2016 The Doctl Authors All rights reserved.
 Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


do/pagination_test.go
```diff
@@ -14,7 +13,74 @@ limitations under the License.
...
+		{
+			r: &godo.Response{
+				Links: &godo.Links{
+					Pages: &godo.Pages{Last: "http://example.com/?page=1"},
+				},
+			},
+			lastPage: 1,
+			isValid:  true,
+		},
+		{
```
aybabtme: maybe add just 1 more page to test the 2nd page issue

aybabtme:
:ship:

do/pagination_test.go
bryanl: The `Test_PaginateResp` test covers this regression.

Issue #22, opened 2016-03-30
doctl 1.0.0 version shows as 1.0.0-dev


suizman:
$ doctl version
doctl version 1.0.0-dev
Git commit hash: d8e2e0b

$ sha256sum /usr/bin/doctl
9f8ebe3c1d69ac5cb4ce313d82ceeac83485a57904ea9ef11accc62db8944e7b)  /usr/bin/doctl

![doctl](https://cloud.githubusercontent.com/assets/12419155/14155770/6fb9a724-f6c3-11e5-848d-82915a3b1adb.png)

PR #39, opened 2016-04-05
Remove dev label from release builds

bryanl:
Release builds will now have a version string that looks like: `doctl version 1.0.0-release`

Fixes: #22


nickvanw:
LGTM

PR #49, opened 2016-04-12
use image instead of image id when rebuilding droplets

bryanl:
Use image instead of image id when rebuilding droplets. The argument was tagged with `ArgImage` when it should have been tagged with `ArgImage`. Since doctl has to be GIGO, it was happy to pass the blank string upstring to the API.


bryanl:
@aybabtme

nickvanw:
LGTM

PR #54, opened 2016-04-19
rename the root package to match import path

aybabtme:
some left over. there's more but this was the most annoying =P 

r: @bryanl


bryanl:
:+1:

PR #57, opened 2016-04-20
make new-version message appeal to action

aybabtme:
The current message reads kind of weird. When I saw it, I found
it a bit confusing, because it just states "x is greater than y".
It took me a couple seconds parse the information and understand
"oh you mean there's a newer version and I should update?".
```
doctl version 1.0.0-dev
"1.0.2" is a newer release than "1.0.0-dev"
```

Since the current version is already printed, we don't need to
repeat "this is version x, version y is greater than version x".
We can just say "this is version x, version y is available!".

Then we can make the message a bit less impersonal, and try to encourage
users to perform an action "hey this version is avail, try it!".

So it ends up like this:
```
doctl version 1.0.0-dev
release 1.0.2 is available, check it out!
```

r: @bryanl

### DIFF
--- a/doit.go
+++ b/doit.go
@@ -105,7 +105,7 @@ func (v Version) Complete(lv LatestVersioner) string {
 		v1, err2 := semver.Make(v.String())
 
 		if err1 == nil && err2 == nil && v0.GT(v1) {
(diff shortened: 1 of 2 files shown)


bryanl:
:+1:

PR #58, opened 2016-04-21
add beta features

bryanl:
Beta features use a the `beta` command config option, and
are enabled with the an environment variable `DIGITALOCEAN_ENABLE_BETA`
set to a `true` value (e.g. 1, true)

Signed-off-by: bryanl <bryanliles@gmail.com>

/c @aybabtme


commands/doit.go
```diff
@@ -51,36 +53,59 @@ var Writer = os.Stdout
...
+	viper.SetDefault("output", "text")
+
 }
 
-// Init initializes the root command.
-func Init() *Command {
-	initializeConfig()
-	addCommands()
+func homeDir() string {
+	return os.Getenv("PATH")
```
aybabtme: should be `HOME`? `PATH` is a colon separated string of path where binaries reside, not what you want I think

aybabtme:
minus the `PATH` thing, LGTM

PR #68, opened 2016-05-04
ssh: respect ssh-user flag

vkurchatkin:
Currently `ssh-user` is always overwritten by user extracted from droplet id. I'm not sure even that there are cases where droplet id contains user, so it's almost always an empty string.


bryanl:
Would you mind add a test for this regression?

vkurchatkin:
@bryanl done!

Issue #73, opened 2016-05-06
Cannot execute binary file when following instructions in Readme.md


jaredcaputo:
Hi, I get the error "-bash: /usr/local/bin/doctl: cannot execute binary file" when following the instructions in the readme file.

aybabtme:
Hi @jaredcaputo !  What OS are you using?

jaredcaputo:
Mac OS X 10.11.4

aybabtme:
@jaredcaputo can you list out exactly what commands you performed?

jaredcaputo:
cd ~

curl -OL
https://github.com/digitalocean/doctl/releases/download/v1.0.0/doctl-1.0.0-linux-amd64.tar.gz

tar xf doctl-1.0.0-linux-amd64.tar.gz

sudo mv ./doctl /usr/local/bin

doctl auth login

Then, on the final one, I get: -bash: /usr/local/bin/doctl: cannot execute
binary file

On Fri, May 6, 2016 at 7:54 PM Antoine Grondin <notifications@github.com>
wrote:

> @jaredcaputo <https://github.com/jaredcaputo> can you list out exactly
> what commands you performed?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/digitalocean/doctl/issues/73#issuecomment-217589606>
>
-- 
Jared Caputo
Silver Palm Productions, LLC
(941) 404-8145
Jared@silverpalmproductions.com

jaredcaputo:
Also tried on another Mac running the same OS, didn't work.

aybabtme:
thanks @jaredcaputo, the problem is that our readme tells people to download the Linux executable, which doesn't work on OS X. Try installing with `brew install doctl`, or this link if you want to download manually: 

```
https://github.com/digitalocean/doctl/releases/download/v1.1.0/doctl-1.1.0-darwin-10.6-amd64.tar.gz
```

jaredcaputo:
Works! Thanks very much!!

Aprea123:
I missing myself inside of the com.apple /PrivateFrameworks and you find several times "cannot execute binary file".  Just finish now to install doctl. Thank you very much.

PR #74, opened 2016-05-07
make install instructions more specific

aybabtme:
Current instructions are not as explicit as they could be. In particular, the links provided in the code blocks are for Linux only and the text doesn't capture attention enough to remind users to change the string.

Also, the distinction between `wget` and `curl` is not necessary on OS X, since `curl` ships by default.

1. highlight the preferred method (package managers)
2. fence install methods by OS
3. make the code block clearly show what link to use
4. skip the "and then you can untar" part, which isn't necessary/adds noise.

r: @bryanl

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,46 +25,49 @@ Use "doctl [command] --help" for more information about a command.
 
 ## Installation
 
-### Option 1 - Use a Package Manager
+### Option 1 - Use a Package Manager (preferred method)
 
-Currently, you can use [Homebrew](http://brew.sh) to install `doctl` on Mac OS X by using the command below:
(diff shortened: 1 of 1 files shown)


aybabtme:
this is motivated by https://github.com/digitalocean/doctl/issues/73

bryanl:
There are some links to the 1.0.0 and some links to 1.1.0, I believe they should all be updated to 1.1.0.

aybabtme:
@bryanl I believe you to be right!

aybabtme:
@bryanl PTAL

bryanl:
:+1:

Issue #78, opened 2016-05-22
droplet list for larger number of droplets contains duplicates


jphines:
![image](https://cloud.githubusercontent.com/assets/1100763/15457019/14e065c6-204f-11e6-8e6f-284abf8bd402.png)

Intuitively, I would guess there is a bug in the way pagination is expanded.

PR #79, opened 2016-05-23
doctl: pagination is 1 indexed, not 0 indexed

jphines:
I think this is the bug, but I'll let CI run the tests.

Essentially pagination is 1 indexed, and not 0 indexed. So we re-fetch the first page, but never fetch the second, getting a "second page" full of dupes.

### DIFF
--- a/do/pagination.go
+++ b/do/pagination.go
@@ -52,7 +52,7 @@ func PaginateResp(gen Generator) ([]interface{}, error) {
 	fetchChan := make(chan int, maxFetchPages)
 
 	var wg sync.WaitGroup
-	for i := 0; i < 4; i++ {
+	for i := 0; i < maxFetchPages-1; i++ {
 		wg.Add(1)
 		go func() {
 			for page := range fetchChan {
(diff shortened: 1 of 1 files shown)


jphines:
Fixes https://github.com/digitalocean/doctl/issues/78

aybabtme:
this `i := 1` and `i := 2` stuff feels fragile to me, I feel there's got to be a way to do this with `resp.Links`

jphines:
@aybabtme clearly it is fragile because its broken

do/pagination.go
```diff
@@ -80,7 +80,7 @@ func PaginateResp(gen Generator) ([]interface{}, error) {
 	}
 
 	// start with second page
-	for i := 1; i < lp; i++ {
+	for i := 2; i <= lp; i++ {
```
aybabtme: can you increment `opt.Page++` instead? feels more obvious to me, since the `:=2` here really just means "the page after what we just fetched"

jphines:
image

aybabtme:
🚢

PR #85, opened 2016-06-08
Don't report new release available when already installed.

andrewsomething:
`doctl` reports that there is a new version available even though the most recent release is installed:

```
$ doctl version
doctl version 1.2.0-release
Git commit hash: 4d9207c
release 1.2.0 is available, check it out!
```

This is because the release process adds the `release` label while the git tag for the release does not contain that. See:

https://github.com/digitalocean/doctl/blob/master/scripts/stage.sh#L34

`github.com/blang/semver` is used to compare the versions and `1.2.0` is considered a higher version than `1.2.0-release`. You can test this with:

```go
package main

import (
	"fmt"
	"github.com/blang/semver"
)

func main() {
	v0, err1 := semver.Make("1.0.0")
	v1, err2 := semver.Make("1.0.0-release")

	if err1 == nil && err2 == nil && v0.GT(v1) == true {
		fmt.Printf("%q gt %q", v0.String(), v1.String())
	}
}
```

This PR compensates for that by internally adding the `release` label to the tag if no other is already assigned. This ensures things still correctly function for releases labeled `beta`, `dev`, etc... 

Of course, the alternative is to tag the release in git differently...

### DIFF
--- a/doit.go
+++ b/doit.go
@@ -104,6 +104,10 @@ func (v Version) Complete(lv LatestVersioner) string {
 		v0, err1 := semver.Make(tagName)
 		v1, err2 := semver.Make(v.String())
 
+		if len(v0.Build) == 0 {
+			v0, err1 = semver.Make(tagName + "-release")
+		}
+
 		if err1 == nil && err2 == nil && v0.GT(v1) {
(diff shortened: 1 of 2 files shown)


aybabtme:
nice, thanks @andrewsomething !

Issue #77, opened 2016-05-19
SSH Agent Forwarding


sebgie:
First and foremost, this tool is great and it makes my work a lot easier! Thank you!

One of the things I use quite often is SSH agent forwarding (`ssh -A user@server`) for example for deployment tasks. I wonder if it would be possible to add this functionality to the `doctl compute ssh <droplet-name>` command?

bryanl:
I concur. This is an idea on the list. I want to make this available sooner than later. I'll make sure to update this issue with the status.

bryanl:
Support for ssh agent forwarding is now in master. A new release including this support will be shipped soon.

Issue #65, opened 2016-04-29
Windows SSH client agent support


bryanl:
Doctl uses a custom ssh client because it can't assume there is a ssh executable in the path. In order to be more friendly, there should exist support for SSH agents.

xmudrii:
Is this still a problem on Windows?

What about:
https://godoc.org/golang.org/x/crypto/ssh#Permissions
It talks about `permit-ssh-agent`.
But also more important - https://godoc.org/golang.org/x/crypto/ssh/agent

Is it possible to fix issue with this, is this still a issue?

cc @aybabtme @nanzhong

xmudrii:
By looking more into issue, it looks like it's already implemented/fixed long time ago if I'm thinking of right thing.
On [lines 77-85](https://github.com/digitalocean/doctl/blob/master/pkg/ssh/ssh_internal.go#L77-L85), as well as [lines 188-194](https://github.com/digitalocean/doctl/blob/master/pkg/ssh/ssh_internal.go#L188-L194) of [`pkg/ssh/ssh_internal.go`](https://github.com/digitalocean/doctl/blob/master/pkg/ssh/ssh_internal.go), you can see Agent Forwarding code.

Commit ec6d5c2 implemented it.

If this is right thing, this issue could be closed. If not, provide more context on what is needed if it's possible.

cc @viola @caueguerra @mauricio

mauricio:
Yup, this has been sorted out already.

PR #88, opened 2016-06-15
Add a --ssh-agent-forwarding flag

tbalthazar:
It doesn't work on Windows, since the custom ssh client for Windows
doesn't have agent support yet, as stated in issue #65.

This feature has been requested in issue #77.

That's my first contribution to doctl, comments are welcome!

### DIFF
--- a/commands/commands_test.go
+++ b/commands/commands_test.go
@@ -23,6 +23,7 @@ import (
 	"github.com/digitalocean/doctl/do"
 	domocks "github.com/digitalocean/doctl/do/mocks"
 	"github.com/digitalocean/doctl/pkg/runner"
+	"github.com/digitalocean/doctl/pkg/ssh"
 	"github.com/digitalocean/godo"
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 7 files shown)


bryanl:
Hello,

Thanks for the suggestion. SSH agent forwarding is a needed feature. I have a few questions about the implementation, and I'll add those inline.

commands/commands_test.go
```diff
@@ -196,15 +196,15 @@ func withTestClient(t *testing.T, tFn testFn) {
 }
 
 type TestConfig struct {
-	SSHFn func(user, host, keyPath string, port int) runner.Runner
+	SSHFn func(user, host, keyPath string, port int, agentForwarding bool) runner.Runner
```
bryanl: I'm thinking this won't be the last option for SSH, so shouldn't we pass this option in a more forward-thinking way?

```go
type SSHOptions map[string]interface{}

func(user, host, keyPath string, port int, options SSHOptions) runner.Runner {}
```

This would allow you do something like:

```go
opts[doctl.ArgsSSHAgentForwarding] = true
```

Not sure what this means long term, but it would solve the immediate issue of longer function signatures.

tbalthazar:
Hi Bryan,

Thanks for the suggestion, it's definitely an improvement!

Cheers,
Thomas.

PR #91, opened 2016-06-23
fix call to t.Fatal where test goroutine is gone

aybabtme:
Can't call `t.Fatal` on a `t` who's goroutine has returned. At least in Go 1.7, this will panic.

Also this test can't pass since `stdout` isn't a terminal.


aybabtme:
redundant with #93

PR #98, opened 2016-06-24
Remove output when deleting droplets

### LINKED ISSUE #55: if json output is selected should only produce json output
when JSON output is selected, don't `fmt.Print...` stuff to stdout.

e.g: https://github.com/digitalocean/doctl/blob/master/commands/droplets.go#L408

bryanl:
Fixes #55


aybabtme:
:shipit:

PR #99, opened 2016-07-03
build test bin in out directory

bryanl:
Reduces clutter in root directory by building test bin in subdirectory


aybabtme:
:shipit:

PR #103, opened 2016-07-12
display public ipv6 in droplet list

utlemming:
It would be really, really nice if we could have the PublicIPv6 displayed too. 

Signed-off-by: Ben Howard <bh@digitalocean.com>


bryanl:
I haven't forgotten about this. A PR with the ability to include arbitrary items is incoming in the very near future.

PR #111, opened 2016-07-23
Introduce `doctl auth init`

### LINKED ISSUE #59: [Suggestion] XDG conformance for .doctlcfg
This would be a nice to have feature for those of us with OCD.  :)

### LINKED ISSUE #94: Replace `doctl auth login` with `doctl auth init`
The `doctl auth login` process has a few issues that doesn't allow it to work the same way on every platform. I'm proposing we replace this with a `doctl auth init` that will take a DO token on the command line and do the initial configuration which will include creating a doctl config file.

bryanl:
Introduces `dolt auth init` command which replaces `dolt auth login`.

Instead of trying to fetch the token from the API, the user will have to manually enter it instead.

Also moves the default configuration location to `$XDG_CONFIG_HOME/doctl/doctlcfg`. On Windows, the default location is `%APPDATA%/doctl/config` and on other platform without the environment variable set, it will look in `$HOME/.config/doctl`. 

If the legacy location is detected, a warning will be displayed.

Fixes #59 
Fixes #94

### DIFF
--- a/commands/auth.go
+++ b/commands/auth.go
@@ -15,21 +15,13 @@ package commands
 
 import (
 	"bufio"
-	"encoding/json"
 	"errors"
 	"fmt"
-	"io/ioutil"
-	"net/http"
-	"net/url"
 	"os"
-	"runtime"
+	"strings"
 
 	"golang.org/x/crypto/ssh/terminal"
 
-	"github.com/bryanl/doit-server"
-	"github.com/bryanl/webbrowser"
-	"github.com/gorilla/websocket"
-	"github.com/satori/go.uuid"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
@@ -46,7 +38,7 @@ var retrieveUserTokenFunc = func() (string, error) {
 	}
 
 	reader := bufio.NewReader(os.Stdin)
-	fmt.Print("Enter token: ")
+	fmt.Print("DigitalOcean access token: ")
 	return reader.ReadString('\n')
 }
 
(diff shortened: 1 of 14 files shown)


commands/auth.go
```diff
@@ -71,187 +63,39 @@ func Auth() *Command {
...
-	return &m, nil
-}
-
-type keyPair struct {
-	k, v string
-}
-
-func (dsa *doitServerAuth) createAuthURL(ac *doitserver.AuthCredentials, kps ...keyPair) (string, error) {
-	authURL, err := url.Parse(dsa.url)
+// RunAuthInit initializes the doctl config. Configuring is stored in $XDG_CONFIG_HOME/doctl. On Unix, if
```
aybabtme: `s/Configuring/Configuration/`

commands/auth.go
```diff
@@ -71,187 +63,39 @@ func Auth() *Command {
...
-
-	authURL.Path = "/auth/digitalocean"
-
-	q := authURL.Query()
-	q.Set("id", ac.ID)
-	q.Set("cs", ac.CS)
-
-	for _, kp := range kps {
-		q.Set(kp.k, kp.v)
+		return errors.New("unable to read input")
```
aybabtme: should include the error, and this scope doesn't know about _user input_, it knows about _user token_ so maybe the message should be like:
```go
return fmt.Errorf("unable to obtain user token: %v", err)
```

commands/auth.go
```diff
@@ -71,187 +63,39 @@ func Auth() *Command {
...
-	case "http":
-		u.Scheme = "ws"
-	case "https":
-		u.Scheme = "wss"
-	default:
-		return nil, &UnknownSchemeError{Scheme: u.Scheme}
+	if _, err := c.Account().Get(); err != nil {
+		fmt.Fprintln(c.Out, "invalid token")
+		fmt.Fprintln(c.Out)
+		return errors.New("unable to use supplied token to access API")
```
aybabtme: missing the error here too

commands/auth_test.go
```diff
@@ -14,228 +14,49 @@ limitations under the License.
...
+		cfgFileWriter = cfw
 	}()
 
 	retrieveUserTokenFunc = func() (string, error) {
-		return "token", nil
+		return "valid-token", nil
 	}
 
-	dsa := newDoitServerAuth()
+	cfgFileWriter = func() (io.WriteCloser, error) { return &discardCloser{}, nil }
```
aybabtme: `ioutil.NopCloser(ioutil.Discard)` =]

bryanl: Guess I didn't look hard enough. Felt weird putting that in.

commands/doit.go
```diff
@@ -93,7 +103,26 @@ func initConfig() {
...
 
 	viper.SetDefault("output", "text")
+}
+
+func findConfig() (string, error) {
+	if cfgFile != "" {
+		return cfgFile, nil
+	}
+
+	legacyConfigPath := filepath.Join(homeDir(), ".doctlcfg")
```
aybabtme: since this is in `.config/doctl`, any reason not to just call this `config.json`?

```
$ tree ~/.config
/Users/antoine/.config
├── fish
     └── config.fish
└── gocode
     └── config.json
...
```

bryanl: It's yaml, but an extension could be helpful

commands/doit.go
```diff
@@ -343,4 +392,35 @@ func CmdBuilder(parent *Command, cr CmdRunner, cliText, desc string, out io.Writ
...
+	_, err = f.Write(b)
+	if err != nil {
+		return errors.New("unable to write configuration")
+	}
+
+	return nil
+}
+
+func defaultConfigFileWriter() (io.WriteCloser, error) {
+	f, err := os.Create(cfgFile)
```
aybabtme: since this contains API tokens, you might want to reduce the permissions to something like `0640`

commands/doit.go
aybabtme: ah, sorry, but the idea is to call it `.config/doctl/config.extension`

commands/auth_test.go
bryanl: actually, i was right. this is how you would do it. nopCloser wants an `io.Reader`

bryanl:
Looks like the windows paths aren't quite right.  They should be `C:\Users\%USERNAME%\AppData\Local\doctl\config`. Will update this PR.

bryanl:
Chmod isn't supported on windows. Don't do that check on that platform.

bryanl:
@aybabtme: ^^

commands/doit.go
```diff
@@ -56,7 +59,11 @@ var Trace bool
 // cfgFile is the location of the config file
 var cfgFile string
 
-var cfgFileName = ".doctlcfg"
+// cfgFileWriter is the config file writer
+var cfgFileWriter = defaultConfigFileWriter
+
+// ErrNoAccesToken is an error for when there is no access token.
+var ErrNoAccesToken = errors.New("no access token has been configured")
```
aybabtme: `s/ErrNoAccesToken/ErrNoAccessToken/ `

commands/doit.go
```diff
@@ -93,7 +105,26 @@ func initConfig() {
...
+			legacyConfigPath, configHome())
+		warn(msg)
+	}
+
+	ch := configHome()
+	if err := os.MkdirAll(ch, 0755); err != nil {
+		return "", err
+	}
+
+	return filepath.Join(ch, "doctlcfg.yaml"), nil
```
aybabtme: so hum why not `config.yaml` in the end?

bryanl: no reason at all. It just carried over from the original name.

aybabtme: I think `config.yaml` would be a more conventional name, given that other users of the XDG locations call their config's `myapp/config.ext`

aybabtme:
:shipit: at will

PR #113, opened 2016-07-24
Fix SSH client on Windows

### LINKED ISSUE #112: SSH Client doesn't work on Windows
It happens on current master (a57555c), when trying to connect to a droplet from Windows 10 in VirtualBox.

There are 2 different problems:

**1/ When trying to connect with a SSH key that isn't password protected, the error is**:
```
doctl.exe compute ssh playground --ssh-key-path D:\ssh-tmp-keys\id_rsa_without_password
Error: read /dev/stdout: The handle is invalid.
```

The problem is that [this line](https://github.com/digitalocean/doctl/blob/a57555c195d06bc7aa5037af77fde0665ad1231f/pkg/ssh/ssh_internal.go#L48) is using `os.Stdout` while it should be using `os.Stdin`.

There's also a logical error [here](https://github.com/digitalocean/doctl/blob/a57555c195d06bc7aa5037af77fde0665ad1231f/pkg/ssh/ssh.go#L83), we check for the _function scoped_ `err` before starting the shell,  but it can be _non-nil_ because of [this line](https://github.com/digitalocean/doctl/blob/a57555c195d06bc7aa5037af77fde0665ad1231f/pkg/ssh/ssh.go#L67).

**2/ When trying to connect with a SSH key that is password protected, the error is**:
```
doctl.exe compute ssh playground --ssh-key-path D:\ssh-tmp-keys\id_rsa_with_password
Error: asn1: structure error: tags don't match (16 vs {class:1 tag:6 length:50 isCompound:false}) {optional:false explicit:false application:false defaultValue:<nil> tag:<nil> stringType:0 timeType:0 set:false omitEmpty:false} pkcs1PrivateKey @2
```

The problem is that `ssh.ParsePrivateKey(key)` on [this line](https://github.com/digitalocean/doctl/blob/a57555c195d06bc7aa5037af77fde0665ad1231f/pkg/ssh/ssh_internal.go#L33) expects a [PEM encoded private key](https://godoc.org/golang.org/x/crypto/ssh#ParsePrivateKey) so we have to decrypt it first.

The SSH keys used to reproduce this issue have been generated using:
`ssh-keygen -t rsa -b 4096 -C "t@example.com"`

One key has a passphrase, the other doesn't. They are attached here:
[ssh-test-keys.tar.gz](https://github.com/digitalocean/doctl/files/380347/ssh-test-keys.tar.gz)

tbalthazar:
Use os.Stdin when putting the terminal into raw mode and creating a new terminal.
Decrypt SSH key if needed.

Fixes #112


bryanl:
Hello, thanks for the PR! It will be reviewed once we get the new auth init code integrated.

PR #117, opened 2016-07-26
Fix download links for OSX and Linux in README

garykrige:
The links point to a non-existent version.


bryanl:
This whole section needs to be revamped as Markdown and Github doesn't give us an automatic way to point to these. I think we can do better, but I'm not quite sure what it is yet. Stay tuned.

garykrige:
Is it not worth having the correct links in the meantime?

Issue #119, opened 2016-07-26
"doctl compute droplet untag" - droplet not found bug


snoopdouglas:
I'm trying to remove a tag from a droplet, and so far have tried all the following commands:

```
doctl compute droplet untag dropletname tagname
doctl compute droplet untag tagname dropletname
doctl compute droplet untag tagname --droplet-name dropletname
```

Each one gives these errors respectively:

```
Error: (droplet.untag) command is missing required arguments
Error: (droplet.untag) command is missing required arguments
Error: droplet with name "[dropletname]" could not be found
```

Am I doing something wrong or is this a bug? The droplet and tag definitely exist.

In addition, the usage statement for this is rather unclear:
```
Usage:
  doctl compute droplet untag <droplet id or name> [flags]

Flags:
      --droplet-name value   Droplet names (default [])
```

Version: 1.3.1

bryanl:
Confirmed, this is a bug.

bryanl:
It looks like we are shaping up to do a release hopefully by the end of the week. I expect a fix in master earlier than that, however.

bryanl:
As for the statement, I agree, `doctl compute droplet untag <tagname> --droplet-name` is backwards.  Perhaps it should be `doctl compute droplet untag <droplet names or ids> --tag-name <foo> --tag-name <bar>`

snoopdouglas:
This'd indeed be good IMO:

`doctl compute droplet untag <droplet names or ids> --tag-name <foo> --tag-name <bar>`

...because then it matches the syntax of `compute droplet tag`. :)

snoopdouglas:
@bryanl do you reckon this will be merged before 1.4.1?

bryanl:
Yes, it'll be merged today or tomorrow. And I think we'll do a 1.4.1 early next week.

snoopdouglas:
Excellent, cheers.

PR #121, opened 2016-07-27
Allow untagging droplets

bryanl:
* Change command for untagging to `doctl compute untag <droplet id or
name> [droplet id or name] —tag-name foo`
* Catch broken pflags earlier in the chain, so string slice fetching
works sanely

fixes: #119


aybabtme:
:shipit:

PR #122, opened 2016-07-28
Add SSH agent forwarding support to windows

tbalthazar:
Minor refactoring of SSH support for windows.
Move the sshConnect function from ssh.go to ssh_internal.go (makes more
sense IMHO).
Refactor tests for ssh_internal.go.


tbalthazar:
I was wondering if there was a reason to use a copy of the `term` package [in the pkg/term](https://github.com/digitalocean/doctl/tree/master/pkg/term) directory or if it was just legacy code. I have a commit ready that removes it and replaces its usage ([only 3 times here](https://github.com/digitalocean/doctl/blob/master/pkg/ssh/ssh.go#L59)) by the equivalent functions from the `golang.org/x/crypto/ssh/terminal` package.

Please let me know if that makes senses.
Thanks!

tbalthazar:
@bryanl Done :)

bryanl:
@tbalthazar I was manually running through this, and hit a snag. What if you have a droplet that wasn't created with a pub key, and you need to enter a password? Right now, this will error if there isn't a private key.

tbalthazar:
@bryanl Good catch! The _internal_ SSH client now mimics the `ssh` command on Linux: it warns the user and asks for password.
ssh-on-windows

aybabtme:
@tbalthazar 

> I have a commit ready that removes it and replaces its usage (only 3 times here) by the equivalent functions from the golang.org/x/crypto/ssh/terminal package.

The author of that package don't really like that people be using it, it was created before the introduction of the `internal/` mechanism, the authors have mentioned regret at not having put the `internal/` label on it. For those reasons, I recommend against using it. There's work elsewhere in the stdblib + exp trees to bring proper term support to Go.

tbalthazar:
@aybabtme I ignored that, thanks for rising the issue. I did some research to try to get the context of this and found [this discussion](https://github.com/golang/go/issues/13104) where [mattn says](https://github.com/golang/go/issues/13104#issuecomment-156182504) that it should be an internal package. Is that what you are referring to? My understanding of this discussion is that the issue seems to be _"where to put/how to best name that package?"_. It seems they agreed to move it to `x/term` in November 2015. Then I found [another issue (July 2016)](https://github.com/golang/go/issues/16420) where the discussion is about removing the `x/term` package because it's empty and nobody took the time to move the terminal related code there. The `x/crypto/ssh/terminal` package [seems to still be maintained](https://github.com/golang/crypto/commits/master/ssh/terminal) though.
So, my understanding of the issue you raised is that terminal related code might be moved to another package at some point. But since the package is vendored in the `doctl` project, I guess it shouldn't be a problem to keep on using it. Please let me know what you think about that and tell me if I'm missing something. Thanks!

bryanl:
:+1: Going to rebase this, and merge it in. Thanks for your contribution @tbalthazar!

aybabtme:
@tbalthazar sgtm!

PR #123, opened 2016-07-28
Use window compatible ansi colors

### LINKED ISSUE #114: color support does not work in Windows
Color support isn't working in windows.

bryanl:
Current color outputs don’t work well with windows.

Fixes #114

### DIFF
--- a/commands/errors.go
+++ b/commands/errors.go
@@ -19,13 +19,14 @@ import (
 	"os"
 
 	"github.com/fatih/color"
+	"github.com/shiena/ansicolor"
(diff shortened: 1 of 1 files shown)


aybabtme:
any way to test that this works? otherwise non-windows devs may silently break this inadvertently

bryanl:
I built it on and ran it in macOS, and it works right. The change just adds what github.com/fatih/color does by default to `color.Output`. When I moved output to stderr, i forgot to wrap it.

aybabtme:
kk :shipit:

Issue #131, opened 2016-08-23
[Bug] [Auth] [Linux-Ubuntu] 'doctl auth init' command not found on new installation of doctl


xMudrii:
# Update:
Problem is in README.md, links given in are for old version, I found on release page new version with 'doctl auth init'. Links could be fixed in both, README.md and [DigitalOcean Community Tutorial](https://www.digitalocean.com/community/tutorials/how-to-use-doctl-the-official-digitalocean-command-line-client).

## Steps to reproduce problem :
* Execute following commands to install `doctl` :
 * `cd ~`
 * `curl -L https://github.com/digitalocean/doctl/releases/download/v1.3.1/doctl-1.3.1-linux-amd64.tar.gz  | tar xz`
 * `sudo mv ./doctl /usr/local/bin`
 * `doctl` *- to verify is doctl available (yes)*
* Execute `doctl auth init`

## Console output of `doctl auth init`
```
auth is used to access auth commands

Usage:
  doctl auth [command]

Available Commands:
  login       login to DigitalOcean account

Global Flags:
  -t, --access-token string   API V2 Access Token
  -c, --config string         config file (default is $HOME/.doctlcfg)
  -o, --output string         output format [text|json] (default "text")
      --trace                 trace api access
  -v, --verbose               verbose output

Use "doctl auth [command] --help" for more information about a command.
```

## Console output of `doctl auth login`
```
login to DigitalOcean account

Usage:
  doctl auth login [flags]

Flags:
  -h, --help   help for login

Global Flags:
  -t, --access-token string   API V2 Access Token
  -c, --config string         config file (default is $HOME/.doctlcfg)
  -o, --output string         output format [text|json] (default "text")
      --trace                 trace api access
  -v, --verbose               verbose output

Error: access token is required
```

### Side notes:
* I tried executing doctl as sudo.
* I tried executing dcotl as root user (`sudo su`).
* I tried rebooting my workstation after installing `doctl`.

## Workstation :
OS: **Ubuntu Desktop 16.04.1 x64**
### Console output of `lsb_release -a`
```
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.1 LTS
Release:	16.04
Codename:	xenial
```

PR #132, opened 2016-08-23
Update README.md with latest Release version

xMudrii:
This is result of bug report #131

Myself I had problem using `doctl auth init` that would make me Auth my PC and DigitalOcean on easy way. 

As far as I noticed that feature is missing in v1.3.1.

Because I think some people will just copy link if they have OS X or Linux x64, it would not be bad to update links in README.md.
This could be applied to [DigitalOcean Community Tutorial](https://www.digitalocean.com/community/tutorials/how-to-use-doctl-the-official-digitalocean-command-line-client) too

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,12 +51,12 @@ You can download the archive from your browser, or copy its URL and retrieve it 
 cd ~
 
 # OS X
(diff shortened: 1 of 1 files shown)


andrewsomething:
:+1: Thanks for pointing this out. I've updated the tutorial as well.

PR #138, opened 2016-09-04
Add snapcraft packaging.

cmars:
This patch adds [snap](http://snapcraft.io/) packaging for the doctl command. A custom plugin is used to invoke the build.sh script.

To build the snap, with [snapcraft installed](http://snapcraft.io/create/), just run
`snapcraft`.

If this looks useful, I'd be happy to assist with
[publishing](https://snapcraft.io/docs/build-snaps/publish) the snap.

### DIFF
--- a/parts/plugins/x-build-sh.yaml
+++ b/parts/plugins/x-build-sh.yaml
@@ -1,0 +1,6 @@
+options:
+    source:
+        required: true
+    source-type:
+    source-tag:
+    source-branch:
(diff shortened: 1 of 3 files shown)


bryanl:
Hello,

We're putting together a plan to get precompiled binaries for one platform. The team is investigating this right now, and we should have consensus this week.

bryanl:
Does the `parts` directory need to be at the root?

cmars:
@bryanl It just needs to be located in a subdirectory relative to the snapcraft.yaml. What would be a better location for it?

mariogrip:
this is a much cleaner way of doing it: https://github.com/digitalocean/doctl/pull/154

cmars:
#154 looks good to me!

Issue #133, opened 2016-08-26
Ability to delete multiple snapshots


daoenix:
Being billing will begin on October 1st for customers storing snapshots. It would be beneficial to allow doctl to permit to delete multiple snapshots at once. Much like you can do with droplets. 

At the moment, we have noticed attempting to mass delete images, it provides syntax errors.

akshaychhajed:
Looks like #153 takes care of this. It did not get correctly linked due to a typo.

PR #139, opened 2016-09-15
Delete multiple Images by id

xmudrii:
Hey there.

I noticed Issue #133 - Ability to delete multiple snapshots.
I changed it for myself, tested and as I see it works. I managed to create possibility for deleting multiple images by id.

Now, I want to discuss about it, what do you guys think, is this bad idea, how can I improve it?

One negative downside could be that if you execute: `doctl compute image delete 1 2 b 3`, it will not test images, it will just delete. So when it find error (in example case it is b), it will delete 1 and 2, but 3 will still exist.

I don't know is it problem but I can get on idea to fix it if needed.
Also, I hope I didn't done it on most awful way :D

### DIFF
--- a/commands/confirmation.go
+++ b/commands/confirmation.go
@@ -1,0 +1,23 @@
+package commands
+
+import (
+	"fmt"
+	"strings"
+)
+
+func askForConfirm(message string) bool {
+	var answer string
+	warn("Are you sure you want to " + message + " (y/N) ? ")
+	_, err := fmt.Scanln(&answer)
+	if err != nil {
+		return false
+	}
+	return verifyAnswer(answer)
+}
+
+func verifyAnswer(answer string) bool {
+	if strings.ToLower(answer) == "y" || strings.ToLower(answer) == "yes" {
+		return true
+	}
+	return false
+}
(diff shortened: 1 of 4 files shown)


xmudrii:
Hello, I was away for few days and now I got onto problem.
I'm trying to do it myself, but looks like I have something I don't understand. :(

Looking at `images_test.go`, I see only one `testImage`, so only one `testImage.ID`. I don't see how would I be able to get one more image for testing.
This is original function:
```go
func TestImagesDelete(t *testing.T) {
                tm.images.On("Delete", testImage.ID).Return(nil)

                config.Args = append(config.Args, strconv.Itoa(testImage.ID), strconv.Itoa(testImage.ID))

                err := RunImagesDelete(config)
                assert.NoError(t, err)
        })
}
```

I don't see solution as changing it to:
```go
...
tm.images.On("Delete", testImage.ID, testImage.ID).Return(nil)
...
config.Args = append(config.Args, strconv.Itoa(testImage.ID), strconv.Itoa(testImage.ID))
```
In my opinion that will definitely not work, but I got is idea.

I would mostly like little instructions in what way I should do it. If you have time for it, of course :)

xmudrii:
Closing pull request as it got broken with forking thing... :( will open new after finishing confirmations - and I hope there will tests too. :)

PR #110, opened 2016-07-20
added confirmation before delete droplet

colkito:
just add a confirmation message before exec a delete action.

based in: https://gist.github.com/albrow/5882501

fixes #109

Issue #109, opened 2016-07-20
Confirm before delete droplet


colkito:
Feature request: We need a confirmation before delete (or any other critical action) a droplet. 😰

snoopdouglas:
#110 - pull request for this appears to have gone cold

xmudrii:
@bryanl I don't think it is polite to original author, but can someone of us take on this problem? It it already more than 2 months from last activity on #110.
I could do it, make it working and possibly add it to more locations (like deleting volumes, images, etc).

I was also thinking about, something in lines of `assume-yes` config. If it is set to `true` in cfg file, you will not be asked for confirmation (doctl will work in mode like it is now). Maybe it would not be bad for users who can live without it, or use doctl in integration with other tools/scripts

bryanl:
@xmudrii I don't think it would be impolite seeing as #110 has gone code for a while. The only requirement from my point of view would be to ensure it works in interactive and non-interactive settings (eg. in a script and triggered with a flag). As for the configuration item, let's revisit that in a second PR. Personally, I have no problem with it, but I would like to float the idea internally as well.

colkito:
Sorry for my delay! but at the moment I'm overloaded of work and sadly my free time to learn more about GO is very little...
If somebody can finish my PR could be great!
we are all agree about this requested feature, right?
greetings!

xmudrii:
@colkito 
Not a problem, I will have soon a first version on it so will make PR for @bryanl to check if is it ok for now. :)

xmudrii:
@bryanl This should be closed, I fixed it in #109. It should be Fixes, not fixs, at least TIL :P

xmudrii:
@aybabtme @nanzhong I think you can close this, I fixed this in #109

PR #148, opened 2016-10-04
[Fixs #109] Adding confirmation on delete actions

xmudrii:
Fixs #109 
Continues work from #110 
@bryanl I would like review if possible, in free time ofc. :D

# Adding confirmation on delete actions

This is basic yes/no confirmation. User need to answer `y`, `ye` or `yes`. For every other answer, command will be stopped.
For now it is only implemented for Droplet delete action.

List of done things that are working:
- [x] Created `confirmation.go` 
- [x] Added confirmation before Droplet delete (both on ID & Tag name)
- [x] Added local force flag
- [x] Added special function for improved output
- [x] Created test units for `confirmation.go`
- [x] Updated test units for `droplets.go`
- [x] Fixed Travis CI errors

Things could be done:
- [ ] Add confirmation on delete for other actions

## Explanation

First of all I created confirmation.go because of reusability. It would not be bad to extend it on volume delete, images delete, but this is something we will discuss later. Anyways I decided to make it reusable from beginning.

It is working like, you have to answer `y` or `yes` (case doesn't matters as I convert it to lower). If you answer anything other, operation will be aborted.

I added Force flag to `doit`, but I'm not sure is it right way, I would like some suggestion on this.

### DIFF
--- a/commands/doit.go
+++ b/commands/doit.go
@@ -53,9 +53,6 @@ var Output string
 // Verbose toggles verbose output.
 var Verbose bool
 
-// Force froces command execution.
-var Force bool
-
 var requiredColor = color.New(color.Bold, color.FgWhite).SprintfFunc()
 
 // Writer is where output should be written to.
@@ -80,7 +77,6 @@ func init() {
 	DoitCmd.PersistentFlags().StringVarP(&Token, "access-token", "t", "", "API V2 Access Token")
 	DoitCmd.PersistentFlags().StringVarP(&Output, "output", "o", "text", "output format [text|json]")
 	DoitCmd.PersistentFlags().BoolVarP(&Verbose, "verbose", "v", false, "verbose output")
-	DoitCmd.PersistentFlags().BoolVarP(&Force, "force", "f", false, "force execution")
 	DoitCmd.PersistentFlags().BoolVarP(&Trace, "trace", "", false, "trace api access")
 
 	viper.SetEnvPrefix("DIGITALOCEAN")
--- a/commands/doit.go
+++ b/commands/doit.go
@@ -53,6 +53,9 @@ var Output string
 // Verbose toggles verbose output.
 var Verbose bool
 
+// Force froces command execution.
+var Force bool
+
 var requiredColor = color.New(color.Bold, color.FgWhite).SprintfFunc()
 
 // Writer is where output should be written to.
@@ -77,6 +80,7 @@ func init() {
 	DoitCmd.PersistentFlags().StringVarP(&Token, "access-token", "t", "", "API V2 Access Token")
 	DoitCmd.PersistentFlags().StringVarP(&Output, "output", "o", "text", "output format [text|json]")
 	DoitCmd.PersistentFlags().BoolVarP(&Verbose, "verbose", "v", false, "verbose output")
+	DoitCmd.PersistentFlags().BoolVarP(&Force, "force", "f", false, "force execution")
 	DoitCmd.PersistentFlags().BoolVarP(&Trace, "trace", "", false, "trace api access")
 
 	viper.SetEnvPrefix("DIGITALOCEAN")
--- a/commands/droplets.go
+++ b/commands/droplets.go
@@ -69,7 +69,7 @@ func Droplet() *Command {
 
 	cmdRunDropletDelete := CmdBuilder(cmd, RunDropletDelete, "delete ID [ID|Name ...]", "Delete droplet by id or name", Writer,
 		aliasOpt("d", "del", "rm"), docCategories("droplet"))
-	AddBoolFlag(cmdRunDropletDelete, doctl.ArgDeleteForce, false, "Froce droplet delete")
+	AddBoolFlag(cmdRunDropletDelete, doctl.ArgDeleteForce, false, "Force droplet delete")
 
 	CmdBuilder(cmd, RunDropletGet, "get", "get droplet", Writer,
 		aliasOpt("g"), displayerType(&droplet{}), docCategories("droplet"))
--- a/commands/droplets.go
+++ b/commands/droplets.go
@@ -67,8 +67,9 @@ func Droplet() *Command {
 
 	AddStringSliceFlag(cmdDropletCreate, doctl.ArgVolumeList, []string{}, "Volumes to attach")
 
-	CmdBuilder(cmd, RunDropletDelete, "delete ID [ID|Name ...]", "Delete droplet by id or name", Writer,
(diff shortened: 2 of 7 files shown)


xmudrii:
## Update 10/06/2016:

First of all I did some `AskForConfirm` code refactor. Now it is only one function and it is using `bufio.NewReader` instead of `fmt.Scanf`. It is because I'm in progress of creating tests and it found it easier. 
It can be still reverted, but epic low chance.

I have improved question, it is not over 4 lines, it is only in one line now:
```
Warning: Are you sure you want to delete droplets(s) (y/N)? *user input - y, n...*
```
This involved creation of new `warn` function - `warnConfirm`, so it would like @bryanl to confirm me is it OK to create new func.

CI is still unresolved, but it will be soon. :)

xmudrii:
## Update 10/07/2016 - whole changelog:

### confirmation.go:
Confirmation input is changed to be in fashion like `retrieveUserTokenFunc`. It made me easier to make tests for it and I wanted to use same input method across whole project.
It is now returning `error` instead of `bool` as I see `error` is mostly used and I want to create it with some standard.

### confirmation_test.go
I created basic test unit that test:
* Input `yes`
* Input `no`
* Input `any`
* Input with error
I make sure that it uses `testify` like other tests at `doctl`.

### doit.go
Added global bool flag - `force`. It can be called as `--force` or only `-f`.

### droplets.go
Implemented ask for confirmation.

### droplets_test.go
Added `Force` for tests.

### errors.go
Added `warnConfirm` function for displaying `Are you sure` message.
Problem with `warn` was `\n\n`.
`warn` would output:
```
Warning: Are you sure you want to delete droplet(s)? (y/N) ?
--- new line ---
--- new line ---
*User input goes here* - y/n
```
As I wanted it to be in one line e.g. :
```
Warning: Are you sure you want to delete droplet(s)? (y/N) ? *User input goes here* - y/n
```
I created new function `warnConfirm` that works like this.

### Comments:
My main concern is `Force` flag. I have feel like it can be implemented much better. I'm not sure is it good idea to make it global, and is it good idea to call Force from `droplets.go` and `droplets_test.go`.
@bryanl Removed [WIP] flag, it is usable and stable now. Improvements can probably be done, but I would like review at first, as it has bit changes. :)

commands/doit.go
```diff
@@ -77,6 +80,7 @@ func init() {
 	DoitCmd.PersistentFlags().StringVarP(&Token, "access-token", "t", "", "API V2 Access Token")
 	DoitCmd.PersistentFlags().StringVarP(&Output, "output", "o", "text", "output format [text|json]")
 	DoitCmd.PersistentFlags().BoolVarP(&Verbose, "verbose", "v", false, "verbose output")
+	DoitCmd.PersistentFlags().BoolVarP(&Force, "force", "f", false, "force execution")
```
bryanl: Should `force` be global option or is it specific to the command?

xmudrii:
@bryanl `force` flag has been updated!
I made it now to work as local flag tied to `cmdRunDropletDelete`. Global flag was overkill IMO, it is for now only usable on one action, and even in case we extend to other commands, it is only for delete.
I can revert to global if needed, but I think it should be specific to command. Anyways you know it better to me, so I'm open to suggestions. ;)

It's ready to be merged, I think.
I would like opinion on extending it to other delete actions. As far as I see there are following delete actions:
* Domains
* Domains - Records
* Floating IPs
* Images
* SSH keys
* Tags
Don't does all of this needs delete confirmation, but they could use it.

One drawback of it is that it need to be called with `--force`. I'm out of luck to make it only with `-f`, as `aliasOpt("f")` is not working on `AddBoolFlag`. I would like to make it work for `-f` because it is faster and somehow natural to Unix users.

Tests are corrected to use new `force` flag and Travis CI/Appveyor passes successfully :)

bryanl:
@xmudrii There is an issue with short flags. I've seen it before, but haven't tracked down the culprit in pflags.

commands/droplets.go
```diff
@@ -67,8 +67,9 @@ func Droplet() *Command {
 
 	AddStringSliceFlag(cmdDropletCreate, doctl.ArgVolumeList, []string{}, "Volumes to attach")
 
-	CmdBuilder(cmd, RunDropletDelete, "delete ID [ID|Name ...]", "Delete droplet by id or name", Writer,
+	cmdRunDropletDelete := CmdBuilder(cmd, RunDropletDelete, "delete ID [ID|Name ...]", "Delete droplet by id or name", Writer,
 		aliasOpt("d", "del", "rm"), docCategories("droplet"))
+	AddBoolFlag(cmdRunDropletDelete, doctl.ArgDeleteForce, false, "Froce droplet delete")
```
bryanl: typo :)

xmudrii:
@bryanl didn't noticed, it happens :D Fixed typo.
Maybe I could change it to `Force droplet deletion` ? What would sound better

justintime4tea:
When doctl is used within automation how do we "bypass" the dialog or perform "--non-interactive" actions?

xmudrii:
‎You can do it by using `force` flag, e.g. `doctl compute droplet delete id --force`. :)                                                                                                                                                                                                                From: Justin GrossSent: sreda, 26. oktobar 2016. 23.32To: digitalocean/doctlReply To: digitalocean/doctlCc: Marko Mudrinic; MentionSubject: Re: [digitalocean/doctl] [Fixs #109] Adding confirmation on delete actions (#148)When doctl is used within automation how do we "bypass" the dialog or perform "--non-interactive" actions?

—You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or mute the thread.


  
  




{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/digitalocean/doctl","title":"digitalocean/doctl","subtitle":"GitHub repository","main_image_url":"https://cloud.githubusercontent.com/assets/143418/17495839/a5054eac-5d88-11e6-95fc-7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent.com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-b52498112777.png","action":{"name":"Open in GitHub","url":"https://github.com/digitalocean/doctl"}},"updates":{"snippets":[{"icon":"PERSON","message":"@justintime4tea in #148: When doctl is used within automation how do we \"bypass\" the dialog or perform \"--non-interactive\" actions?"}],"action":{"name":"View Pull Request","url":"https://github.com/digitalocean/doctl/pull/148#issuecomment-256483430"}}}

PR #160, opened 2016-10-26
Improves tag handling for droplet create

### LINKED ISSUE #127: Specifying a nonexisting tag to 'droplet create' gives an error but still creates the droplet
```
doctl compute droplet create --size 512mb --tag-name test-metadata-issue --image centos-7-2-x64 --region fra1 doctl-issue-test.example.org
```

returns

```
doctl computID      Name                Public IPv4 Public IPv6 Memory  VCPUs   Disk    Region  Image       Status  Tags
22002227    doctl-issue-test.example.org                    512 1   20  fra1    CentOS 7.2 x64  new
create droplet

Usage:
  doctl compute droplet create NAME [NAME ...] [flags]
..snip..

Error: POST https://api.digitalocean.com/v2/tags/test-metadata-issue/resources: 404 The resource you were accessing could not be found.
```

Although you can see some information about a droplet being created at the top, the whole usage information is a long enough wall of text for me to not notice that, but to do notice the error at the end. So I was assuming the droplets weren't created, but it turns out that they actually were.

This is caused by doctl first creating the machine, and then running the tag API. As these are [not atomic](https://digitalocean.uservoice.com/forums/136585-digitalocean/suggestions/15593823-allow-setting-tags-when-creating-a-droplet), using a nonexisting tag will not prevent you from creating the droplet. But it's not really obvious that this has happened.

Unless the API is improved, the best fix for the client would probably to verify the tag exists before using them.

akshaychhajed:
Fixes #127 
Runs check to ensure tags exist before creating droplet

### DIFF
--- a/commands/droplets.go
+++ b/commands/droplets.go
@@ -244,6 +244,19 @@ func RunDropletCreate(c *CmdConfig) error {
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
+
+			if tagName != "" {
+				tag, err := ts.Get(tagName)
+				if err != nil {
+					errs <- err
+					return
+				}
+				if tag == nil {
+					errs <- fmt.Errorf("Specified Tag must exist")
+					return
+				}
+			}
+
 			d, err := ds.Create(dcr, wait)
 			if err != nil {
 				errs <- err
@@ -272,13 +285,13 @@ func RunDropletCreate(c *CmdConfig) error {
 	close(errs)
 
 	item := &droplet{droplets: createdList}
-	c.Display(item)
 
 	for err := range errs {
 		if err != nil {
 			return err
 		}
 	}
+	c.Display(item)
 
 	return nil
 }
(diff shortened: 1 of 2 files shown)


xmudrii:
Looks good.

But what about creating tag alongside Droplet? That can be done automatically, when you try to create Droplet with new tag, create tag too. Or ask user to create tag/ignore/cancel creation.
I personally I would go automatically, as I think less interaction == better, especially because doctl is made for scripts integration and so on.
Probably @bryanl can give us his opinion on this topic. :+1:

mchitten:
This looks fine. Today we implemented [tags in droplet create](https://developers.digitalocean.com/documentation/changelog/api-v2/tagging-in-droplet-create/), which does essentially what @xmudrii is suggesting (creating a tag and applying it to the droplet during create). That is implemented in #165.

For this, 👍 .

nanzhong:
lgtm 👍

PR #161, opened 2016-10-26
Added multiple tag delete, tag delete confirmation

xmudrii:
Droplet Tagging is released today, tags are more relevant.

First change is that you have possibility to delete multiple tags.
Second one is thing I add one by one is delete confirmation. Now you're asked are you sure or you have to specify `--force` flag.

Also test for multiple delete is added.


mchitten:
LGTM

PR #165, opened 2016-11-11
allow "tags" attribute for droplet create

mchitten:
This allows `tag-names` as an attribute for droplet create, in accordance with https://developers.digitalocean.com/documentation/changelog/api-v2/tagging-in-droplet-create/.

cc @hugocorbucci @daveworth

### DIFF
--- a/args.go
+++ b/args.go
@@ -110,6 +110,8 @@ const (
 	ArgPollTime = "poll-timeout"
 	// ArgTagName is a tag name
 	ArgTagName = "tag-name"
+	// ArgTagNames is a slice of possible tag names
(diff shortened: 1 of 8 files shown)


hugocorbucci:
LGTM

nanzhong:
@mchitten create also is able to take a single tag, and when specified it runs https://github.com/mchitten/doctl/blob/def538d81418272b107d9cdbea2321ab3897715d/commands/droplets.go#L260-L272. Pretty sure tagging is idempotent in the API, but just wanted to double check with you and see whether we need to do anything about tag consolidation if both `tag-name` and tag-names` are specified.

PR #174, opened 2016-12-30
Imply that doctl is actively doing something

fcoelho:
The "Validating token: " message used to end with ": ", which looks like a prompt for more info. The ellipsis implies that doctl itself is doing something instead of waiting for more user input.

This is a really small change. I needed to re-`auth init` on a slow connection and had that "prompt" sitting in for more than a few seconds. I then pasted the access token again because I though it was waiting for something. The error really is on my part, but I think this makes things clearer.

### DIFF
--- a/commands/auth.go
+++ b/commands/auth.go
@@ -81,7 +81,7 @@ func RunAuthInit(c *CmdConfig) error {
 	viper.Set("access-token", string(token))
 
 	fmt.Fprintln(c.Out)
-	fmt.Fprint(c.Out, "Validating token: ")
+	fmt.Fprint(c.Out, "Validating token... ")
 
 	// need to initial the godo client since we've changed the configuration.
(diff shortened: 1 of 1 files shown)


xmudrii:
I agree with this one. If you have slow connection, or even in case your connection is not working, `Validating token:` is confusing. 
I'm just thinking now about splitting `OK` message to two lines. So instead of `Validating token... OK`, to have something like:
```
Validating token...
Token validation successful.
```
or
```
Validating token...
Token validation unsuccessful - invalid token!
```
Not needed, just something I got on my mind, let's see what others say. :+1:

nanzhong:
The `:` making this feel like a prompt is definitely confusing. This change looks good to me and should definitely help with clarify.

@xmudrii I think both inline and two line versions are reasonable, but perhaps for now let's keep it the same as before (inline).

PR #177, opened 2017-01-03
Fixed #176

### LINKED ISSUE #176: Some output is forced to use white color
The help messages include some parts in white, even when background is white. It is very hard to read.

Example (see last line):

---

whitetext

---

xmudrii:
Fixes #176 
This should work, at least it does for me.

cc @aybabtme @nanzhong


nanzhong:
Looks good to me.

PR #178, opened 2017-01-03
[Fixes #168] Implementing command forwarding for external SSH

### LINKED ISSUE #168: pass ssh command line down to the actual ssh
Right now ``doctl compute ssh droplet`` simply ignores all command line args specified so things like these do not work:

```
$ doctl compute ssh mydroplet uptime
$ doctl compute ssh mydroplet free
```

I think it would be cool to add ability to simply pass through all command line args not processed by ``doctl`` itself to the underlying ssh command for easy scripting

xmudrii:
With this PR, we should fix #168. However, this is not really same idea as it's in issue.
Issue is proposing following implementation:
```
doctl compute ssh mydroplet uptime
```
I decided to follow template of SSH unit, so I went with flag (`--ssh-command`).
When I execute `doctl compute ssh mydroplet`, I'll work as usual.
When I execute `doctl compute ssh test1 --ssh-command uptime`, it'll will only return output of command, as expected:
```
 22:32:26 up  1:07,  0 users,  load average: 0.00, 0.00, 0.00
```
I want to know do you like it, how it looks and could it be done better.
Just to note, this is only implemented for external SSH (OS X & Linux). I need to look more in Windows implementation of SSH client, but first I want to know is this one OK. If this is right, I'll do the same for internal if possible.

cc @nanzhong @aybabtme

### DIFF
--- a/args.go
+++ b/args.go
@@ -92,6 +92,8 @@ const (
 	ArgsSSHAgentForwarding = "ssh-agent-forwarding"
 	// ArgsSSHPrivateIP is a ssh argument.
 	ArgsSSHPrivateIP = "ssh-private-ip"
+	// ArgSSHCommand is a ssh argument.
+	ArgSSHCommand = "ssh-command"
 	// ArgUserData is a user data argument.
 	ArgUserData = "user-data"
 	// ArgUserDataFile is a user data file location argument.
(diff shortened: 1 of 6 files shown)


mdlayher:
Code looks good.  Mind writing a test?

xmudrii:
@mdlayher Added test, how does it looks?

nanzhong:
lgtm! @grampajoe also gave it a 👍

Issue #172, opened 2016-12-15
[Idea] Add list of resources that will be deleted before asking for confirmation


xmudrii:
Now, when you execute command like `doctl compute droplet delete 1`, you'll see:
`Warning: Are you sure you want to delete droplet(s) (y/N) ? y`.
But that's all. If you used `--force` flag, you'll neither see confirmation which is OK.

Mine idea, is to show list of Droplets that will be deleted in time of asking for confirmation. If you execute command like above again, you could see something like:
```
Following Droplets will be deleted:
ID		Name			Public IPv4	Private IPv4	Public IPv6	Memory	VCPUs	Disk	Region	Image			Status	Tags
1	droplet	ip					1024	1	30	nyc2	Ubuntu 16.04.1 x64	active	

Warning: This is not reversible process! Are you sure you want to delete droplet(s) (y/N) ?
```

Something like you would see when you execute `apt-get remove` or any other package manager remove/install.

This is just idea, I want feedback from maintainers. I don't think I'll have time to do so, but in some future, it could be done by me if you approve idea. :+1: 

Will not cc lots of maintainers now, let's just call @aybabtme and @andrewsomething =)

bnordland:
Since this was just an idea, I didn't submit a pull request yet, but I did commit code that has the following behavior (https://github.com/bnordland/doctl/commit/d2ca54903966c9a2181817bade2eb94cbdfc3fa5):

```
$ doctl compute droplet delete test test2
Following Droplet(s) will be deleted:
ID		Name	Public IPv4	Private IPv4	Public IPv6	Memory	VCPUs	Disk	Region	Image		Status	Tags
36243836	test	192.241.149.114					512	1	20	nyc1	Debian 8.6 x64	active	
36243845	test2	159.203.191.220					512	1	20	nyc1	Debian 8.6 x64	active	
Warning: Are you sure you want to delete droplet(s) (y/N) ? 
```

Behavior:

1. If force is specified the list is **not** shown
2. If one or more of the droplets specified is not found, an error is shown without showing the list (e.g. Error: droplet with name "test3" could not be found)
2. If doing by ID or Name it'll list the droplets to be deleted is shown before the confirmation as is shown above.

_Note: When doing a delete by tagName it does not show the list (I didn't get to that part as this is more a sample)_

xmudrii:
LGTM. :+1: It's working flawless and as I wanted in idea. Please create pull request so maintainers can take look and merge it. =)

[...]

PR #179, opened 2017-01-04
Display list of droplets before deleting

bnordland:
For issue #172, this will show the list of droplets being deleted if we
aren’t forcing so that the confirmation has more information for us.
Note: does not handle printing the list when doing a DeleteByTag.

We want to make sure that the list shown before the confirmation is actually the list that will be deleted which is why we retrieve the list once and use the same list before and after the confirmation. That way if the confirmation had been up and someone else deletes and recreates droplet "test" as a different one, since it is working off of IDs we don't delete one that was not shown.
If we wanted to do deleteByTag as well we would probably want a similar mechanism to do ListByTag, show the confirmation, and then loop through the same list after the confirmation doing the ds.Delete from the original list rather than ds.DeleteByTag directly.

### DIFF
--- a/commands/droplets.go
+++ b/commands/droplets.go
@@ -475,9 +475,26 @@ func RunDropletDelete(c *CmdConfig) error {
 		return nil
 	}
 
-	if force || AskForConfirm("delete droplet(s)") == nil {
-
-		fn := func(ids []int) error {
+	fn := func(ids []int) error {
+		if !force {
+		
+			// If we aren't forcing, print out the list
+			fmt.Printf("Following Droplet(s) will be deleted:\n")
+			
+			var list do.Droplets
+			for _, id := range ids {
+				d, err := ds.Get(id)
+				if err != nil {
+					return err
+				}
+				list = append(list, *d)	
+			}
+			
+			item := &droplet{droplets: list}
+			c.Display(item)
+		}
+			
+		if force || AskForConfirm("delete droplet(s)") == nil {
 			for _, id := range ids {
 				if err := ds.Delete(id); err != nil {
 					return fmt.Errorf("unable to delete droplet %d: %v", id, err)
@@ -485,9 +502,12 @@ func RunDropletDelete(c *CmdConfig) error {
 			}
 			return nil
 		}
-		return matchDroplets(c.Args, ds, fn)
+		return fmt.Errorf("operation aborted")
+			
+			
 	}
-	return fmt.Errorf("operation aborted")
+	return matchDroplets(c.Args, ds, fn)
+	
 
 	return nil
 
--- a/commands/droplets.go
+++ b/commands/droplets.go
@@ -479,7 +479,7 @@ func RunDropletDelete(c *CmdConfig) error {
 		if !force {
 		
 			// If we aren't forcing, print out the list
-			fmt.Printf("Following Droplet(s) will be deleted:\n")
+			fmt.Println("Following Droplet(s) will be deleted")
 			
 			var list do.Droplets
 			for _, id := range ids {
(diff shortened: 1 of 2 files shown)


xmudrii:
LGTM, but you need LGTM from some maintainer. :D

cc @aybabtme @nanzhong

commands/droplets.go
```diff
@@ -475,19 +475,39 @@ func RunDropletDelete(c *CmdConfig) error {
...
 	}
 
-	if force || AskForConfirm("delete droplet(s)") == nil {
-
-		fn := func(ids []int) error {
+	fn := func(ids []int) error {
+		if !force {
+		
+			// If we aren't forcing, print out the list
+			fmt.Printf("Following Droplet(s) will be deleted:\n")
```
xmudrii: Maybe a tiny nitpick, but this could be changed with `Println` (in that case `\n` is not needed).

bnordland: Good point! I'll make that change.

commands/droplets_test.go
```diff
@@ -160,7 +160,9 @@ func TestDropletDelete(t *testing.T) {
 
 func TestDropletDeleteByTag(t *testing.T) {
 	withTestClient(t, func(config *CmdConfig, tm *tcMocks) {
-		tm.droplets.On("DeleteByTag", "my-tag").Return(nil)
```
xmudrii: What is the reason behind this change? This function is made to test DeleteByTag and I think it should stay as it is, but if there is good reason behind it, it can be changed.

xmudrii:
As #180 got accepted, does we need a rebase now?

commands/droplets_test.go
bnordland: Since DeleteByTag now actually under the covers does a ListByTag, and then a delete on the returned list items by id, this means that the test needs to have mocks for ListByTag, and Delete instead of calling the droplet service DeleteByTag.

The reason that I changed the actual implementation to do a ListByTag and then do the deletes separately instead of the obvious answer of:
ListByTag
Show confirmation
DeleteByTag

Is to prevent the following from happening:
1. The user does a DeleteByTag named "test" and is shown a list of three droplets to delete.
2. something else, either the user in another process or script or something tags another droplet under the same tag, so there are now four droplets tagged with the tag "test"
3. The user confirms the DeleteByTag request. But instead of the original three droplets they were shown four are deleted which could lead to inadvertent data loss.

Because of the implementation, the mocks need to be enhanced.

Does that help answer your question?

bnordland:
This pull request is now up to date with the master.

commands/droplets.go
```diff
@@ -462,33 +463,60 @@ func RunDropletDelete(c *CmdConfig) error {
...
+
 	}
-	return fmt.Errorf("operation aborted")
 
-	return nil
+	if len(c.Args) < 1 && tagName == "" {
+		return doctl.NewMissingArgsErr(c.NS)
+	} else if len(c.Args) > 0 && tagName != "" {
+		return fmt.Errorf("please specify droplets identifiers or a tag name")
+	} else if tagName != "" {
```
xmudrii: Thank you for answer and explanation. Code is working very well. However, this is probably a point where something can be improved.
This part looks OK, but this is not needed for force action. In case of `force`, you don't want to see list and you don't need for list to be generated. In my opinion, you should check is `force` true, and if it is — exit function with `return ds.DeleteByTag(tagName)`. If it's false, execute below code as it is now.
Not something glorious, but it'll save one API call and speed up process (maybe not noticeable) because it doesn't need to query for all Droplets with tag, generate list and go through the `deleteFn` function.

About mocks, I think there is no anything to be improved. There are substituting API actions with desired response so it doesn't return error (like unauthenticated, resource doesn't exist, etc.). There are generated for code in `do` folder using [mockery](https://github.com/vektra/mockery).

mauricio:
This, unfortunately, changes existing behavior, we could introduce a flag that will print the droplets and ask before deleting them but adding this confirmation check now will break existing callers, which is not ideal so we can't really go forward with this.

PR #185, opened 2017-01-25
Updated command descriptions to match template

xmudrii:
What do you think about this change? I updated descriptions to match one template and changed a few ones.
I'm not sure is `create` descriptions good, but I'll revert if it is problem. =)

cc @nanzhong

### DIFF
--- a/commands/droplets.go
+++ b/commands/droplets.go
@@ -43,13 +43,13 @@ func Droplet() *Command {
 		IsIndex:       true,
 	}
 
-	CmdBuilder(cmd, RunDropletActions, "actions <droplet id>", "droplet actions", Writer,
+	CmdBuilder(cmd, RunDropletActions, "actions <droplet-id>", "droplet actions", Writer,
 		aliasOpt("a"), displayerType(&action{}), docCategories("droplet"))
 
-	CmdBuilder(cmd, RunDropletBackups, "backups <droplet id>", "droplet backups", Writer,
+	CmdBuilder(cmd, RunDropletBackups, "backups <droplet-id>", "droplet backups", Writer,
 		aliasOpt("b"), displayerType(&image{}), docCategories("droplet"))
 
-	cmdDropletCreate := CmdBuilder(cmd, RunDropletCreate, "create NAME [NAME ...]", "create droplet", Writer,
+	cmdDropletCreate := CmdBuilder(cmd, RunDropletCreate, "create [droplet-name ...]", "create droplet", Writer,
 		aliasOpt("c"), displayerType(&droplet{}), docCategories("droplet"))
 	AddStringSliceFlag(cmdDropletCreate, doctl.ArgSSHKeys, "", []string{}, "SSH Keys or fingerprints")
 	AddStringFlag(cmdDropletCreate, doctl.ArgUserData, "", "", "User data")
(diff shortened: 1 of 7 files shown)


nanzhong:
Oh wow, I didn't realize there was so much variation. I see that there is still some variance in the use of `<>` and `[]`; should we just stick to one? I feel like `[]` was initially used for optional arguments, but it seems consistency has been lost. 

cc @mauricio @grampajoe what do you folks think?

mauricio:
I think it's awesome. And we can just continue to use [] for everything to make sure it's all the same.

nanzhong:
I agree about being consistent; I was actually thinking about standardizing on using the convention of `<>` for required and `[]` for optional arguments.

> On Unix-like platforms, usage messages usually follow the same common pattern:
> - They often begin with "Usage:" , the command, followed by a list of arguments.
> - To indicate optional arguments, Square brackets are commonly used, and can also be used to group parameters that must be specified together.
> - To indicate required arguments, Angled brackets are commonly used, following the same grouping conventions as square brackets.
> - Exclusive parameters can be indicated by separating them with vertical bars within groups.
https://en.wikipedia.org/wiki/Usage_message

xmudrii:
In this case, I agree with @nanzhong. I'm going to change it now, so we can see how does it look.
Expect new commit in few minutes =)

xmudrii:
In meanwhile, what is this https://github.com/digitalocean/doctl/blob/master/commands/droplet_actions.go#L62 ?
It's taking `droplet-id` but there is no any explanation what is `action-id` in this case?
I skipped it, but i'll add desc in next change.

xmudrii:
@nanzhong updated =)

commands/droplets.go
```diff
@@ -43,13 +43,13 @@ func Droplet() *Command {
...
-	CmdBuilder(cmd, RunDropletActions, "actions <droplet id>", "droplet actions", Writer,
+	CmdBuilder(cmd, RunDropletActions, "actions <droplet-id>", "droplet actions", Writer,
 		aliasOpt("a"), displayerType(&action{}), docCategories("droplet"))
 
-	CmdBuilder(cmd, RunDropletBackups, "backups <droplet id>", "droplet backups", Writer,
+	CmdBuilder(cmd, RunDropletBackups, "backups <droplet-id>", "droplet backups", Writer,
 		aliasOpt("b"), displayerType(&image{}), docCategories("droplet"))
 
-	cmdDropletCreate := CmdBuilder(cmd, RunDropletCreate, "create NAME [NAME ...]", "create droplet", Writer,
+	cmdDropletCreate := CmdBuilder(cmd, RunDropletCreate, "create <droplet-name>", "create droplet", Writer,
```
grampajoe: One more nitpick: this and `delete <droplet-id|droplet-name>` below can accept one _or more_ names/ids, so we might need one param in `<>`s followed by a list in `[]`s.

nanzhong:
@xmudrii `RunDropletActionGet` returns an event for a droplet, eg.

```
~/d/r/d/c/doctl ❯❯❯ ./doctl compute droplet-action get 36383102 --action-id 191941476
ID              Status          Type    Started At                      Completed At                    Resource ID     Resource Type   Region
191941476       completed       backup  2017-01-23 10:20:47 +0000 UTC   2017-01-23 10:24:00 +0000 UTC   36383102        droplet         nyc1
```

commands/droplets.go
xmudrii: Something like `<droplet-name[s]>` ?

nanzhong: Perhaps `create <droplet-name> [<droplet-name> ...]` to denote 1 or more arguments?

grampajoe: I was thinking closer to the original, like `<droplet-name> [droplet-name...]`, but either would be good.

xmudrii:
Check now, I think I found it in more than two places =)

grampajoe:
LGTM! Thanks!

nanzhong:
👍

PR #187, opened 2017-01-28
Bumping Dockerfile doctl version to 1.5.0

xmudrii:
Changed `Dockerfile` to use version `1.5.0` instead of `1.4.0`.

cc @nanzhong


grampajoe:
LGTM, thanks! 🌈

PR #188, opened 2017-02-01
commands: implement command aliases

### LINKED ISSUE #67: git-style aliases
## What

Allow git-style aliases for commands.
## Why

Some commands are more frequently used than others by users, so allowing people to alias them to their convenience will provide a more ergonomic experience.
## How

Possible impl:
- Add the aliases definitions to the config file.
- Before starting the `cobra` stuff, expand the `os.Args` according to the aliases.

xmudrii:
# Aliases
Fixes #67 

## What
Allow aliases in doctl

## Why
Some commands are used more than others. This will allow you to alias commands so you have better experience using doctl.

## How does this work
This works by rewriting `os.Args` before initializing `cobra`. It tries to find pattern in aliases file and assembles it as one command.

## Where are aliases stored
In `~/.config/doctl/alias.yaml`.

## How does this file should look
Format for alias file is `yaml`. Type of key is string (written in `''`).
Example:
```yaml
'dls': compute droplet list
'dg': compute droplet get
'r': account ratelimit
```

## Why is `alias.yaml` separate file from `config.yaml`
Separating it form config can help sharing and versioning your aliases. Many of us have `dotfiles` repository where we store stuff like this (config, aliases...). As `config.yaml` is not supposed to be shared, as it contains sensitive data such as `access-token`, I believe it's the best idea to use a separate file for aliases..

## What's the status of this PR
This Pull Request is ready to be merged feature wise. It's fully working on both Linux and Windows.

### DIFF
--- a/commands/alias.go
+++ b/commands/alias.go
@@ -1,0 +1,164 @@
+/*
+Copyright 2017 The Doctl Authors All rights reserved.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package commands
+
+import (
+	"fmt"
+
+	"github.com/digitalocean/doctl"
+	"github.com/gobwas/glob"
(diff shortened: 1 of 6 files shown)


xmudrii:
Ready! I can officially say that I'm done with aliases. =D
I hope it's OK and it'll be useful to `doctl`. =)

If you find something and/or want something to be added, please let me know.

cc @nanzhong @grampajoe @aybabtme

xmudrii:
@mauricio out of curiosity, any update on this :D
is it planned for doctl? :100:

mauricio:
I feel like we're better off letting users configure aliases on their own envs instead of trying to replicate a feature that's already available at their own shells.

PR #203, opened 2017-03-17
replace 'auth login' with 'auth init' in readme

xmudrii:
README was stating to use `auth login` which is deprecated as of 1.4.0.
If above can be said better, please let me know and I'll update it.

cc @mauricio @viola


viola:
@xmudrii ah thank you for that! LGTM

Issue #90, opened 2016-06-22
Bash Completion


Grynn:
Feature request: We need bash completion...

doctl <tab> should show compute, account, auth ...

andrewsomething:
Running this little program will create a bash completion file for doctl.

```go
package main

import "github.com/digitalocean/doctl/commands"

func main() {
	cmd := commands.Init()
	cmd.GenBashCompletionFile("doctl.sh")
}
```

Save it as `doctl-bash.go` and then with the go compiler installed, you can run it with `go run doctl-bash.go` It will produce a file called `doctl.sh` which if you can place in `/etc/bash_completion.d/`

The question is how to best get it in the hands of the user. We could create it as part of the build/release process. Then homebrew (and hypothetical rpms/debs) could install it. An alternative might be to add a new command that does this. For instance, Hugo provides a genautocomplete sub-command.

bryanl:
This is currently on the plate. I'm thinking bash, zsh, and fish (mostly because I use it) for the first version.

Grynn:
For what it's worth I'd suggest generating the completion as part of the build process; then homebrew (or rpm/deb) can just copy it into the right place. (As suggested by @andrewsomething). 

```
$>brew info git
...
==> Caveats
Bash completion has been installed to:
  /usr/local/etc/bash_completion.d

zsh completion has been installed to:
  /usr/local/share/zsh/site-functions
```


The one other way I've seen it work, which is much harder, is to have a stub completion that actually executes `doctl compute --help --porcelain` or similar and then uses that to compute the completion. This can take into account chars you've already typed. So `doctl compute droplet delete abc-` would complete names beginning with `abc-`. See this: https://github.com/HiroKws/ArtisanBashCompletion for a pretty awesome example.

Grynn:
Also @andrewsomething 

```
$ go get github.com/digitalocean/doctl/commands
# cd /Users/vishal/src/golang/src/github.com/digitalocean/doctl; git submodule update --init --recursive
fatal: No url found for submodule path 'vendor/github.com/vektra/errors' in .gitmodules
package github.com/digitalocean/doctl/commands: exit status 128
```

biji:
@andrewsomething please share resulted script.sh

andrewsomething:
@biji Sure thing. This is the output for 1.3.0: https://gist.github.com/andrewsomething/7ba993a7f7897936c6946ffed2f1edc6

stormbeta:
[...]

PR #204, opened 2017-03-18
bash completion file

phlogisto:
Bash completion file. Copy doctl to `/etc/bash_completion.d/` on most linux distros.

### DIFF
--- a/bash_completion/doctl
+++ b/bash_completion/doctl
@@ -1,0 +1,98 @@
+_doctl_prev()
+{
+    local cur prev level compreply
+    level=$1
+    cur="${COMP_WORDS[COMP_CWORD]}"
+    prev="${COMP_WORDS[COMP_CWORD - level - 1]}"
+    compreply=""
+
+    case "$prev" in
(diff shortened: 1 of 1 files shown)


xmudrii:
Looks good and it works great. It would be good if we could 'add it right into' `doctl`. 

To accomplish this, I'm trying similar to `kubectl` completion feature and it is now available [in this commit](https://github.com/xmudrii/doctl/commit/b73dae21bcc981adb63255a4576fa1c51a4e0cfa). Discussion about issue is in #90. However, I don't have too much time now, so it stays for now.
Currently only available for `bash` but `zsh` and `fish` should be done some day. Beware, it's far from good, but it generates it.

If you are interested, this is [completion file generated by `cobra` with above code](https://gist.github.com/xmudrii/3bf4571c5444cf347e1cdea0e7606a64)

phlogisto:
Hi,
Great. I made this for my self and shared it more as a polite way to ask for a feature request.
Keep up the great work!

PR #214, opened 2017-04-09
Rename DetachByDropletID function to Detach

xmudrii:
This PR renames function `detach-by-droplet-id` to `detach`. Providing Droplet ID is as of [April 4th, 2017](https://github.com/digitalocean/godo/pull/136#issuecomment-291612894) required and it's not possible to detach without anymore.
`detach-by-droplet-id` is long for a function and makes its usage inconvenient.

This PR also regenerates mocks as it was forgotten in #208.

cc @mauricio @hugocorbucci @mchitten @magicgrl111

### DIFF
--- a/commands/volume_actions.go
+++ b/commands/volume_actions.go
@@ -61,8 +61,11 @@ func VolumeAction() *Command {
 	CmdBuilder(cmd, RunVolumeAttach, "attach <volume-id> <droplet-id>", "attach a volume", Writer,
 		aliasOpt("a"))
 
-	CmdBuilder(cmd, RunVolumeDetachByDropletID, "detach-by-droplet-id <volume-id> <droplet-id>", "detach a volume by droplet ID", Writer,
-		aliasOpt("dd"))
+	CmdBuilder(cmd, RunVolumeDetach, "detach <volume-id> <droplet-id>", "detach a volume", Writer,
+		aliasOpt("d"))
+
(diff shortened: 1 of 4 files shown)


mauricio:
@xmudrii I think we're better off having `detach` and `detach-by-droplet-id` both for now and just print a deprecation warning when `detach-by-droplet-id` gets called so people have some time to update.

xmudrii:
@mauricio good reasoning, I agree on it.

I will update PR to add Detach function and update DetachByDropletID to do the same, but just with depreciation warning.
I would like to check is it okay to print warning to output or just to include deprecated in command description?

mauricio:
We can just include deprecated at the description.

xmudrii:
@mauricio is it good this way? :100:

mauricio:
I think we should instruct people to use `detach`, so instead of just `(depecated)` use `(deprecated - use detach instead)`.

xmudrii:
@mauricio yep, I agree. Change pushed :smile:

xmudrii:
@mauricio travis failed on tip for unknown reason. can you restart build?

joonas:
> @mauricio travis failed on tip for unknown reason. can you restart build?

I've gone ahead and restarted the build, looks like we're back to 💚

xmudrii:
@mauricio @joonas Can confirm, it's green now, thanks for restarting :green_heart: :green_apple:

PR #233, opened 2017-06-02
doctl/snap: bash/zsh completions on install

xmudrii:
Automatically install bash and zsh completions on doctl install via snap.
To be in effect, doctl needs to be re-released with new script.
It's not guaranteed to work, but we will see results once we build doctl with new script. It appears to work for juju, so I hope it will work for us as well.

/cc @mauricio @lxfontes


mauricio:
Great!

PR #237, opened 2017-06-10
Fix releases link

igalic:


xmudrii:
Looks good. :+1:

There is one more mistake, same as this one, but for Windows release.
Also, doctl command description could be updated as well.

I already created #231 but no one took a look so far. I probably need to rebase it by now.

PR #239, opened 2017-06-21
commands/output: adds features and volumes to droplet's list output

### LINKED ISSUE #213: Show features enabled in droplet list table
I was experimenting with monitoring on DigitalOcean and released that doctl doesn't explicitly show what features are enabled on the droplet. I think it would be nice if when listing the droplets, it shows which droplets have which features enabled. Willing to send a PR for this if it makes sense to do so.

xmudrii:
This PR adds Features list and Volumes list to the Droplet's list output.
Volumes list was there before, but under `isBeta()` flag. As volumes are in beta phase, I removed this and it appeared back in the output.

Fixes #213 

/cc @mauricio @viola

### DIFF
--- a/commands/output.go
+++ b/commands/output.go
@@ -235,10 +235,7 @@ func (d *droplet) JSON(out io.Writer) error {
 
 func (d *droplet) Cols() []string {
 	cols := []string{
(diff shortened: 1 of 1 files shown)


viola:
LGTM!

PR #245, opened 2017-07-19
proposal doctl/cmd: add scp

xmudrii:
This way I'm proposing adding SCP to the `doctl`. If you're interested, I would love to fully develop it and make sure it works both on Linux and Windows. I can't test macOS, but if it works on Linux, it will work on macOS as well.

I plan to do it gradually — first finish Linux support and get it merged if there is interest. Then go on Windows support. 

I believe this is very useful. As we have SSH, I believe adding SCP would be great.
Also Windows `doctl` users would appreciate one more SCP as native one is not available.

Fire up suggestions. I would love to hear from you.

/cc @viola @mauricio @lxfontes


xmudrii:
@viola @mauricio @lxfontes ready for code review.

This implements SCP runner for doctl. It works just like SSH ones work. I would like to know what do you think about this, can this be improved and similar?

Command is `doctl compute scp`, and it's being used like native SCP. For example:
```
doctl compute scp sammy@droplet-name:~/text.txt test.txt
```
or
```
doctl compute scp sammy@123456:~/text.txt test.txt
```
You can also omit username and same it will be populated using default account for Droplet's OS (`root` and `core` for CoreOS). I was thinking to default to your username, but for SSH it's working same as explained so I decided to stick with it.

If you have any question, idea or suggestion, don't hesitate to write down.

mauricio:
Hey @xmudrii, this looks great but it's doesn't seem like something we should provide ourselves, there are easy to use command line options to do the same.

xmudrii:
Hi @mauricio, sounds good to me. I thought as we already have SSH utility, SCP could be great addition so I gave it a try (especially for folks who works with containers, k8s, and so).

Issue #240, opened 2017-07-01
config.yaml file path on Windows


Linda-chan:
Looks like doctl use hardcoded path on user profile folder in its code. So on Windows XP it saves config.yaml in «C:\Users\Crow\AppData\Local\doctl\config\» folder (in my case) instead of «C:\Documents and Settings\MSI U100-455RU\Application Data\doctl\config\». doctl should use APPDATA environment variable or SHGetSpecialFolderLocation() function to get correct config folder location.

xmudrii:
I believe this is the problem - https://github.com/digitalocean/doctl/blob/master/commands/xdg_windows.go#L23

It should use `APPDATA` or `LOCALAPPDATA instead of `XDG_CONFIG_HOME` (AFAIK it doesn't exist on Windows).

However, as this sounds like easy change, we could break stuff for some customers (such as XP users). I think we need someone from DO to discuss it internally and propose way to change. One possibility is to ignore that portion of users, while another one (and better) is to check for legacy path if it doesn't exist in new one. This has been already done in past for Linux users so I believe it shouldn't be problem.

/cc @viola @mauricio @lxfontes

Linda-chan:
I didn't get why changing this

````
userName := os.Getenv("USERNAME")
configHome = filepath.Join("C:/", "Users", userName, "AppData", "Local", "doctl", "config")
````

to this

````
configHome := os.Getenv("APPDATA")
configHome = filepath.Join(configHome, "doctl", "config")
````

should break something. Windows XP users my be legacy users but this kind of approach can break everything even on Windows 10 if for some reason APPDATA folder will be in different location.

xmudrii:
True - I just want to collect opinions from DO employees on this one, it's very easy to change.

Linda-chan:
Ok ^^'

mauricio:
@Linda-chan that makes sense, we'd still have to allow for the old place (if there's a config there) but for new installs use the expected place. `SHGetSpecialFolderLocation` might be a better option since it lets the operating system to provide instead of us trying to figure out folder paths.

I don't have a windows box available here, can you test if we get something like this in place?

Linda-chan:
Okay, I will test it. But I don't have Go environment Oo

xmudrii:
If you want I can test it, I've Windows on my box

mauricio:
@Linda-chan @xmudrii did any of you folks make any progress here?

xmudrii:
[...]

PR #249, opened 2017-08-13
xdg: fix path when XDG_CONFIG_HOME is present

### LINKED ISSUE #246: Default configuration path is inconsistent
I am using version 1.7.0 on Linux.
When ```$XDG_CONFIG_HOME``` is unset, doctl reads the configuration from ```$HOME/.config/doctl/config.yaml```.
But when ```XDG_CONFIG_HOME=$HOME/.config```, doctl reads the configuration from ```$HOME/.config/config.yaml```.
This also happens when generating a configuration using ```doctl auth init```.
I think this can be changed by modifying the file ```doit.go``` however my attempts were unsuccessful.

xmudrii:
Closes #246 

This should fix #246 and make config location more consistent. Should we check is there config in old location as well or we can assume it's not that frequently used?

/cc @viola @mauricio

### DIFF
--- a/commands/xdg.go
+++ b/commands/xdg.go
@@ -22,8 +22,9 @@ import (
 func configHome() string {
 	configHome := os.Getenv("XDG_CONFIG_HOME")
 	if configHome == "" {
-		configHome = filepath.Join(homeDir(), ".config", "doctl")
+		configHome = filepath.Join(homeDir(), ".config")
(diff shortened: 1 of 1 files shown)


xmudrii:
I will send new unified PR that should fix both #246 and #240, as well as refactor logic there.

PR #254, opened 2017-09-11
Fix Releases link 404

ralfonso:
This fixes the link and works no matter where the README file is being displayed.


xmudrii:
Hi and thanks for contribution and for fixing this problem. Changes looks good but somebody from DO with write access will have to push it.

grampajoe:
:+1: Thanks!

PR #259, opened 2017-10-07
Switch from gvt to dep

joonas:
Seems to me that [`dep`](https://github.com/golang/dep) has a lot of momentum and [`gvt`](https://github.com/FiloSottile/gvt) maybe little less, so perhaps we should make the switch now that `dep` is deemed [safe for production use](https://github.com/golang/dep#current-status).

Worth mentioning that this switch was done with the HEAD version of `dep` since it recently added [support for importing gvt manifests](https://github.com/golang/dep/pull/1149).


joonas:
Hmm, seems like Github is somehow confused about the `vendor/manifest` file conflict, going to reopen.

PR #262, opened 2017-10-10
Switch from gvt to dep

joonas:
Reopening #259

Seems to me that [`dep`](https://github.com/golang/dep) has a lot of momentum and [`gvt`](https://github.com/FiloSottile/gvt) maybe little less, so perhaps we should make the switch now that `dep` is deemed [safe for production use](https://github.com/golang/dep#current-status).

Worth mentioning that this switch was done with the HEAD version of `dep` since it recently added [support for importing gvt manifests](https://github.com/golang/dep/pull/1149).


mauricio:
LGTM!

Can you update the readme with that info as well?

joonas:
> LGTM!
> 
> Can you update the readme with that info as well?

Was there something beyond what I have in the [changes for README.md](https://github.com/digitalocean/doctl/pull/262/files#diff-04c6e90faac2675aa89e2176d2eec7d8) that you were thinking of?

mauricio:
Wasn't showing up in the diff for me.

PR #270, opened 2017-10-31
Makefile: enable build via make and create local container

utlemming:
`make` will build local go binary
`make {linux_amd64,linux_386,macos}` will compile for Linux/Mac
`make docker_build` will create a container and copy out Linux and Mac Binaries to `builds`


mauricio:
LGTM!

PR #282, opened 2017-12-14
xdg: fix config path when XDG_CONFIG_HOME is set

### LINKED ISSUE #246: Default configuration path is inconsistent

xmudrii:
Closes #246 

When `XDG_CONFIG_HOME` path is set, `doctl` creates configuration file directly in that directory instead of creating it in the `doctl` subdirectory.

This PR also adds warning if configuration file is present in the root of the `XDG_CONFIG_HOME` directory.

/cc @mauricio

### DIFF
--- a/commands/doit.go
+++ b/commands/doit.go
@@ -130,6 +130,12 @@ func findConfig() (string, error) {
 		warn(msg)
 	}
 
+	legacyXDGPath := filepath.Join(os.Getenv("XDG_CONFIG_HOME"), "config.yaml")
+	if _, err := os.Stat(legacyXDGPath); err == nil {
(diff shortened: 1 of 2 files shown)


mauricio:
Nice!

PR #284, opened 2017-12-18
commands: General simplifications

ferhatelmas:
- in error checking
`if err != nil { return err }; return nil` -> `return err`


mauricio:
Thanks @ferhatelmas !

Issue #291, opened 2018-02-15
doctl compute droplet list - does not show "size slug"


eRadical:
Command `doctl compute droplet list` does not show in output the slug of the size the machine was originally started with.

It should be the same slug from `doctl compute size list`.

If the values for „ Memory    VCPUs    Disk ” do not match it means the machine was resized.

lxfontes:
Hey @eRadical , could you provide an example of what you would like to see?

if scripting, you can get size slugs with `-o json`
```
bin ❯ doctl compute droplet list -o json | jq '.[].size_slug'
"2gb"
"4gb"
"4gb"
1vcpu-1gb"
1vcpu-1gb" 
```

eRadical:
Nice Thanx... I can use that.

But nonetheless for 
```
      --format string     Columns for output in a comma separated list. Possible values: ID,Name,PublicIPv4,PrivateIPv4,PublicIPv6,Memory,VCPUs,Disk,Region,Image,Status,Tags,Features,Volumes
```
Can we also have there an option for SizeSlug ?

mauricio:
Fixed by #292 .

mr-rooftop:
You should also update the help message accordingly. Currently it's still with the `SizeSlug` option:

```
Flags:
      --format ID         Columns for output in a comma-separated list. Possible values: ID, `Name`, `PublicIPv4`, `PrivateIPv4`, `PublicIPv6`, `Memory`, `VCPUs`, `Disk`, `Region`, `Image`, `VPCUUID`, `Status`, `Tags`, `Features`, `Volumes`
```

PR #292, opened 2018-02-15
Adds `SizeSlug` to format fields

lxfontes:
`size_slug` is present on json but not present on plain text output.

This adds `Size Slug` as an output column.
However, I have *not* added it to `droplet list` due to the amount of info already being displayed.

Related #291


viola:
LGTM on 🍏

mauricio:
LGTM

xmudrii:
:green_apple: :green_heart:

PR #296, opened 2018-03-21
Confirm dialog for deleting by ID now specifies number of droplets to…

justinbeal:
… be deleted.

Problem: I commonly nest `doctl compute droplet list` commands within `doctl compute droplet delete` to delete droplets by name. It would be a nice last-minute sanity check to see the length of the ID's generated by the list command in the confirm dialog.


mauricio:
:green_heart:

Issue #266, opened 2017-10-25
Feature Request: Multiple Accounts


benyanke:
I am not sure what it would look like for the command syntax, but multiple accounts would be an excellent feature.

Personally, I set up all my clients on their own DO account, for billing purposes, so doctl is borderline unusable for quick tasks.

I see two simple options for the syntax and implmenetation (there may be more):

1) add another auth command with shortcuts - `doctl auth switch [accountnickname]` switches the context to the other account. This is backward compatible. 

2) The breaking change option would be specifying the account nickname in every command. Also more stateless though.

xmudrii:
@benyanke In the meanwhile.. there should be the `--access-token` flag which can help you.
However, that would require you to store tokens somewhere accessible (yet secure). Maybe you can use environmental variables as a workaround solution.

eRadical:
On AWS the aws-cli uses a simple file to store only credentials:
```
[eRadical@hp-g3-acasa ~]$ cat ~/.aws/credentials 

[default]
aws_access_key_id = default_AK
aws_secret_access_key = default_SAK

[demo1]
aws_access_key_id = X1
aws_secret_access_key = Y1

[demo2]
aws_access_key_id = X2
aws_secret_access_key = Y2
```

Then in command line you can do:

```
aws --profile=demo2 [....]
```

If you don't provide a profile it goes to the default one if it exists otherwise it errors out.

It would be lovely to do it like that... as I read the credentials file from ansible and use the keys from those files. This way credentials are in one place only and used by  doctl and ansible code.

ReggaePanda:
a `doctl auth switch [accountnickname]` will be a nice feature (also for teams) is this on the roadmap or can I place somewhere a feature request?

mauricio:
It is, we should have it available soon :)

jrm780:
This feature is now available in doctl v1.8.0 :) thanks @kamaln7 !

PR #299, opened 2018-03-29
Implement context switching, allowing for multiple configured API access keys

kamaln7:
Related issue: #266 

This is a proposed implementation of support for multiple accounts. The idea behind this is each account is associated with a context, which contains the API access key.

A new global option is added: `context`. It defaults to `default`, and should contain the name of the DO account that will be used.

If it is set to `default`, all existing auth functionality is preserved (using `--access-token` or the corresponding environment variable). This makes this feature backwards-compatible.

To configure a "named" account called, say, `foo`, run `doctl auth init --context foo`. This will prompt the user for the API key and save it in the config file. Whenever `--context foo` is passed to any other command, that API key will be used.

[This suggestion](https://github.com/digitalocean/doctl/issues/266#issuecomment-372329792) could also be added with the current implementation. It would simply update the `context` config value.

@mauricio please let me know what you think and if this is what you had in mind for how this would work.

### DIFF
--- a/args.go
+++ b/args.go
@@ -16,6 +16,8 @@ package doctl
 const (
 	// ArgAccessToken is the access token to be used for the operations
 	ArgAccessToken = "access-token"
+	// ArgContext is the name of the auth context to use
+	ArgContext = "context"
 	// ArgActionID is an action id argument.
 	ArgActionID = "action-id"
 	// ArgActionAfter is an action after argument.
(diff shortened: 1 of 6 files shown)


jrm780:
Thank you for your contribution!

Generally, I would expect a flag like `--context` to have an effect only when it is supplied to the command and have no effect on subsequent calls without the flag. It would be best to change the implementation so the config value would only be written to if `doctl auth switch [context name]` is run.

kamaln7:
Thanks for the feedback @jrm780! I've updated the flag to work as you suggested, with a new command `doctl auth switch` that writes the current context (so it can be passed using an env var or `--context`) permanently to the config file.

PR #301, opened 2018-04-11
Mark completion command as hidden

### LINKED ISSUE #297: Remove `completion` from auto completion
When typing out `doctl comp` and hitting tab `compute` does not immediately complete because `completion` is also an option. I think it's safe to not expect anyone to need to auto complete `completion`.

image

eddiezane:
Closes #297 

Ref https://github.com/spf13/cobra/blob/cd30c2a7e91a1d63fd9a0027accf18a681e9d50b/command.go#L74


mauricio:
Awesome!

PR #303, opened 2018-04-12
cmd: add --no-header and --format flags for firewall list command

xmudrii:
This PR adds the `displayerType` attribute for the relevant `firewall` commands, ensuring they have the `--no-header` and `--format` flags. Those flags can be used to format output, which helps if you want to automate some processes.


mauricio:
Great stuff!

PR #304, opened 2018-04-12
cmd: allow multiple firewalls to be deleted in same time

xmudrii:
Currently, to delete multiple firewalls, you need to execute several commands to delete them, as one command can delete only one firewall. This PR allows users to delete multiple firewalls using single `firewall delete` command, like they can do for Droplets.

With #303 merged, users can also delete all firewalls in the same time by chaining `delete` and `list` commands, e.g. `doctl compute firewall delete -f $(doctl compute firewall list --no-header --format ID)`.


mauricio:
\o/

Issue #238, opened 2017-06-11
man pages and fish completions


igalic:
the fish shell allows to easily generates its completions from the man pages on the system.
while doctl provides a rich inline documentation, it's not persisted on disk, and as such, unfortunately,  not quite useful to fish.

Would you consider providing (an export option for) man pages?

xmudrii:
@igalic this actually sounds good, :+1: for idea. I was looking into fish completion, but due to lack of the free time, I was not able to finish it.

Right now, I found a way to do this. We should let user generate man pages from the doctl, just like they can generate Bash/ZSH completion. I didn't know that was possible, but I found a doc in [spf13/cobra](https://github.com/spf13/cobra/blob/master/doc/man_docs.md) that actually shows code snippet.
User needs to install it manually, but we will provide instructions, that seems easy.

If you are interested into this, try playing around with cobra-way for generating man pages. I think this would be best way. If somebody has better idea, chime in 👍 
I would love to give it a try, but I've got a lot of work so that's impossible now.

frbrz:
Sorry to bother but do you guys have any news on this?
Is there something I could do to have man pages on my system?

bentranter:
Hey @frbrz,

We’re working on improving the docs in https://github.com/digitalocean/doctl/pull/658. We may or may not ship man pages as a result of that, but are actively looking into it. I'll keep this issue updated with the progress on man pages.

PR #305, opened 2018-04-12
move completion into a new 'generate' namespace

kamaln7:
This creates a new "namespace" called `generate` and moves the `completion` command (and its subcommands) under it. This potentially takes care of #297 and makes room for #238


kamaln7:
I'm not sure if we want this, but I'd need to update the README if so

xmudrii:
@kamaln7 [Cobra can generate man pages as well, so maybe we can utilize it like we do for Bash/Zsh completions](https://github.com/spf13/cobra/blob/master/doc/man_docs.md).

PR #315, opened 2018-05-10
Remove upgrade command from droplet-action

tfogo:
Hi, the upgrade droplet action is no longer supported by the v2 API. (See the [droplet-actions endpoint in the docs](https://developers.digitalocean.com/documentation/v2/#droplet-actions))

Running `doctl compute droplet-action upgrade <droplet-id>` results in:

```
Error: POST https://api.digitalocean.com/v2/droplets/<droplet-id>/actions: 410 The specified action type is no longer available.
```

This PR removes the `upgrade` subcommand from `droplet-action` as well as associated tests and mocks.

Let me know if it looks good or if there's anything else I need to do.


mauricio:
Thanks!

PR #319, opened 2018-05-25
Update dependencies in preparation for adding volume format support

adamwg:
A few vendoring changes here in preparation for adding volume format support to doctl:

1. Set versions for `github.com/spf13/viper` and `github.com/spf13/cast`. The latter appears to be necessary even though `cast` is a transitive dependency, as dep doesn't seem to update it otherwise and the version we have is too old for viper to build.
2. Remove `github.com/davegch/go-spew` from `Gopkg.toml`, as it's a transitive dependency (via testify/assert) and dep will pull it in automatically.
3. Finally, with the dep stuff in a consistent state, update godo to v1.3.0 to get support for volume formatting. This will be used in a follow-on change.

Note that I'd be happy to drop the first two changes, but I haven't been able to get `dep ensure` to update godo without updating everything else, and updating viper breaks the build as noted in (1).


mauricio:
All good here, merge away!

pdonorio:
Can I ask if this means we will be able to format volumes with `doctl`?

adamwg:
@pdonorio We don't support formatting existing volumes, but you can now create new formatted volumes via `doctl`. See the description of the `filesystem_type` attribute in the [DO API documentation](https://developers.digitalocean.com/documentation/v2/#create-a-new-block-storage-volume) for details.

pdonorio:
Thanks! I've tried it and failed.
It seems I'm using `1.8.0`, there is a new `1.8.2` but brew won't update on my mac.

adamwg:
@pdonorio The brew recipe has not been updated yet. I'll do that shortly. In the meantime, you can download a 1.8.2 binary from the [doctl releases page](https://github.com/digitalocean/doctl/releases).

pdonorio:
Ah thank you!
Downloaded the `darwin` release, I'm `1.8.2` now.

But:
```bash
$ doctl compute volume create --size 10GB --region lon1 --fs-type xfs test
Error: strconv.ParseFloat: parsing "": invalid syntax
```

Not sure where the problem is, any hint? 🤔 Sorry to bother

adamwg:
@pdonorio Thanks for the report. I can reproduce the problem. It doesn't appear to be related to volume formatting, since I can reproduce it by trying to create an unformatted volume as well. Will investigate.

adamwg:
@pdonorio Issue fixed in v1.8.3. Thanks again for reporting the problem!

pdonorio:
I've just tested it, works like a charm :D

PR #323, opened 2018-06-12
volumes: Add support for formatted volumes

adamwg:
Allow creation of formatted volumes, and show volume format information when displaying volumes.


mauricio:
Can you add the changelog entry for this change?

LGTM otherwise.

PR #326, opened 2018-06-13
Revert "vendor: Set versions for viper and cast, remove go-spew"

### LINKED ISSUE #325: Volume creation fails with `Error: strconv.ParseFloat: parsing "": invalid syntax`
```console
$ doctl compute volume create --size "10GiB" --region lon1 doctl-test-volume
create a volume

Usage:
  doctl compute volume create <volume-name> [flags]

Aliases:
  create, c


Flags:
      --desc string       Volume description
      --format string     Columns for output in a comma separated list. Possible values: ID,Name,Size,Region,Filesystem Type,Filesystem Label,Droplet IDs
      --fs-label string   Volume filesystem label
      --fs-type string    Volume filesystem type (ext4 or xfs)
      --no-header         hide headers
      --region string     Volume region (required)
      --size string       Volume size (required) (default "4TiB")

Global Flags:
  -t, --access-token string   API V2 Access Token
  -u, --api-url string        Override default API V2 endpoint
  -c, --config string         config file (default is $HOME/.config/doctl/config.yaml)
      --context string        authentication context name
  -o, --output string         output format [text|json] (default "text")
      --trace                 trace api access
  -v, --verbose               verbose output

Error: strconv.ParseFloat: parsing "": invalid syntax
```

This appears to be a fairly recent regression. Will test a few versions to see where it was introduced.

adamwg:
This reverts commit 56d9f3222f1e3b3498f53feb608293484da81cc5.

The updated versions of viper and cobra have broken required arguments entirely, and it's not clear what the issue is. We didn't actually need/want new versions of these deps to start with, but `dep` forced them on us. Go back to the versions we were using before, which work fine.

This will fix #325.


mauricio:
LGTM, merge away.

PR #331, opened 2018-07-05
Validate type and confirm

jaxxstorm:
When creating a certificate, if you specify certificate of type `lets_encrypt` you get the following error:

```
doctl compute certificate create --type lets_encrypt --name test --dns-names "testing.example.net"
```
```
Error: (certificate.create.private-key-path) command is missing required arguments
```

You shouldn't have to specify the private key, leaf certificate or any other string.

I've wrapped these calls in an if statement, if anyone can tell me a better way I'm happy to oblige.

### DIFF
--- a/commands/certificates.go
+++ b/commands/certificates.go
@@ -92,46 +92,48 @@ func RunCertificateCreate(c *CmdConfig) error {
 		Type:     cType,
 	}
 
-	pkPath, err := c.Doit.GetString(c.NS, doctl.ArgPrivateKeyPath)
-	if err != nil {
-		return err
-	}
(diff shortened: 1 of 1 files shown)


zachgersh:
Hey @jaxxstorm - thanks a ton for the PR. Have you run into this recently? I've currently got a test here that mimics your behavior but doesn't get the same result:

https://github.com/digitalocean/doctl/blob/master/integration/certificate_create_test.go

Also, really sorry for the delay here.

jaxxstorm:
There's every possibility this is no longer a problem, so closing this

PR #343, opened 2018-08-27
Add Spaces CDN commands

sunny-b:
* Adds Spaces CDN commands to cobra/viper
* Updates `doctl` version to 1.9.0 and ChangeLog
* Fixes checks for empty int string and empty string slice flags.

cc: @mauricio


commands/cdns.go
```diff
@@ -0,0 +1,173 @@
...
+	origin := c.Args[0]
+
+	ttl, err := c.Doit.GetInt(c.NS, doctl.ArgCDNTTL)
+	if err != nil {
+		return err
+	}
+
+	createCDN := &godo.CDNCreateRequest{
+		Origin: origin,
+		TTL:    uint32(ttl),
```
jrm780: Is there an upper-bound on the ttl server-side? Just wondering what happens here if `ttl` is negative then cast to `uint32`.

commands/cdns.go
```diff
@@ -0,0 +1,173 @@
...
+		displayerType(&cdn{}))
+
+	cmdCDNUpdate := CmdBuilder(cmd, RunCDNUpdateTTL, "update <cdn-id>", "update a cdn", Writer,
+		aliasOpt("u"), displayerType(&cdn{}))
+	AddIntFlag(cmdCDNUpdate, doctl.ArgCDNTTL, "", 0, "cdn ttl",
+		requiredOpt())
+
+	cmdCDNFlushCache := CmdBuilder(cmd, RunCDNFlushCache, "flush <cdn-id>", "flush cdn cache", Writer,
+		aliasOpt("fc"))
+	AddStringSliceFlag(cmdCDNFlushCache, doctl.ArgCDNFiles, "", []string{}, "cdn files",
```
adamwg: Is the empty slice a valid default here? The `godo` docs indicate that the argument should be non-empty.

sunny-b: You're correct, it's not a valid choice. The endpoint will return an error. But the `--files` flag is a required flag, so the default should not be called since a `doctl` error will be raised if the user doesn't include the `--files` flag.

commands/cdns.go
```diff
@@ -0,0 +1,173 @@
...
+	if len(c.Args) == 0 {
+		return doctl.NewMissingArgsErr(c.NS)
+	}
+
+	force, err := c.Doit.GetBool(c.NS, doctl.ArgForce)
+	if err != nil {
+		return err
+	}
+
+	if force || AskForConfirm("delete volume") == nil {
```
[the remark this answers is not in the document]
sunny-b: Good catch

commands/cdns.go
adamwg: Ah, right, I missed that it's required. Given that wildcards are allowed `[]string{"*"}` might be a reasonable default, but keeping it required makes sense too.

commands/cdns.go
sunny-b: I just tested it and the endpoint returns a 500 error if the user tries to submit a negative ttl. Most likely because negative numbers cause `uint32` to overflow and return an error. https://play.golang.org/p/VHHvC9BDnpI

I'll add a check to return a `doctl` error if the ttl passed in is negative

commands/cdns.go
sunny-b: Yeah I thought about that and had it as the wildcard at first, I was just wary of assuming the User's intention, though flushing the cache isn't the end of the world. I agree that `*` would make a good default. I'll update that.

PR #349, opened 2018-10-02
Expose 'tag-name' argument to the RunDropletDelete command.

timoreimann:
The functionality was already implemented in `RunDropletDelete` but never exposed. With this change, we allow deletion of droplets by tag name through the CLI.

Ideally, I'd also like to add a test case. However, from a quick glance at `droplet_test.go` I cannot easily see how to do that -- the existing tests (like `TestDropletDeleteByTag`) seems to attach the tag name argument separately from the `Droplet()` function. It looks like we should be testing things differently to actually run through the system-under-test, though chances are I might be missing something. Pointers welcome. :)

### DIFF
--- a/commands/droplets.go
+++ b/commands/droplets.go
@@ -478,7 +478,7 @@ func RunDropletDelete(c *CmdConfig) error {
 			fmt.Fprintf(c.Out, fmt.Sprintf("nothing to delete: no droplets found by \"%s\" tag\n", tagName))
 			return nil
 		}
-		var ids []string
+		ids := make([]string, 0, len(list))
 		for _, droplet := range list {
 			ids = append(ids, strconv.Itoa(droplet.ID))
 		}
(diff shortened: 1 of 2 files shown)


aybabtme:
lgtm

eddiezane:
Thank you for the PR!

I think we should prevent the user from shooting themselves in the foot by listing the droplets that will be deleted in the confirmation prompt.

Thoughts?

timoreimann:
@eddiezane sounds like a great idea to me. Will get back to you as soon as I manage to extend the PR accordingly.

timoreimann:
Pushed the suggested change. PTAL again.

eddiezane:
@timoreimann thanks for the PR!

Would you please shoot me an email when you get a chance? My GH username [at] digitalocean [dot] com.

PR #352, opened 2018-10-02
Ralfonso patch 1

BuddhikaShaminda:


mauricio:
Not sure what this is supposed to be.

PR #354, opened 2018-10-17
volumes: Fix droplet ID display when listing volumes

adamwg:
Due to a missing space in the droplet ID column name, the `Droplet IDs` column in the output of `doctl compute volumes ls` was always showing `<nil>`, even for attached volumes.

Add the necessary space so that droplet IDs display again.

### DIFF
--- a/commands/displayers/volume.go
+++ b/commands/displayers/volume.go
@@ -34,7 +34,7 @@ func (a *Volume) JSON(out io.Writer) error {
 
 func (a *Volume) Cols() []string {
 	return []string{
-		"ID", "Name", "Size", "Region", "Filesystem Type", "Filesystem Label", "Droplet IDs",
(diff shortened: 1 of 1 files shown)


senechko:
@adamwg i'm curious if the more desirable change here is to update line 49?
https://github.com/digitalocean/doctl/blob/master/commands/displayers/volume.go#L49
to 
```
		"DropletIDs":      "Droplet IDs",
```

prior art here:
https://github.com/digitalocean/doctl/blob/master/commands/displayers/firewall.go#L57
and
https://github.com/digitalocean/doctl/blob/master/commands/displayers/load_balancer.go#L62

adamwg:
@senechko Sure, that works as well - done. It's a two-line change either way, but the way you described is indeed more consistent with other displayers.

PR #356, opened 2018-10-17
fix typo in test filename

MorrisLaw:
The filename had a typo and spelled `command` as `commmand`. I've updated the filename is this PR.


senechko:
@MorrisLaw do you mind updating the name of the function as well? `TestMockCommmand_Run` -> `TestMockCommand_Run`

PR #362, opened 2018-11-08
Default to showing public images for image list command

bengadbois:
Because the other `image list-application` and `image list-distribution` commands display public images by default, it's confusing that `image list` (which logically should have everything present in the others) doesn't show the public images.


aybabtme:
Yeah this will probably break scripts.

roypaulandrachel (approved):

ChiefMateStarbuck:
Closed to stale input.

PR #364, opened 2018-11-13
add godo to doctl adapter

aybabtme:
This is a dumb wrapper that takes `godo`'s types and make them nicer to consume inside this project. This implementation follows how all the other services are done.


nanzhong:
lgtm

PR #365, opened 2018-11-14
k8saas: basic k8s support

aybabtme:
First stab at K8S support, behind a beta flag (`DIGITALOCEAN_ENABLE_BETA=true`).


nanzhong:
lgtm, been testing with it for the past few days and so far so good. Let's 🚢 and iterate if needed.

PR #372, opened 2018-11-28
k8saas: better kubeconfig support

aybabtme:
- Adds/updates/deletes cluster credentials from a user's local kubeconfig.
- Automatically sets the created cluster as the default context for the user.
- Provide facilities to print/save/remove a cluster from a user's kubeconfig.
- Allow waiting for a cluster to be `running` on a `create` call.


commands/kubernetes.go
```diff
@@ -222,14 +250,43 @@ func RunKubernetesUpdate(c *CmdConfig) error {
...
+		if tries >= maxAPIFailures {
+			warn(fmt.Sprintf("couldn't get credentials for cluster, it will not be added to your kubeconfig: %v", err))
+			return
+		}
+		time.Sleep(2 * time.Second)
+	}
+	if err := writeOrAddToKubeconfig(kubeconfig); err != nil {
+		warn(fmt.Sprintf("couldn't write cluster credentials: %v", err))
+	}
+}
```
fatih: Why don't you return an error and print it here instead ?

aybabtme: because this isn't a reason to fail the whole operation and not print the cluster that was just created, it would confuse users into thinking their cluster wasn't created

commands/kubernetes.go
```diff
@@ -579,6 +711,121 @@ func buildNodePoolUpdateRequestFromArgs(c *CmdConfig, r *godo.KubernetesNodePool
...
+			if failCount >= maxAPIFailures {
+				return err
+			}
+			// tolerate transient API failures
+			time.Sleep(time.Second)
+		} else {
+			failCount = 0 // API responded, reset it's error counter
+		}
+		if cluster.Status == nil {
+			time.Sleep(1 * time.Second)
```
bouk: Should do a `fmt.Print(".")` here for every second, and then a `Println` at the end.

commands/kubernetes.go
```diff
@@ -579,6 +711,121 @@ func buildNodePoolUpdateRequestFromArgs(c *CmdConfig, r *godo.KubernetesNodePool
...
+	delete(local.AuthInfos, remoteCtx.AuthInfo)
+	if local.CurrentContext == remote.CurrentContext {
+		local.CurrentContext = ""
+	}
+	return nil
+}
+
+// waitForClusterRunning waits for a cluster to be running.
+func waitForClusterRunning(kube do.KubernetesService, clusterID string) error {
+	failCount := 0
```
bouk: Printing out that the command is doing something here would be good. Right now it's just not showing anything while it's waiting

commands/kubernetes.go
```diff
@@ -63,61 +71,101 @@ func Kubernetes() *Command {
...
+
+	cmdKubeClusterCreate := CmdBuilder(cmd, RunKubernetesClusterCreate(defaultNodeSize, defaultNodeCount), "create <name>", "create a cluster", Writer, aliasOpt("c"))
+	AddStringFlag(cmdKubeClusterCreate, doctl.ArgRegionSlug, "", defaultRegion, "cluster region location, example value: nyc1", requiredOpt())
+	AddStringFlag(cmdKubeClusterCreate, doctl.ArgClusterVersionSlug, "", "", "cluster version")
+	AddStringSliceFlag(cmdKubeClusterCreate, doctl.ArgTagNames, "", nil, "cluster tags")
+	AddStringFlag(cmdKubeClusterCreate, doctl.ArgSizeSlug, "", defaultNodeSize, "size of the nodes in the default node pool (incompatible with --"+doctl.ArgClusterNodePool+")")
+	AddStringFlag(cmdKubeClusterCreate, doctl.ArgNodePoolCount, "", strconv.Itoa(defaultNodeCount), "size of the nodes in the default node pool (incompatible with --"+doctl.ArgClusterNodePool+")")
+	AddStringSliceFlag(cmdKubeClusterCreate, doctl.ArgClusterNodePool, "", nil, `cluster node pools in the form "name=your-name;size=droplet_size;count=5;tag=tag1;tag=tag2"`, requiredOpt())
+	AddBoolFlag(cmdKubeClusterCreate, doctl.ArgClusterUpdateKubeconfig, "", true, "whether to add the created cluster to your kubeconfig")
```
nanzhong: This default also changes the current context to the new cluster cluster, should this maybe be a separate value that defaults to false? I've had a few occasions where this happened and I forgot about it, and got very confused why my kubectl seemed to not be doing what I wanted it to the cluster that was previously my active context.

hilary:
Looks (and feels) ready to me.

commands/kubernetes.go
alexellis: Is there a way you could make the node-pool flag more usable or show how to print the various "slug" sizes in the help text?

alexellis:
> Automatically sets the created cluster as the default context for the user.

I know minikube does this, but is there a flag to preserve context? When creating clusters with eksctl for example I'm not forced to get a context-switch.

aybabtme:
@alexellis I'll add one, and change the default behavior to _not_ change the context. It seems that enough people find this jarring to suggest it's probably not a good default behavior.

aybabtme:
actually I'll just remove that changing the context thing, it's already easily done with `kubectl`.

commands/kubernetes.go
aybabtme: Yeah I'll try to make this better in a next iteration.

PR #373, opened 2018-11-28
k8saas: vendor deps

aybabtme:
Vendor the deps separately to make #372 smaller.


fatih:
lgtm I would prefer to squash them into a single commit before merging it.

PR #377, opened 2018-12-05
bump godo version

aybabtme:


aybabtme:
agree it would have been better, i'll do it next time tho. the commits are separated so history wise it should be alright.

PR #383, opened 2018-12-08
k8s: default to usable settings

aybabtme:
The default node size slug baked in the binary isn't valid for Kubernetes clusters. Fix that, and improve the help strings a bit while at it.


aybabtme:
Should probably be done, but this would make the k8s code look drastically different from the rest. I'd say this is better done in a separate PR that changes it across all commands.

PR #387, opened 2018-12-19
k8s: use ExecCredential for authentication

bouk:
This changes the behavior of saving kubeconfig locally to set up an [exec plugin](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins) in the kubeconfig, which gets called to talk to our API and retrieve credentials to authenticate against the Kubernetes server.

It caches the credentials under `~/.config/doctl/cache/kubeconfig/<uuid>.json`

### DIFF
--- a/commands/kubernetes.go
+++ b/commands/kubernetes.go
@@ -16,10 +16,12 @@ package commands
 import (
 	"context"
 	"crypto/x509"
+	"encoding/json"
 	"encoding/pem"
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"sort"
 	"strconv"
 	"strings"
@@ -33,6 +35,8 @@ import (
 	"github.com/pborman/uuid"
 	"github.com/spf13/cobra"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientauthentication "k8s.io/client-go/pkg/apis/clientauthentication/v1beta1"
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 )
@@ -144,9 +148,15 @@ func kubernetesKubeconfig() *Command {
 	}
 
 	CmdBuilder(cmd, RunKubernetesKubeconfigShow, "show <cluster-id|cluster-name>", "show a cluster's kubeconfig to standard out", Writer, aliasOpt("p", "g"))
+	cmdExecCredential := CmdBuilder(cmd, RunKubernetesKubeconfigExecCredential, "exec-credential <cluster-id>", "INTERNAL print a cluster's exec credential", Writer, hiddenCmd())
+	AddStringFlag(cmdExecCredential, doctl.ArgVersion, "", "", "")
 	CmdBuilder(cmd, RunKubernetesKubeconfigSave, "save <cluster-id|cluster-name>", "save a cluster's credentials to your local kubeconfig", Writer, aliasOpt("s"))
 	CmdBuilder(cmd, RunKubernetesKubeconfigRemove, "remove <cluster-id|cluster-name>", "remove a cluster's credentials from your local kubeconfig", Writer, aliasOpt("d", "rm"))
 	return cmd
+}
+
+func kubeconfigCachePath() string {
+	return filepath.Join(configHome(), "cache", "exec-credential")
 }
 
 func kubernetesNodePools() *Command {
@@ -322,7 +332,7 @@ func tryUpdateKubeconfig(kube do.KubernetesService, clusterID, clusterName strin
 			break
 		}
 	}
-	if err := writeOrAddToKubeconfig(clusterName, kubeconfig); err != nil {
+	if err := writeOrAddToKubeconfig(clusterID, kubeconfig); err != nil {
 		warn("couldn't write cluster credentials: %v", err)
 	}
 }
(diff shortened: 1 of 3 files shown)


bouk:
If you want to give it a shot locally, just check out my branch, run `go get github.com/digitalocean/doctl/cmd/doctl` and then `doctl k8s cluster create ...` or `doctl k8s kubeconfig save ...`

commands/kubernetes.go
```diff
@@ -144,11 +148,17 @@ func kubernetesKubeconfig() *Command {
 	}
 
 	CmdBuilder(cmd, RunKubernetesKubeconfigShow, "show <cluster-id|cluster-name>", "show a cluster's kubeconfig to standard out", Writer, aliasOpt("p", "g"))
+	cmdExecCredential := CmdBuilder(cmd, RunKubernetesKubeconfigExecCredential, "exec-credential <cluster-id>", "INTERNAL print a cluster's exec credential", Writer, hiddenCmd())
+	AddStringFlag(cmdExecCredential, doctl.ArgVersion, "", "", "")
```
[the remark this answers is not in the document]
bouk: Yeah it's internal and hidden, and I want the caller to be explicit about what version so we can support multiple versions when new versions of the API are released.

commands/kubernetes.go
```diff
@@ -395,22 +405,169 @@ func RunKubernetesKubeconfigShow(c *CmdConfig) error {
...
+	t := execCredential.Status.ExpirationTimestamp
+	if t.IsZero() || t.Time.Before(time.Now()) {
+		if err := os.Remove(path); err != nil && Verbose {
+			fmt.Fprintln(os.Stderr, err)
+		}
+		return nil
+	}
+
+	return execCredential
+}
```
fatih: I would return an error in this function and log it where you call it and still continue. It's more maintainable in my opionion

fatih: like with ` -debug` or `-verbose` flag. I'm ok if you don't log it in this PR, but we should be able to do it later.

commands/kubernetes.go
```diff
@@ -395,22 +405,169 @@ func RunKubernetesKubeconfigShow(c *CmdConfig) error {
...
+	}
+
+	defer f.Close()
+
+	var execCredential *clientauthentication.ExecCredential
+	if err := json.NewDecoder(f).Decode(&execCredential); err != nil {
+		if Verbose {
+			fmt.Fprintln(os.Stderr, err)
+		}
+		return nil
```
fatih: I would return here (see comment below) and log it where you call it

commands/kubernetes.go
```diff
@@ -395,22 +405,169 @@ func RunKubernetesKubeconfigShow(c *CmdConfig) error {
...
+	}
+	defer f.Close()
+
+	if err := json.NewEncoder(f).Encode(execCredential); err != nil && Verbose {
+		fmt.Fprintln(os.Stderr, err)
+	}
+}
+
+// RunKubernetesKubeconfigExecCredential displays the exec credential. It is for internal use only.
+func RunKubernetesKubeconfigExecCredential(c *CmdConfig) error {
```
fatih: Why is this exported if it's for internal use only? I would unexport it.

commands/kubernetes.go
```diff
@@ -395,22 +405,169 @@ func RunKubernetesKubeconfigShow(c *CmdConfig) error {
...
+	cacheExecCredential(clusterID, execCredential)
+
+	return json.NewEncoder(c.Out).Encode(execCredential)
+}
+
+func execCredentialFromConfig(config *clientcmdapi.Config) (*clientauthentication.ExecCredential, error) {
+	current := config.CurrentContext
+	context, ok := config.Contexts[current]
+	if !ok {
+		return nil, fmt.Errorf("received invalid config from API")
```
fatih: Error return should be fixed. `fmt.Errorf("context %q doesn't exist", current)`

commands/kubernetes.go
```diff
@@ -395,22 +405,169 @@ func RunKubernetesKubeconfigShow(c *CmdConfig) error {
...
+	}
+
+	authInfo, ok := config.AuthInfos[context.AuthInfo]
+	if !ok {
+		return nil, fmt.Errorf("received invalid config from API")
```
fatih: Error return should be fixed. `fmt.Errorf("auth info for context %q doesn't exist", contextName)`

commands/kubernetes.go
bouk: The CLI command is for internal use, the function is exported because the others are too. I don't feel strongly either way.

commands/kubernetes.go
bouk: This error is bubbled up to the user, so maybe it should be something user friendly telling the user to file a ticket

fatih: yeah but how do you know which error it is? Let treat our users more smartly, especially a Kubernetes user would understand what a "missing context" is.

bouk: No, they would have no idea what to do with it, since the context is missing from the config that is returned by out API, it's not a problem with the configuration of the user. The file that is interpreted here is the kubeconfig that our API provides

So if it can't find the context in this case it means that our API returned a broken kubeconfig

commands/kubernetes.go
bouk: good call

commands/kubernetes.go
fatih: gotcha. how do you gonna know during debugging what is wrong? Suppose the user gets the error, is the authinfo missing or there is no context? I think we should at least add something like this: `received invalid config from API (code - 1)` or something alike so we know what it means

bouk:
@fatih I've improved the errors, please take another look

fatih:
lgtm :+1:

PR #392, opened 2019-01-09
Simplify newline trimming in retrieveUserInput.

timoreimann:
Small readability improvement.

### DIFF
--- a/commands/confirmation.go
+++ b/commands/confirmation.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"bufio"
 	"fmt"
+	"io"
 	"os"
 	"strings"
 )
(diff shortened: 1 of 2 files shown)


eddiezane:
@timoreimann thanks for the PR!

I think this one is intentional because Windows uses a different line ending than *nix.

http://www.cs.toronto.edu/~krueger/csc209h/tut/line-endings.html

Thoughts?

timoreimann:
@eddiezane thanks for the link, I had a look. AFAICS, my PR should not change the behavior of the existing code. We previously ran two replacements to get rid of any potential carriage returns in either operating system. With my PR, it should be just one. (It might be good to note that [`TrimRight`](https://golang.org/pkg/strings/#TrimRight) takes a cutset with _any_ contained characters being removed from the tail of the string).

To be absolutely certain, I pushed a unit test for the logic in question (which required a slight refactoring for the input to become injectable) and rearranged it before the new logic history-wise to verify that the behavior doesn't change. The tests stay green with both commits for me.

Let me know if that makes sense or whether I'm missing something.

hilary:
@timoreimann Thanks, nice contribution!

PR #401, opened 2019-01-28
kubernetes: Fetch credentials after cluster is provisioned

### LINKED ISSUE #399: fetching of Kubernetes credentials / configuration fails occasionally with a "500 Server Error"
When you create a Kubernetes cluster, doctl fetches the configuration and integrates it into the user kubeconfig file. I noticed that this operation is failing occasionally with a HTTP 500 error. Example:

```
$ doctl kubernetes cluster create test-cluster --count 1 --region fra1 --version 1.13.1-do.2
Notice: cluster created, fetching credentials

Warning: couldn't get credentials for cluster, it will not be added to your kubeconfig: GET https://api.digitalocean.com/v2/kubernetes/clusters/059873d7-a92c-447b-99f7-xxxxxxxxxxxx/kubeconfig: 500 Server Error
```

If I were to put this into numbers, I'd say that the failure happens to me at 10 to 15 percent of my cluster creation invocations.

FWIW, I tend to use the FRA1 region.

bouk:
We currently try to fetch the credentials before the cluster is provisioned, which can lead to a 404 if the credentials aren't ready yet. This changes the order of operations so that the credentials are fetched after the cluster is done provisioning.

Fixes #399

### DIFF
--- a/commands/kubernetes.go
+++ b/commands/kubernetes.go
@@ -261,17 +261,17 @@ func RunKubernetesClusterCreate(defaultNodeSize string, defaultNodeCount int) fu
 			return err
 		}
 
-		if update {
-			notice("cluster created, fetching credentials")
-			tryUpdateKubeconfig(kube, cluster.ID, clusterName)
-		}
-
 		if wait {
 			notice("cluster is provisioning, waiting for cluster to be running")
 			cluster, err = waitForClusterRunning(kube, cluster.ID)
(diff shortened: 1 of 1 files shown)


timoreimann:
Out of curiosity / book-keeping reasons: does this one address #399?

bouk:
Yes it does!

PR #404, opened 2019-01-30
Fix typo in Makefile

aude:
Ain't got no phones in `make` :D


andrewsomething:
:man_facepalming: Thanks @aude!

PR #412, opened 2019-03-03
Fix multiple linting issues in doctl

tariq1890:
i) govet - unkeyed fields
ii) golint - redundant err !-nil checks
iii) staticcheck - improper use of printf-style statements when there are no arguments used.
iv) removed use of oauth.NoContext which is deprecated.
v) run gofmt -s -w

### DIFF
--- a/commands/displayers/certificate.go
+++ b/commands/displayers/certificate.go
@@ -14,7 +14,6 @@ limitations under the License.
 package displayers
 
 import (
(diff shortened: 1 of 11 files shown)


tariq1890:
@bouk @timoreimann Requesting your attention on this.

bouk:
Thanks for your contribution!

PR #414, opened 2019-03-05
doctl: clean up out of sync vendor deps

xornivore:
Clean up vendor deps so that latest `dep ensure` diff is clean.


mjsteger:
LGTM

PR #423, opened 2019-04-08
bump version and update changelog for databases support

sunny-b:
This updates the doctl `version` to `1.15.0` and updates the CHANGELOG to include the addition of Managed Databases.

cc: @hilary


hilary:
:shipit:

PR #426, opened 2019-04-16
kubernetes: add maintenance policy flags

fatih:
Create:

```
doctl kubernetes cluster create foo --maintenance-window monday=00:00
```

Update:

```
doctl kubernetes cluster update foo --maintenance-window any=12:00
```

There is no deletion of a maintenance policy as it's not needed.

Depends on https://github.com/digitalocean/godo/pull/224

### DIFF
--- a/args.go
+++ b/args.go
@@ -46,6 +46,8 @@ const (
 	ArgNodePoolCount = "count"
 	// ArgNodePoolNodeIDs is a cluster's node pool nodes argument.
 	ArgNodePoolNodeIDs = "node-ids"
+	// ArgMaintenanceWindow is a cluster's maintenance window argument
+	ArgMaintenanceWindow = "maintenance-window"
 	// ArgCommandWait is a wait for a resource to be created argument.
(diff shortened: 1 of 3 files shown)


args.go
```diff
@@ -48,6 +48,8 @@ const (
 	ArgNodePoolCount = "count"
 	// ArgNodePoolNodeIDs is a cluster's node pool nodes argument.
 	ArgNodePoolNodeIDs = "node-ids"
+	// ArgMaintenanceWindow is a clsuter's maintenance window argument
```
aybabtme:
```suggestion
	// ArgMaintenanceWindow is a cluster's maintenance window argument
```

nanzhong:
Curious, is using `=` a commonly used standard? To me something like `monday@2:00` seems more natural to me, but I could be missing something.

fatih:
@nanzhong I used it because we're using the same syntax for `node-pools`. I'm ok for changing it, but I made it so it's similar anywhere.

PR #428, opened 2019-04-18
Disable terminal ECHO flag when prompting for auth token

waynr:
This makes use of the [`ReadPassword` function in `golang.org/x/crypto/ssh/terminal`](https://godoc.org/golang.org/x/crypto/ssh/terminal#ReadPassword) to avoid forcing the user to expose their auth token when entering it on the command line in response to the `doctl auth init` prompt.

This does not yet address the issue of printing the auth token when running `doctl auth init` when one is already configured. I think I will file an issue to initiate discussion as to whether or not displaying the auth token is even a concern (I am accustomed to services like Github which more tightly control visibility of auth tokens).

### DIFF
--- a/commands/auth.go
+++ b/commands/auth.go
@@ -14,7 +14,6 @@ limitations under the License.
 package commands
 
 import (
-	"bufio"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


hilary:
@waynr Please rebase onto master and resubmit so it runs green so I can approve :)

hilary:
w/re the question of displaying the auth token when one is configured: I think it is a concern, as making it visible presents an issue when recording a session, for example. Would welcome a PR addressing that issue as well!

waynr:
>Please rebase onto master and resubmit so it runs green so I can approve :)

Done!

> Would welcome a PR addressing that issue as well!

Excellent, thanks for letting me know! I guess I forgot to actually file an issue to discuss the other day but it sounds like it's unnecessary anyway.

hilary:
Looking forward to the next PR :)

PR #434, opened 2019-05-01
Display context being used instead of token

waynr:
In https://github.com/digitalocean/doctl/pull/428 we disabled the terminal ECHO flag to prevent `doctl` from echoing a user's auth token when calling `doctl auth init` for the first time (ie no auth token currently set in the `doctl` config file).

The behavior not covered in the PR was to prevent `doctl auth init` from displaying the auth token when it is already set. Instead, this PR is proposing that we display the auth context name similar to what is shown when running `doctl auth switch`. The effect is that `doctl auth init` when an auth token is already set shows:

```
Using token for context [default]

Validating token... OK
```

An alternative approach might be to instead show something like:

```
Using token [XXXXXX**********************************************************]
Validating token... OK
```

Where `XXXXXX` shown above would be the first 6 characters of the auth token. I obviously prefer using the context name but thought I would mention this possibility for maintainers to consider.

### DIFF
--- a/commands/auth.go
+++ b/commands/auth.go
@@ -78,6 +78,11 @@ func RunAuthInit(retrieveUserTokenFunc func() (string, error)) func(c *CmdConfig
 	return func(c *CmdConfig) error {
 		token := c.getContextAccessToken()
 
+		context := Context
+		if context == "" {
+			context = viper.GetString("context")
+		}
+
 		if token == "" {
 			in, err := retrieveUserTokenFunc()
 			if err != nil {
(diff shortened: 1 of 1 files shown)


kamaln7:
Thank you—this is great! I am fine with showing the context name only, although I wonder if some users rely on being able to see the token (or at least part of it). I'll defer to @hilary on that, but either way, LGTM

hilary:
I think it's best to only show the context rather than any part of the token.

In re: @kamaln7's concern about users relying on being able to see the token (or at least part of it), yeah, that's a possibility. So this change is a breaking change. It's also a security issue; folks shouldn't be depending on the token, should be using context in their scripts. As it's a security issue, I'm fine with it being a breaking change. We have one other breaking change waiting for v2.x. (https://github.com/digitalocean/doctl/pull/362)

Looks like we're coming up on a new major version...

Neikie75:
Funds

andrewsomething:
This change was Incorporated into https://github.com/digitalocean/doctl/pull/1337

PR #436, opened 2019-05-02
Create a volume from a volume snapshot

### LINKED ISSUE #420: Create a Volume from a Snapshot
I have been working on backup solution which is based on volume snapshots... So, 
is it possible to create a volume from already created volume snapshot using `doctl`?

bentranter:
Adds a `--snapshot` flag to the `doctl compute volume create` command, allowing users to create a new volume from an existing volume snapshot.

If merged, this should fix #420.

I wasn't exactly sure how to handle the `--region` flag. In `doctl` it's currently marked as required, but in the [API documentation](https://developers.digitalocean.com/documentation/v2/#create-a-new-block-storage-volume), it's stated that, 
> "The region [...] should not be specified with a snapshot_id".

My initial approach was to remove the `requiredOpt()` argument from the region flag, and update the usage string to indicate that either a region or snapshot must be included.

This is where trouble began.

I suspect there's some issue when using a non-required flag that is already registered as a required flag on a different command. Even with the `requiredOpt()` call removed from the `--region` flag, `doctl` still reported the flag as being required when I was manually testing this change. If I changed the name of the flag from `--region` to `--region-id`, `doctl` no longer reported that I was missing a required flag, presumably because that was the only usage of `--region-id` in any command.

With that said, despite the API documentation's recommendation, the command works when a region flag is set, although the API seems to ignore this flag. If I try to create a new volume from a volume snapshot and set the region to NYC1, but the volume snapshot is in TOR1, the new volume will be created in TOR1. I'm assuming that it uses the region of the snapshot, rather than the provided region value.

Anyway — I did spend some time digging, but not a ton of time, so I could be wrong about all of this. Just wanted to share my findings to hopefully save others some time if they run into this same potential issue. I figure that for expected backwards compatibility, and due to the issue I ran into, it makes sense for `doctl` to keep the `--region` flag required for now.

### DIFF
--- a/commands/volumes.go
+++ b/commands/volumes.go
@@ -44,8 +44,8 @@ func Volume() *Command {
 	AddStringFlag(cmdVolumeCreate, doctl.ArgVolumeSize, "", "4TiB", "Volume size",
 		requiredOpt())
 	AddStringFlag(cmdVolumeCreate, doctl.ArgVolumeDesc, "", "", "Volume description")
-	AddStringFlag(cmdVolumeCreate, doctl.ArgVolumeRegion, "", "", "Volume region",
-		requiredOpt())
+	AddStringFlag(cmdVolumeCreate, doctl.ArgVolumeRegion, "", "", "Volume region; should not be specified with a snapshot")
+	AddStringFlag(cmdVolumeCreate, doctl.ArgVolumeSnapshot, "", "", "Volume snapshot; should not specified with a region")
 	AddStringFlag(cmdVolumeCreate, doctl.ArgVolumeFilesystemType, "", "", "Volume filesystem type (ext4 or xfs)")
 	AddStringFlag(cmdVolumeCreate, doctl.ArgVolumeFilesystemLabel, "", "", "Volume filesystem label")
 	AddStringSliceFlag(cmdVolumeCreate, doctl.ArgTag, "", []string{}, "tags to apply to the volume; comma separate or repeat --tag to add multiple tags at once")
@@ -147,7 +147,16 @@ func RunVolumeCreate(c *CmdConfig) error {
 	region, err := c.Doit.GetString(c.NS, doctl.ArgVolumeRegion)
 	if err != nil {
 		return err
+	}
 
+	snapshotID, err := c.Doit.GetString(c.NS, doctl.ArgVolumeSnapshot)
+	if err != nil {
+		return err
+	}
+
+	if region == "" && snapshotID == "" {
+		errorMsg := fmt.Sprintf("%s.%s || %s.%s", c.NS, doctl.ArgVolumeRegion, c.NS, doctl.ArgVolumeSnapshot)
+		return doctl.NewMissingArgsErr(errorMsg)
 	}
 
 	fsType, err := c.Doit.GetString(c.NS, doctl.ArgVolumeFilesystemType)
@@ -170,6 +179,7 @@ func RunVolumeCreate(c *CmdConfig) error {
 	createVolume.SizeGigaBytes = int64(size / (1 << 30))
 	createVolume.Description = desc
 	createVolume.Region = region
+	createVolume.SnapshotID = snapshotID
 	createVolume.FilesystemType = fsType
 	createVolume.FilesystemLabel = fsLabel
 	createVolume.Tags = tags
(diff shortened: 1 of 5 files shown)


andrewsomething:
Thanks for your work on this!

> My initial approach was to remove the requiredOpt() argument from the region flag, and update the usage string to indicate that either a region or snapshot must be included.

Investigating this a bit,  it looks like it was set as required in the the doctl config file (`~/.config/doctl/config.yaml`).  Removing the following from the `required` section allows this to work as expected.

```
  volume:
    create:
      region: true
      size: true
    snapshot:
      snapshot-name: true
```

If we remove `requiredOpt()` from the region flag, we should also add some validation around its usage. If the neither the region nor snapshot are specified, we should return a `doctl.NewMissingArgsErr`

In my testing, I also noticed an additional change that will be needed. When creating from a snapshot, the API does not seem to provide the region back in the initial response. This results in a panic in the displayer `commands/displayers/volume.go`  So we'll need something along the lines of:

```
--- a/commands/displayers/volume.go
+++ b/commands/displayers/volume.go
@@ -60,11 +60,14 @@ func (a *Volume) KV() []map[string]interface{} {
                        "ID":               volume.ID,
                        "Name":             volume.Name,
                        "Size":             strconv.FormatInt(volume.SizeGigaBytes, 10) + " GiB",
-                       "Region":           volume.Region.Slug,
                        "Filesystem Type":  volume.FilesystemType,
                        "Filesystem Label": volume.FilesystemLabel,
                        "Tags":             strings.Join(volume.Tags, ","),
                }
+               m["Region"] = ""
+               if volume.Region != nil {
+                       m["Region"] = volume.Region.Slug
+               }
                m["DropletIDs"] = ""
                if len(volume.DropletIDs) != 0 {
                        m["DropletIDs"] = fmt.Sprintf("%v", volume.DropletIDs)
```

bentranter:
@andrewsomething Thanks for investigating my issue – updating my doctl config file worked! I've now,
* Removed the `requiredOpt()` from the `--region` flag, and updated the flag descriptions to mention that either a region or snapshot should be included.
* Added a check where if neither the region nor snapshot flags are specified, we return a `doctl.NewMissingArgsErr`, as suggested.
* Updated the displayer to perform a nil check on the `volume.Region` before accessing the `volume.Region.Slug`, as suggested.

If there are any other changes that should be made, feel free to let me know.

After now having implemented the change, I wonder how user-friendly the missing args check

```go
if region == "" && snapshotID == "" {
	return doctl.NewMissingArgsErr(c.NS)
}
```
is. One thing I really like about `doctl` is that if you forget a required flag that results in a `*doctl.MissingArgsErr`, it'll tell you which flag that is, as long as it was set to required. For example, before this change, if I didn't include a region flag, I'd see

```
Error: (volume.create.region) command is missing required arguments
```

In this case however, it will only show

```
Error: (volume.create) command is missing required arguments
```

which doesn't immediately help the user, unlike the previous example. With all that said, this feels like a one-off case, and hopefully shouldn't cause issues for too many users. On the other hand, I suppose adding a new type of missing argument error that can indicate that one of two possible arguments must be present could work. I'm curious as to what you think.

commands/volumes.go
```diff
@@ -147,7 +147,15 @@ func RunVolumeCreate(c *CmdConfig) error {
...
 		return err
+	}
 
+	snapshotID, err := c.Doit.GetString(c.NS, doctl.ArgVolumeSnapshot)
+	if err != nil {
+		return err
+	}
+
+	if region == "" && snapshotID == "" {
+		return doctl.NewMissingArgsErr(c.NS)
```
andrewsomething:
```suggestion
		errorMsg := fmt.Sprintf("%s.%s || %s.%s", c.NS, doctl.ArgVolumeRegion, c.NS, doctl.ArgVolumeSnapshot)
		return doctl.NewMissingArgsErr(errorMsg)
```

Good point on the error messaging. If we want to maintain a similar format, we could do something like this. Then the message would be:
> Error: (volume.create.region || volume.create.snapshot) command is missing required arguments

bentranter: Ah that's a great suggestion, thanks!

bentranter:
@hilary Thanks for the review. I've just squashed into one commit.

andrewsomething:
:shipit: Thanks for your work on this @bentranter

PR #440, opened 2019-05-08
Add flag to set current-context in local KubeConfig

### LINKED ISSUE #394: Setting Kubernetes context on first config save
When the config saved with the `doctl kubernetes cluster kubeconfig save <cluster-id|cluster-name>` command, the saved configuration has the `current-context` as empty, while the `show` command uses the cluster context in the output message. Apart from inconsistency between these two commands, I believe this is a decision taken to prevent setting cluster context automatically to the retrieved cluster context, though it may be good if it was possible to override.

In my usecase, I am using doctl on CI, in which the retrieved config file is the only one and the environment doesn't have `kubectl` available, which means I need to set `--kube-context` for every `helm` call for example, making the deploy scripts more verbose than they can be. Since the functionality is already there, I believe it would be good if `current-context` on the saved configuration is also configurable via a flag, to whether to set it to the retrieved config or not.

eddiezane:
This PR adds a new flag of `--set-current-context` when creating or updating a cluster, or saving a KubeConfig, that will set the `current-context` of your local KubeConfig to the new cluster. It defaults to true as this should be the expected behavior.

edit: Closes #394

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All changes to `doctl` will be documented in this file.
 
+## [1.18.0] - UNRELEASED
+
+ - #440 Add flag to set local KubeConfig's current-context. - @eddiezane
+
 ## [1.17.0] - UNRELEASED
 
  - #438 Remove need to opt-in to database commands. - @andrewsomething
--- a/args.go
+++ b/args.go
@@ -48,6 +48,8 @@ const (
 	ArgNodePoolNodeIDs = "node-ids"
 	// ArgCommandWait is a wait for a resource to be created argument.
 	ArgCommandWait = "wait"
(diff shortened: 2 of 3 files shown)


eddiezane:
1. There currently aren't any tests in there for any of the KubeConfig functionality and I wasn't really sure where to get started. Maybe @bouk can point me in the right direction or lend a hand.

2. Felt the same way. I think a few of these functions could be refactored into methods on something—but I'm far from a golang expert to know the right way to do that.

hilary:
Love to see what @bouk suggests!

As to 2, let's at least put our heads together and tug at it a bit, see what we come up with. 👷 👷‍♀

eddiezane:
@hilary any opposition to merging and opening a backlog issue to clean up with a shared struct as @bouk mentioned?

hilary:
@eddiezane sounds like a plan to me! :shipit:

PR #442, opened 2019-05-14
kubernetes: Support cluster upgrades

adamwg:
Add three pieces of functionality to support upgrading Kubernetes clusters:

1. Allow clusters to be created with auto-upgrade enabled, or have auto-upgrade enabled via `doctl k8s cluster update`. Show the auto-upgrade setting when getting or listing clusters.
2. Allow fetching available upgrade versions with `doctl k8s cluster get-upgrades`.
3. Allow for immediate upgrade of a cluster with `doctl k8s cluster upgrade`, which optionally takes a version. By default the latest available upgrade version will be used.

### DIFF
--- a/commands/kubernetes.go
+++ b/commands/kubernetes.go
@@ -108,6 +108,7 @@ func kubernetesCluster() *Command {
 
 	CmdBuilder(cmd, RunKubernetesClusterGet, "get <id|name>", "get a cluster", Writer, aliasOpt("g"))
 	CmdBuilder(cmd, RunKubernetesClusterList, "list", "get a list of your clusters", Writer, aliasOpt("ls"))
+	CmdBuilder(cmd, RunKubernetesClusterGetUpgrades, "get-upgrades <id|name>", "get available upgrades for a cluster", Writer, aliasOpt("gu"))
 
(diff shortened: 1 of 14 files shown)


commands/kubernetes.go
```diff
@@ -338,6 +366,67 @@ func tryUpdateKubeconfig(kube do.KubernetesService, clusterID, clusterName strin
...
+		return "", false, nil
+	case 1:
+		return versions[0].Slug, true, nil
+	}
+
+	releases, err := latestReleases(versions)
+	if err != nil {
+		return "", false, err
+	}
+	i, err := versionMaxBy(releases, func(v do.KubernetesVersion) string {
```
nanzhong: Not relevant at the moment, but in the case where both patch and minor version upgrade are available this would pick the minor upgrade rather than the patch upgrade right? This seems in line with what `latest` represents, but I feel like it'd be better/safer for the customer to default to the latest patch version.

adamwg: Good call - we should try to choose the latest available upgrade within the same minor version if possible. Updated the code to do that, and also fixed a subtle bug in the existing latest version selection code.

hilary: I would rather stick with the industry standard meaning of `latest`, which is, well, `latest`, not restricted to minor or patch (unless there's some special understanding in the k8s community). I think it would be great to also offer a `latest-minor` and a `latest-patch` as well, but let's not change the commonly understood meaning of `latest`.

Issue #446, opened 2019-05-17
Suggestion: Show the completion command in help


GiannisSialmas:
I noticed that the help text does not show completion as an available command. I had to search for it and read the read me to know the command exists. I think it would be helpful to other users to find in directly in help

eddiezane:
This is an unfortunate side effect of the dependency used for completion.

More context here https://github.com/spf13/cobra/pull/671

Maybe there's a way we can manually print out a string to mention this somehow?

eddiezane:
The original issue was `completion` was auto completing when trying to complete `compute` :facepalmpicard:

kamaln7:
Forgot to say the magic words. I merged #447 which fixes this issue.

PR #447, opened 2019-05-17
Unhide the command "completion" + prevent it from being autocompleted

kamaln7:
Currently the command is set as `Hidden` so it is not shown in the help/usage text. This PR removes the `Hidden` setting and manually deletes it from the resulting autocompletion code.


kamaln7:
Possible fix for #446

hilary:
fixes #446 
😄

Issue #451, opened 2019-05-29
Determine config file's default location at runtime


andrewsomething:
Currently `doctl`'s help output specifies that the default location of the config file is `$HOME/.config/doctl/config.yaml`

```
  -c, --config string         config file (default is $HOME/.config/doctl/config.yaml)
```

This is hardcoded at:

https://github.com/digitalocean/doctl/blob/e5406406016386e95931501fc00d082691189f65/commands/doit.go#L82

In certain environments, this is not true. For instance, on Linux the location of the config file is determined by the `XDG_CONFIG_HOME` environment variable. On Windows, `LOCALAPPDATA` is used. Additionally when using the Ubuntu snap, `XDG_CONFIG_HOME` is overridden leading to misleading help output. E.g.

```
$ locate doctl | grep '\.config/doctl/config.yaml'
/home/asb/.config/doctl/config.yaml
/home/asb/snap/doctl/19/.config/doctl/config.yaml
```

Instead of hardcoding this path in the help output, we should determine config file's default location at runtime.

Hint: A helper function already exists for this:

https://github.com/digitalocean/doctl/blob/e5406406016386e95931501fc00d082691189f65/commands/doit.go#L158-L160

hilary:
See also #457

PR #460, opened 2019-06-04
Work with kubectl

### LINKED ISSUE #457: Snap saves kubeconfig file in "/home/user/snap/doctl/39/.kube/config" when you run "doctl kubernetes cluster kubeconfig save cluster-name"

hilary:
Fixes #457 (and #451, which is closely related) using snap's `personal-files` interface and a, well, hack. (Injecting knowledge about a packager into our CLI qualifies as a hack to me, but that's me. And the benefit of having strict confinement for our devs is substantial.) 

At present, using the snap version requires a bit of setup.  After installing the snap, connect the config plugs:

```bash
sudo snap connect doctl:doctl-config
sudo snap connect doctl:kubectl-config
```

The setup only has to be done once; upgrading and/or removing and reinstalling the snap does not remove the connections.

We can submit a request for each of the `personal-files` plugs to [auto-connect](https://forum.snapcraft.io/t/process-for-aliases-auto-connections-and-tracks/455). See, for example, https://forum.snapcraft.io/t/requesting-auto-connection-of-personal-files-to-sam-cli/10641.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -3,9 +3,8 @@ version-script: git tag -l | sort --version-sort | tail -n1 | cut -c 2-
 version: git
 summary: A command line tool for DigitalOcean services
 description: doctl is a command line tool for DigitalOcean services using the API.
-base: core18
 
-grade: stable
+grade: devel
 confinement: strict
 
 build-packages:
@@ -13,7 +12,10 @@ build-packages:
   - git
 
 parts:
+  go:
+    source-tag: go1.11.4
   launcher:
+    after: [go]
     source: snap/local
     plugin: dump
     organize:
@@ -35,7 +37,7 @@ plugs:
   kube-config:
     interface: personal-files
     write:
-      - $HOME/.kube
+      -	$HOME/.kube
       - $HOME/.kube/config
   doctl-config:
     interface: personal-files
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -3,8 +3,9 @@ version-script: git tag -l | sort --version-sort | tail -n1 | cut -c 2-
 version: git
 summary: A command line tool for DigitalOcean services
 description: doctl is a command line tool for DigitalOcean services using the API.
+base: core18
 
-grade: devel
+grade: stable
 confinement: strict
 
 build-packages:
@@ -12,10 +13,7 @@ build-packages:
   - git
 
 parts:
-  go:
-    source-tag: go1.11.4
   launcher:
-    after: [go]
     source: snap/local
     plugin: dump
     organize:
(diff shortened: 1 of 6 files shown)


hilary:
It seems that a snap can only be invoked by the name of the command designated by the snap, so hard coding the name of the command as `doctl` _in the case of the snap_ won't lose our devs any functionality.

I don't like it from a coding standpoint, but I'll see what I can do to minimize the impact.

snap/snapcraft.yaml
```diff
@@ -3,28 +3,53 @@ version-script: git tag -l | sort --version-sort | tail -n1 | cut -c 2-
...
     go-importpath: github.com/digitalocean/doctl
     go-packages: [github.com/digitalocean/doctl/cmd/doctl]
+    organize:
+      bin/doctl: bin/doctl.real
+
+plugs:
+  kube-config:
+    interface: personal-files
+    write:
+      -	$HOME/.kube
```
andrewsomething: This tab breaks the build for me. Replacing it with a space resolves the issue:

```
Issues while validating snap/snapcraft.yaml: found character that cannot start any token on line 38, column 8
```

andrewsomething:
:+1:  Built and validated locally:

```
asb@asb-do:~/projects/doctl$ sudo rm /usr/local/bin/doctl
asb@asb-do:~/projects/doctl$ sudo snap install --dangerous doctl_1.20.1_amd64.snap
doctl 1.20.1 installed
asb@asb-do:~/projects/doctl$ rm ~/.kube/config 
asb@asb-do:~/projects/doctl$ which doctl
/snap/bin/doctl
asb@asb-do:~/projects/doctl$ sudo snap connect doctl:doctl-config
asb@asb-do:~/projects/doctl$ sudo snap connect doctl:kube-config
asb@asb-do:~/projects/doctl$ doctl k8s cluster config save actions-example
Notice: adding cluster credentials to kubeconfig file found in "/home/asb/.kube/config"
Notice: setting current-context to do-nyc1-actions-example
asb@asb-do:~/projects/doctl$ kubectl get nodes
NAME                     STATUS   ROLES    AGE    VERSION
actions-example-1-x1y9   Ready    <none>   7d6h   v1.13.5
actions-example-1-x1yv   Ready    <none>   7d6h   v1.13.5
```

The kubeconfig file was installed to `~/.kube/config` and the exec stanza contains `command: doctl` as expected. The correct doctl config file (`~/.config/doctl/config.yaml`) is also used.

hilary:
> However, the change it `doit.go` and the corresponding change to use `doctl.CommandName()` make sense to me assuming that `SNAP_NAME` get set when `doctl` in invoked via snap?

snap launches the snap via `exec`. It's pretty much a canonical use of `exec`, which is kinda cool. snap sets a set of environment variables that are accessible to the snap before executing the `exec`, including `SNAP_NAME`.

snap/snapcraft.yaml
hilary: Gah, thought I had gotten rid of that. Thanks for catching it. Fixed.

hilary:
I'll clean up the commits before merging!

PR #463, opened 2019-06-05
Add kubernetes delete-node and replace-node command

bouk:
* Upgrade godo
* Deprecate `recycle` command.


commands/kubernetes.go
```diff
@@ -238,17 +238,17 @@ func kubernetesNodePools() *Command {
...
 	cmdKubeNodePoolDelete := CmdBuilder(cmd, RunKubernetesNodePoolDelete,
 		"delete <cluster-id|cluster-name> <pool-id|pool-name>",
 		"delete node pool from a cluster", Writer, aliasOpt("d", "rm"))
 	AddBoolFlag(cmdKubeNodePoolDelete, doctl.ArgForce, doctl.ArgShortForce,
 		false, "force node pool delete")
 
+	cmdKubeNodeDelete := CmdBuilder(cmd, RunKubernetesNodeDelete, "delete-node <cluster-id|cluster-name> <pool-id|pool-name> <node-id>", "delete node in a node pool", Writer)
+	AddBoolFlag(cmdKubeNodeDelete, doctl.ArgForce, doctl.ArgShortForce, false, "force node delete")
+	AddBoolFlag(cmdKubeNodeDelete, "skip-drain", "", false, "skip draining the node before deletion")
+	AddBoolFlag(cmdKubeNodeDelete, "replace", "", false, "replace the node with a new one")
```
adamwg: I'm on the fence about whether this should be a flag or a separate command. If a user wants to replace a node, it doesn't feel entirely intuitive to look under the `delete` command, especially since in the control panel we offer both `Delete` and `Recycle` options.

I'm all for getting rid of the `recycle` terminology since we know it's confusing, but I wonder whether we should have both a `delete-node` command and a `replace-node` command.

Not a blocker for me but interested in other opinions. @hilary @andrewsomething any thoughts?

commands/kubernetes.go
```diff
@@ -912,20 +918,32 @@ func RunKubernetesNodePoolDelete(c *CmdConfig) error {
...
 		return err
 	}
-	if force || AskForConfirm("delete this Kubernetes node pool") == nil {
-		kube := c.Kubernetes()
-		if err := kube.DeleteNodePool(clusterID, poolID); err != nil {
-			return err
-		}
-	} else {
+
+	if !(force || AskForConfirm("delete this Kubernetes node") == nil) {
```
adamwg: Wonder if we should call out here (when `replace` is false) or in the help that this operation will permanently resize the pool. This might be a bit confusing since we can also resize pools via the `node-pool update` command.

commands/kubernetes.go
hilary: Removing the `recycle` command is a breaking change. You can deprecate it, if you wish, and we could remove it in v2.0, but removing it in a single release without messaging and semantic versioning, etc., isn't conducive to a great developer experience for our community :)

hilary: As for the design question, when a semantic question like this feels ambiguous, I'm a fan of expressiveness. I would offer both. Some folks will look for a flag, others will look for a separate command. So alias the separate command to the flag.

commands/kubernetes.go
hilary: More info is generally better with these sorts of confirmations. Especially in a case like this, wherein the number of nodes remaining in the pool could affect the response.

commands/kubernetes.go
hilary: Ok, checked in and the recycle command is seeing usage so we do need to go through the deprecation process. Not a lot of use, but some.

bouk:
I like the suggestion about mirroring the experience of the UI, so I've split it up in `delete-node` and `replace-node`.

Instead of removing the recycle command I've added a deprecation notice to the description, and I've hidden it from the help, so if people using it in scripts or whatever it'll keep working.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All changes to `doctl` will be documented in this file.
 
+## [1.20.0] -
```
hilary: It's traditional to put `UNRELEASED` instead of the date. Not going to hold the PR for it, though :)

bouk:
Done

PR #467, opened 2019-06-12
set snap version using tag

hilary:
Streamlining the release process one annoyance at a time :)

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,7 +1,8 @@
 name: doctl
-version: "1.20.0"
+version-script: git tag -l | sort --version-sort | tail -n1 | cut -c 2-
+version: git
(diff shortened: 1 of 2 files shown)


snap/snapcraft.yaml
```diff
@@ -1,7 +1,8 @@
 name: doctl
-version: "1.20.0"
+version-script: git tag -l | sort --version-sort | tail -n1
```
andrewsomething: This output includes the preceding `v`, eg: `v1.20.0` Not sure if that matters, but since we are switching from not having the `v` it could potentially cause issues with version sorting.

andrewsomething:
One comment, but besides that: LGTM :+1:

snap/snapcraft.yaml
hilary: Nice catch!

PR #469, opened 2019-06-13
set default location for config file to configPath

### LINKED ISSUE #451: Determine config file's default location at runtime

hilary:
fixes #451

### DIFF
--- a/commands/doit.go
+++ b/commands/doit.go
@@ -79,7 +79,8 @@ var ErrNoAccessToken = errors.New("no access token has been configured")
 func init() {
 	cobra.OnInitialize(initConfig)
 
-	DoitCmd.PersistentFlags().StringVarP(&cfgFile, "config", "c", "", "config file (default is $HOME/.config/doctl/config.yaml)")
+	DoitCmd.PersistentFlags().StringVarP(&cfgFile, "config", "c", "",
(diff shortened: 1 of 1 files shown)


andrewsomething:
:shipit:

PR #472, opened 2019-06-19
migrate from go dep to go modules

hilary:
- run `go mod init`
- remove go versions not supporting go modules from Travis
- use go 1.12 for building
- update developer docs
- remove go dep files (Gopkg.*)
- update vendor folder


go.mod
```diff
@@ -0,0 +1,51 @@
...
+	golang.org/x/crypto v0.0.0-20181203042331-505ab145d0a9
+	golang.org/x/net v0.0.0-20190206173232-65e2d4e15006 // indirect
+	golang.org/x/oauth2 v0.0.0-20190402181905-9f3314589c9a
+	golang.org/x/sys v0.0.0-20190312061237-fead79001313 // indirect
+	golang.org/x/text v0.3.1-0.20181227161524-e6919f6577db // indirect
+	golang.org/x/time v0.0.0-20181108054448-85acf8d2951c // indirect
+	gopkg.in/inf.v0 v0.9.1 // indirect
+	gopkg.in/yaml.v2 v2.2.2
+	k8s.io/api v0.0.0-20190606204050-af9c91bd2759 // indirect
+	k8s.io/apimachinery v0.0.0-20190404173353-6a84e37a896d
```
bhcleek: Is this the right `k8s.io/apimachinery` as of `k8s.io/api@af9c91bd2759`? If I had to guess, I'd say that you upgraded `k8s.io/client-go` last, and it upgraded `k8s.io/api` on you, too. Kubernetes packages should generally be kept inline with each other.

hilary: Ah, nice catch.

PR #474, opened 2019-06-20
Support "Are you sure?" prompting

DazWilkin:
Running the `doctl` container with `--interactive --tty` will work if `doctl` commands issue `Are you sure?` prompts.

Consistent use of --long-flag-names to help users understand the intent of the Docker commands.


hilary:
nice!

Issue #445, opened 2019-05-15
Cleanup Kubernetes code with a shared struct


eddiezane:
This is in reference to the discussion in #440.

We need to clean up the code base by creating a shared struct to define private methods with scope of passed in arguments.

andrewsomething:
@bentranter can this be closed with the introduction of the `KubernetesCommandService`?

PR #478, opened 2019-06-27
Ensure kubeconfig exec is context-aware

### LINKED ISSUE #459: kubeconfig save, exec-credential not aware of context
**Actual behaviour / how to reproduce**

When authenticated for the first time with `doctl auth init` it uses the `default` context, which is expected. You can then call `doctl kubernetes cluster kubeconfig save CLUSTER_NAME` and it will fetch the users/contexts and save it to the kubeconfig.

The issue occurrs once you need to add a second context to doctl, using a command like `doctl auth init --context=everyday` and then wanting to add a cluster from that context to your kubeconfig using `doctl kubernetes cluster kubeconfig save CLUSTER_NAME2 --context=everyday` 

You would now expect `kubectl cluster-info` to display information about your cluster, but instead it shows an error similar to the following:
`Error: failed to fetch credentials for cluster "37ea18d5-xxxx-xxxx-xxxx-xxxxxxxxxx": cluster not found`

This is because it still assumes it should use the `default` context, when in-fact the `default` context does not have access to this cluster.

**Expected behaviour / fix**

Expected behaviour would be for the  `doctl kubernetes cluster kubeconfig save CLUSTER_NAME2 --context=everyday` to pass the context to `exec-credential` in the kubeconfig as shown in the image below:

image

After adding this line everything works fine, even after removing it again for some reason, which I can only assume is because the context the cluster is supposed to have is cached at this point.

This should be a pretty simple fix for anyone familiar with the doctl codebase or Go in particular

bentranter:
Fixes #459

Background: When an operation via doctl is invoked that writes to the user's kubectl config (for example, `doctl kubernetes cluster kubeconfig`), the `--context` flag is used by doctl, but **not** kubectl.

This problematic, because we use a command called `exec-credential` that grabs certificates on the fly from DO Kubernetes. This `exec-credential` call is invoked by kubectl, so if kubectl is not aware of the value of the `--context` flag passed by the user during the kubeconfig command, then `exec-credential` will not use that context either, causing the kubectl command to fail unexpectedly.

This change updates the `mergeKubeconfig` function responsible for updates to the kubectl config to always pass a `--context` flag. If there is no context set, it uses the default context, otherwise it uses the one set by the command.

### DIFF
--- a/commands/kubernetes.go
+++ b/commands/kubernetes.go
@@ -91,6 +91,9 @@ func Kubernetes() *Command {
 	return cmd
 }
 
+// KubernetesCommandService is used to execute Kubernetes commands.
+type KubernetesCommandService struct{}
+
 func kubernetesCluster() *Command {
 	cmd := &Command{
 		Command: &cobra.Command{
@@ -101,18 +104,20 @@ func kubernetesCluster() *Command {
 		},
 	}
 
+	cmdService := &KubernetesCommandService{}
+
 	cmd.AddCommand(kubernetesKubeconfig())
 
 	cmd.AddCommand(kubernetesNodePools())
 
-	CmdBuilder(cmd, RunKubernetesClusterGet, "get <id|name>", "get a cluster",
+	CmdBuilder(cmd, cmdService.RunKubernetesClusterGet, "get <id|name>", "get a cluster",
 		Writer, aliasOpt("g"), displayerType(&displayers.KubernetesClusters{}))
-	CmdBuilder(cmd, RunKubernetesClusterList, "list", "get a list of your clusters",
+	CmdBuilder(cmd, cmdService.RunKubernetesClusterList, "list", "get a list of your clusters",
 		Writer, aliasOpt("ls"), displayerType(&displayers.KubernetesClusters{}))
-	CmdBuilder(cmd, RunKubernetesClusterGetUpgrades, "get-upgrades <id|name>",
+	CmdBuilder(cmd, cmdService.RunKubernetesClusterGetUpgrades, "get-upgrades <id|name>",
 		"get available upgrades for a cluster", Writer, aliasOpt("gu"))
 
-	cmdKubeClusterCreate := CmdBuilder(cmd, RunKubernetesClusterCreate(defaultKubernetesNodeSize,
+	cmdKubeClusterCreate := CmdBuilder(cmd, cmdService.RunKubernetesClusterCreate(defaultKubernetesNodeSize,
 		defaultKubernetesNodeCount), "create <name>", "create a cluster",
 		Writer, aliasOpt("c"))
 	AddStringFlag(cmdKubeClusterCreate, doctl.ArgRegionSlug, "", defaultKubernetesRegion,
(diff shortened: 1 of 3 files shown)


commands/kubernetes.go
```diff
@@ -1307,6 +1307,11 @@ func mergeKubeconfig(clusterID string, remote, local *clientcmdapi.Config, setCu
 		local.CurrentContext = remote.CurrentContext
 	}
 
+	context := Context
```
andrewsomething: I wonder if we should use something a but more descriptive that `context`? I think there are at least three distinct "contexts" in this file.

bentranter:
@andrewsomething I've updated the approach taken to determine the current context, although I'm not sure this change is the best user experience.

Setting the `--context` flag on the `exec-credential` command in the user's kubeconfig every time they run `doctl kubernetes cluster kubeconfig save` technically works, but I worry that users will run into unexpected issues when they switch between doctl contexts and find that their `kubectl` operations are suddenly failing due to an authentication error, because they haven't re-run `kubeconfig save`.

I wonder if it would be better to set the `--context` in the user's kubeconfig file only when they explicitly run `doctl kubernetes cluster kubeconfig save <name> --context <context>`, and then not include the flag otherwise so that it can be inferred from the doctl config when the user uses `kubectl` and `exec-credential` is called. What do you think?

bentranter:
@hilary Totally agreed. That said, I haven't quite figured out how to test this given our current approach.

What I'd like to test is that, in the user's kubeconfig, every cluster maps one-to-one with the correct auth context, and that the correct `--context` flag is written to the kubeconfig. Ideally we'd test three scenarios:
1. `doctl auth init` followed by `doctl kubernetes cluster kubeconfig save` writes `--context=default` to the kubeconfig.
2. `doctl auth init switch --context=next` followed by `doctl kubernetes cluster kubeconfig save` writes `--context=next` to the kubeconfig.
3. `doctl kubernetes cluster kubeconfig save --context=<some-valid-context>` writes `--context=<some-valid-context>` to the kubeconfig.

However, testing all of that via the `RunKubernetesKubeconfigSave` has been challenging so far, as our saving logic depends on reading from the user's doctl config, their kubeconfig, and a remote DOKS config. In the other tests, we'd mock this behaviour, but for this particular function, there are some calls to third party libs to do things like retrieving the remote kubeconfig that we'd need to mock before this is (more easily) testable.

It looks like there's been some discussion about this before in #440 and #445 – maybe it's time to start on some of the changes suggested there? Refactoring to a shared struct would make it possible to mock some of the third party functionality we currently depend on (I think at least, I'm still getting a feel for the approach we take to mocking).

hilary:
@bentranter Agreed about the shared struct. 

Re: approach. It's the classic case of code that doesn't have tests not being written to be testable. For lots of reasons, but one big one being a lack of separation of concerns. We're just going to have to build in this work as we go.

Otoh, the code is solid, readable code. I've found I can untangle the pieces and refactor fairly fluently.

bentranter:
@hilary Ok, I've refactored to make this more testable. I'll summarize the changes and why they were needed.

#### `RunKubernetes*` commands use a shared struct

I've updated all of the `RunKubernetes*` commands so that they're methods on a shared struct. This was the only way I could come up with the allows me to mock the kubeconfig access methods in a way that allows me to access the written kubeconfig in the test suite.

#### Reading and writing from a local and remote kubeconfig is done via an interface

In the actual code, I've refactored a handful of the helper methods to use this new interface. This allows me to mock the interface in the test suite, allowing me to set my own remote and local kubeconfig values, and see what ends up being written to the user's local kubeconfig, all without writing to any actual files.

#### Getting the current auth context is done via a global `getCurrentAuthContextFn()` call 😞

Unfortunately, any of the persistent flags seem to use a global Viper instance, rather than the one set on the config. This prevented me from passing a `--context` flag in the test suite, because that call wrote the the test suite's config rather than the global one (I think at least, I didn't spend a _ton_ of time on this).

The solution to this is somewhat of a hack unfortunately, in that in any test where I need to pass a `--context` flag, I mock the `getCurrentAuthContextFn()` to use the test suite's viper instance. I then reset it to the original after that test case invocation.

I'd like to write a few more test cases, but wanted to get your feedback early on this approach. I'm not super happy with the global `getCurrentAuthContextFn` situation, but I figured it could be better than starting a much bigger refactor to deal with the global Viper stuff.

hilary:
I really like your approach.

Let me tell you why 😄 

1. Getting rid of the global Viper instance ("code written without tests ~= untestable code") is a separate refactor
1. `KubeconfigProvider` and `KubernetesCommandService` feel like a big step in the direction of the pluggable version of doctl. I particularly like that you aren't prematurely generalizing, but are designing in the direction we want to go.
1. While `getCurrentAuthContextFn` is a bit hacky, it's small and contained. Sometimes you need to introduce small hacky bits as bridges during refactors.

PR #482, opened 2019-07-03
Allow completion of aliases in zsh

eddiezane:
This PR allows the completion of aliases of commands in zsh.

For example:

`doctl k8s <tab>` will now properly complete as if the command was typed as `doctl kubernetes <tab>`

Ref:
https://github.com/helm/helm/pull/5680
https://github.com/helm/helm/commit/48f0e3101d49a3de1466ff75fc786ce012f369cc


kamaln7:
Confirmed working, these autocomplete as expected:

```
doctl kubernetes <tab>
doctl k8s <tab>
```

But `doctl k8<tab>` doesn't autocomplete to `k8s`. I guess that's because `k8s` isn't exported as a command, only as an alias?

PR #483, opened 2019-07-11
1.21.0

hilary:
The change in format to the changelog is a result of using [github-release-notes](https://github.com/buchanae/github-release-notes), which will be baked into the scripts real soon now(tm).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Change Log
 
+## [1.21.0] - 2019-07-11
+
+- PR #482 - @eddiezane - Allow completion of aliases in zsh
+- PR #477 - @hilary - describe current status of snap in README
+- PR #474 - @DazWilkin - Support "Are you sure?" prompting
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Change Log
 
+## [1.21.0] - 2019-07-11
+
+- PR #482 - @eddiezane - Allow completion of aliases in zsh
```
bentranter: For the last year or so, it looks like every changelog entry follows the format,

```
- #<PR number> - <description> - @<github username>
```

but this entry deviates from that. How do you feel about changing it to be consistent?

hilary: The new format is the format we get from `github-release-notes` (see PR comment). While we could wrestle with it to get it to conform with the old format, I don't think it's worth it, especially as we would then have to maintain that code... the changelog is consumed by humans, not code. Humans can handle the switch :)

bentranter: Cool, that was my assumption, but wanted to check in case there was any consumption by machines. Sounds good to me!

PR #492, opened 2019-07-15
track tool dependencies

hilary:
https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module


hilary:
will re-open later when ready

Issue #496, opened 2019-07-18
doctl incorrectly reports that a newer version is available


andrewsomething:
As pointed out in #495, doctl incorrectly points out that a new version is available despite being on the latest release. This is because the current build was built with the `dev` label rather than the `release` one. This looks to be caused by changes introduced in https://github.com/digitalocean/doctl/commit/6e988dfadee614295cd2ec979d658794bb32d847 where the all the logic around setting the `LDFLAGS`  in `scripts/stage.sh` was removed.

```
$ curl -sL https://github.com/digitalocean/doctl/releases/download/v1.21.1/doctl-1.21.1-linux-amd64.tar.gz | tar -xzv
doctl
$ ./doctl version 
doctl version 1.21.1-dev
release 1.21.1 is available, check it out! 
```

PR #497, opened 2019-07-18
Re-add the logic setting LDFLAGS in scripts/stage.sh (Fixes: #496).

andrewsomething:
This fixes the regression introduced in: https://github.com/digitalocean/doctl/commit/6e988dfadee614295cd2ec979d658794bb32d847

The LDFLAGS are necessary for setting the correct version and labels at build time when cutting a new release.

### DIFF
--- a/scripts/stage.sh
+++ b/scripts/stage.sh
@@ -32,14 +32,21 @@ mkdir -p $STAGE_DIR $RELEASE_DIR
 if [[ -z $SKIPBUILD ]]; then
   echo "building doctl"
 
+  baseFlag="-X github.com/digitalocean/doctl"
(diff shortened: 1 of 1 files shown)


scripts/stage.sh
```diff
@@ -32,14 +32,21 @@ mkdir -p $STAGE_DIR $RELEASE_DIR
 if [[ -z $SKIPBUILD ]]; then
   echo "building doctl"
 
+  baseFlag="-X github.com/digitalocean/doctl"
+  ldflags="${baseFlag}.Build=$(git rev-parse --short HEAD)"    cd $DIR/../cmd/doctl
```
hilary: A merge issue?

andrewsomething: Opps. Fixed up.

PR #505, opened 2019-07-22
dbaas: add private-network-uuid to create, migrate, and replica create

sunny-b:
This adds the `private-network-uuid` field to Database creates, migrates and replica creates so that users can use `doctl` to move/create their Databases inside of their private networks/VPCs.

cc: @hilary @mikejholly


mikejholly:
LGTM

PR #506, opened 2019-07-22
1.23.0

bentranter:
Prepare for release 1.23.0.


bentranter:
@hilary #502 and #503 are in now 🙂

PR #508, opened 2019-07-22
Fix distfile basename in staging script

bentranter:
Before this change, the basename would always be generated to contain the string `.exe`. This fixes the script in a shellcheck-compliant way so that we're back to the old behaviour.

It also fixes the error introduced by the `u` flag in `set -euo pipefail` when the script is invoked without a `label` or set `SKIPBUILD` so that those fall back to an empty default value.

### DIFF
--- a/scripts/stage.sh
+++ b/scripts/stage.sh
@@ -5,7 +5,7 @@ set -euo pipefail
 major=$1
 minor=$2
 patch=$3
-label=$4
+label=${4-}
 
 if [[ -z "$major" || -z "$minor" || -z "$patch" ]]; then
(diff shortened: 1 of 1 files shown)


scripts/stage.sh
```diff
@@ -52,7 +52,7 @@ fi
 cd "$RELEASE_DIR"
 
 for f in "$STAGE_DIR"/*; do
-  distfile_basename=$(basename "${f}%.exe")
+  distfile_basename=$(basename "${f%".exe"}")
```
hilary: Doh! That's what I get for continuing to code when frustrated...

PR #512, opened 2019-07-23
Die hardcoded version, die

hilary:
The `VERSION` file serves as a bridge between when we bump the version and when we tag it... I keep oscillating as to whether to run `bump-version` before or after submitting the PR for review.

Of course, once we're at CD, that will go away (yay).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,9 @@ _vendor-*
 
 /out
 
+# Build artifacts
+Dockerfile-e
+
 # Snapcraft
 parts/
 prime/
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -113,12 +113,9 @@ You will also need a valid `GITHUB_TOKEN` environment variable with access to th
 1. Run `make changelog` and add the results to the [CHANGELOG](https://github.com/digitalocean/doctl/blob/master/CHANGELOG.md)
    under the version you're going to release if they aren't already there.
 
-   Update the version in:
-
-   * `doit.go`
-   * `Dockerfile`
-
 1. Generate a PR, get it reviewed, and merge
+
+1. Run `make bump-version`.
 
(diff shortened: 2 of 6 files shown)


bentranter:
Two things, but looks great overall – I'm definitely looking forward to the day when this is all done automatically 😄

1. If I run this without already having `sembump` installed locally, I get the following output:
    ```
    make: sembump: Command not found
    BUMP=<patch|feature|breaking> bump version
    Bumping VERSION from v1.23.1 to v
    ```
    Now my Makefile-foo isn't the best, but I think the issue is that the variable expansion is evaluated before the commands are run, so `VERSION` and `NEW_VERSION` are set before `sembump` is downloaded.

    I know this can be fixed by making the sembump installation step it's own target that is a prerequisite of the `bump-version` target, but maybe there's a better way that I'm not aware of. I'm sure you get what I mean, but below is the code I used that worked for me.

    ```Makefile
    .PHONY: install-sembump
    install-sembump:
        @GO111MODULE=off go get -u github.com/jessfraz/junk/sembump # update sembump tool

    .PHONY: bump-version
    bump-version: install-sembump
        @echo "BUMP=<patch|feature|breaking> bump version"
        $(eval VERSION = $(shell ./scripts/version.sh -s))
        $(eval NEW_VERSION = $(shell sembump --kind $(BUMP) $(VERSION)))
        @echo "Bumping VERSION from v$(VERSION) to v$(NEW_VERSION)"
        @echo "v$(NEW_VERSION)" > VERSION
        @sed -i'' -e 's/${VERSION}/${NEW_VERSION}/g' Dockerfile
    ```
1. Where are we planning on setting the values of Build, Major, Minor, Patch, and Label? I don't see any changes to the way we're building the binaries to populate those values via the `-ldflags "-X <importpath>.<name>=<value>`.

hilary:
Ok, put our heads together and split in two. First of which is #514

PR #516, opened 2019-07-24
WIP: release using goreleaser

hilary:
TODO:

- test the artifacts (there's a set in my github account)
- revert commit with test settings

PR #521, opened 2019-07-26
release in one step leveraging goreleaser

hilary:
see also #516 

Removed dockerhub integration, will submit that separately

CODE NOW RELEASES TO PROD TARGETS!


Makefile
```diff
@@ -116,4 +73,91 @@ docker_build: _base_docker_cntr
...
+# ORIGIN ?= origin
+ORIGIN ?= hilary
+
+BUMP ?= patch
+
+.PHONY: version
+version:
+	@echo "==> determining doctl version"
+	@echo ""
+	@ORIGIN=$$ORIGIN scripts/version.sh
```
bentranter: This syntax didn't work for me during my local testing, I had to use,

```Makefile
        @ORIGIN=${ORIGIN}
```

And then it all worked great.

Makefile
```diff
@@ -116,4 +73,91 @@ docker_build: _base_docker_cntr
...
+_install_sembump:
+	@echo "==> installing/updating sembump tool"
+	@echo ""
+	@GO111MODULE=off go get -u github.com/jessfraz/junk/sembump
+
+.PHONY: _bump_and_tag
+_bump_and_tag: _install_sembump
+	@echo "==> BUMP=${BUMP} bumping and tagging version"
+	@echo ""
+	@ORIGIN=$$ORIGIN scripts/bumpversion.sh
```
bentranter: Same here, I had to use,

```Makefile
        @ORIGIN=${ORIGIN}
```

Dockerfile.goreleaser
```diff
@@ -0,0 +1,9 @@
+FROM alpine:3.8
+
```
bentranter: Adding,

```Dockerfile
RUN apk add --no-cache curl
```

here fixes the issue with missing certs, from my testing at least.

hilary: really? That's great!

PR #528, opened 2019-07-31
add `make snap`

hilary:
The automated build/release pipeline provided by snapcraft is both fragile and opaque. Rather than continue to wrestle with it given that our snap is rather advanced, I dockerized the build.


hilary:
We could probably move building the snap to travis-ci as part of CD, given that we're using the snapcraft docker image.

PR #531, opened 2019-08-01
clean out dist as part of make clean

hilary:
Quick improvement I noticed in passing


hilary:
git cherry-pick is your friend :)

PR #533, opened 2019-08-01
remove obsolete recorder.go

hilary:


hilary:
Ahah! That's where it is used :)

PR #539, opened 2019-08-05
remove obselete implementation of Command

hilary:
Please review carefully...

doctl heavily overloads `Command` :(

I could not find any uses of the `Command interface`, `LiveCommand`, or `MockCommand` types defined in this top level `command.go` other than in `command_test.go`

Note that all uses of `Command` in `commands/*` reference a definition of `Command` `commands/command.go`.


zachgersh:
This looks good to me. I also took a search and couldn't find the code referenced in the deletion.

PR #540, opened 2019-08-05
WIP: remove ancient (circa 2016) plugin system

hilary:
Ok, I'll need to instrument this code so we can figure out if it is in use.

Check out the [history for `plugin.go`](https://github.com/digitalocean/doctl/commits/master/commands/plugin.go) and all will be revealed...


commands/plugin.go
```diff
@@ -1,139 +0,0 @@
...
-}
-
-func searchPlugins() ([]displayers.PlugDesc, error) {
-	envPath := os.Getenv("PATH")
-	paths := strings.Split(envPath, string(os.PathListSeparator))
-
-	var plugs []displayers.PlugDesc
-
-	for _, p := range paths {
-		matches, err := filepath.Glob(filepath.Join(p, "doit-provider-*"))
```
zachgersh: We were worried that this functionality could still be in use based on the fact that the code above just looks for any provider in your `PATH`

hilary:
Closing this until I instrument...

PR #559, opened 2019-08-22
Switch to GitHub changelog generator

bentranter:
Switches from github-release-notes to our newly open-sourced github-changelog-generator. Follows the pattern I used with mockgen, where we build the binary from the vendored library, and then delete it after the command has been run so we don't pollute the working directory or overwrite anything in the user's `$GOBIN` directory.


go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/BurntSushi/toml v0.3.1 // indirect
 	github.com/blang/semver v3.5.1+incompatible
 	github.com/creack/pty v1.1.7
+	github.com/digitalocean/github-changelog-generator v0.0.0-20190822155117-98248440dda1
```
hilary: I don't think you want ghcg in `go.mod` :)

scripts/changelog.sh
```diff
@@ -3,8 +3,6 @@
 set -o pipefail
 
 tfile=$(mktemp /tmp/doctl-CHANGELOG-XXXXXX)
-github-release-notes -org digitalocean -repo doctl -since-latest-release -include-author >"$tfile"
-
-GO111MODULE=on go mod tidy
```
hilary: You'll need the `go mod tidy` to prevent ghcg from ending up in `go.mod`

vendor/modules.txt
```diff
@@ -4,6 +4,9 @@ github.com/blang/semver
 github.com/creack/pty
 # github.com/davecgh/go-spew v1.1.1
 github.com/davecgh/go-spew/spew
+# github.com/digitalocean/github-changelog-generator v0.0.0-20190822155117-98248440dda1
+github.com/digitalocean/github-changelog-generator
```
hilary: These are here because they are in `go.mod`

scripts/changelog.sh
bentranter: This is done, not sure why GitHub isn't picking up the change

hilary: I see it :)

PR #560, opened 2019-08-23
DBaaS: Remove hardcoded default version for creates.

andrewsomething:
The API itself will default to a sane version for the specified database engine. By using a hard-coded default of 11, attempts to create MySQL or Redis clusters without explicitly specifying the version will fail. Eg:

```
$ doctl databases create --engine mysql doctl-mysql
Error: POST https://api.digitalocean.com/v2/databases: 422 (request "a11fa359-5778-42a5-8474-d776c5c097fd") invalid cluster engine version
```

### DIFF
--- a/commands/databases.go
+++ b/commands/databases.go
@@ -25,11 +25,10 @@ import (
 )
 
 const (
-	defaultDatabaseNodeSize      = "db-s-1vcpu-1gb"
-	defaultDatabaseNodeCount     = 1
-	defaultDatabaseRegion        = "nyc1"
-	defaultDatabaseEngine        = "pg"
-	defaultDatabaseEngineVersion = "11"
+	defaultDatabaseNodeSize  = "db-s-1vcpu-1gb"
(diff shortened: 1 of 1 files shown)


senechko:
this makes the version required, but does not give any hints as to the possible values. what if you updated the `buildDatabaseRequestFromArgs`  method to first look at the version argument (if any) and if not present you look up the default version based on the db engine from a map?
```
var defaultDatabaseVersions = map[string]string{
	"pg":    "11",
	"mysql": "8",
	"redis": "5",
}

...

if version == nil {
  version = defaultDatabaseVersions[engine]
}

```

andrewsomething:
> this makes the version required

It does not; it just removes the hardcoded default. Specifying the version is not required by the API.

> look up the default version based on the db engine from a map

Personally, I'd rather allow the API itself to handle this. It already provides sane defaults for each engine if you exclude the version from your request. This allows the DBaaS team to update defaults as they see fit without requiring users to upgrade doctl.

PR #575, opened 2019-09-11
Update to godo v1.20.0

snormore:
https://github.com/digitalocean/godo/releases/tag/v1.20.0


snormore:
Not ready for review yet, still need to get the optional flags on update working properly.

doit.go
```diff
@@ -286,6 +288,23 @@ func (c *LiveConfig) GetBool(ns, key string) (bool, error) {
 	return viper.GetBool(nskey), nil
 }
 
+// GetBoolPtr returns a config value as a bool.
```
adamwg: Nit: Should read `returns a config value as a *bool` or `as a bool pointer`?

doit.go
```diff
@@ -304,6 +323,31 @@ func (c *LiveConfig) GetInt(ns, key string) (int, error) {
 	return val, nil
 }
 
+// GetIntPtr returns a config value as an int.
```
adamwg: Same nit as above.

PR #584, opened 2019-09-17
removes obselete do categories

zachgersh:
- we can't seem to find a place where these are actually used


andrewsomething:
AFAIU, this was used in an old doc generation script to put the output into separate directories based on the category. This was removed in https://github.com/digitalocean/doctl/pull/527

See: https://github.com/digitalocean/doctl/blob/e989a7ec8afa50f9fd5615a295158dcf25b08bc0/cmd/doctl-gen-doc/main.go#L102

PR #588, opened 2019-09-19
build a windows exe if we are on windows

zachgersh:
I left this out when I originally wrote the integration tests.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -1,5 +1,0 @@
-#!/usr/bin/env sh
-
-GO111MODULE="on" go test -race -mod=vendor ./commands/... ./do/... ./pkg/... .
-# disable until integration tests green on windows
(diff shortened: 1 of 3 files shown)


scripts/test.sh
```diff
@@ -2,4 +2,4 @@
 
 GO111MODULE="on" go test -race -mod=vendor ./commands/... ./do/... ./pkg/... .
 # disable until integration tests green on windows
-#GO111MODULE="on" go test -race -mod=vendor ./integration
```
hilary: Rather than uncomment line, just delete this file and use `make test` 😄 

Update `.travis.yml` accordingly

zachgersh: 👯‍♂

zachgersh:
@hilary make seems to not be installed as a package :/ https://docs.travis-ci.com/user/reference/windows/#pre-installed-chocolatey-packages

PR #593, opened 2019-09-24
displayer no longer relies on a config

zachgersh:
- we want to break the dependency that many objects
in doctl have on some form of config. This PR does
that by taking ONLY the values off of the config we
need to dictate display.
- DisplayText has also become a public method as we
would eventually like to remove the need for a displayer
all together BUT we want to attempt to do that incrementally.
A follow on PR will hopefully now use this public method from
within a command itself. Commands have a 3 jobs, parse their
specific config, run their action, and render their output
error.


commands/displayers/output.go
```diff
@@ -71,97 +31,51 @@ type Displayable interface {
...
-		return false
-	}
-
-	return true
-}
-
-func displayText(item Displayable, out io.Writer, includeCols []string) error {
-	w := newTabWriter(out)
+// DisplayText ends up writing tabbed content to the passed in io.Writer
+// while potentially adding our removing headers.
```
bentranter: `our` should be `or`

bentranter: I guess if I want to be ultimate strict about this, the comment should read in the form,

```go
// DisplayText writes tabbed content...
```

Issue #595, opened 2019-09-25
kubectl cannot load credentials from doctl


webartistse:
To use DO access token in CI you need the DIGITALOCEAN_ACCESS_TOKEN env. Kubectl cannot get the token I guess.

Setup:
- Bitbucket pipelines
- DO Kuberenetes

**Reproduce with image atlassian/default-image:2:**
`curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl`

`chmod +x ./kubectl`

`mv ./kubectl /usr/local/bin/kubectl`

`curl -LO https://github.com/digitalocean/doctl/releases/download/v1.32.0/doctl-1.32.0-linux-amd64.tar.gz`

`tar xf ./doctl-1.32.0-linux-amd64.tar.gz`

`chmod +x ./doctl`

`mv ./doctl /usr/local/bin/doctl`

`export DIGITALOCEAN_ACCESS_TOKEN=$DIGITALOCEAN_APIKEY`

`doctl kubernetes cluster kubeconfig save cluster-test`

`kubectl config get-contexts`

`kubectl get pods`

**Get pods will get:**
root@2c3df388eda2:/opt/atlassian/bitbucketci/agent/build# kubectl get pods
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xadb9fc]

goroutine 1 [running]:
github.com/digitalocean/doctl/commands.execCredentialFromConfig(0xc0000bdda0, 0x7f0, 0xe00, 0xc0000bdda0)
	/home/travis/gopath/src/github.com/digitalocean/doctl/commands/kubernetes.go:803 +0xfc
github.com/digitalocean/doctl/commands.(*KubernetesCommandService).RunKubernetesKubeconfigExecCredential(0xc000254ce0, 0xc00026f680, 0x0, 0xc00026f680)
	/home/travis/gopath/src/github.com/digitalocean/doctl/commands/kubernetes.go:775 +0x472
github.com/digitalocean/doctl/commands.cmdBuilderWithInit.func1(0xc00025d180, 0xc00026dc80, 0x1, 0x3)
	/home/travis/gopath/src/github.com/digitalocean/doctl/commands/doit.go:432 +0x118
[...]

PR #596, opened 2019-09-25
Restore old config file location

bentranter:
Relates to #595.

In our most recent release, we created two issues,

• We stopped creating the directory where the config lives for users who had just installed doctl, causing file creates to fail,
• The default Go 1.13 behaviour changed the expected location of the config file, so all users who updated ended up needing to create a new config file, which breaks due to the above point.

This restores the old behaviour, minus the check for the legacy config, as I don't think that's needed anymore.

### DIFF
--- a/commands/doit.go
+++ b/commands/doit.go
@@ -48,22 +48,30 @@ var (
 	Writer           = os.Stdout
 
 	// PFlag vars
-	APIURL           string // customize API base URL
-	Context          string // current auth context
-	Output           string // global output format
-	Token            string // global authorization token
-	Trace            bool   // toggles http tracing output
-	Verbose          bool
+	APIURL  string // customize API base URL
+	Context string // current auth context
+	Output  string // global output format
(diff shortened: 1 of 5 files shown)


bentranter:
To add some more context for anyone reviewing, we _could_ use `user.Current()`'s `HomeDir` value, and then join on our config path from there. That said, we don't have sufficient test coverage of the logic we're changing here to be able to do that with any confidence given the potential for small differences across platforms. There's also the issue of the somewhat unique handling of `XDG_CONFIG_HOME` that we've accumulated over the years, so there's potential for issues there as well.

Better for now to essentially restore the old code that works than have it break again. Once we have integration tests across each platform we want to support, we can come back to this.

zachgersh:
@bentranter and I discussed a `git revert` of the commit introducing the behavior but because it also contained build changes that we wanted / needed to keep we've ended up with the approach above.

bentranter:
Closing in favour of a different approach.

PR #600, opened 2019-09-26
Compute integration tests

zachgersh:
looking to get more coverage over all of the basic functionality that surrounds our current core contract for our CLI.


zachgersh:
This should be merged after #598 since it is based off of that (after that I will rebase and that commit will go away from the tree above).

zachgersh:
easier to review now that it is only two commits :D

PR #601, opened 2019-09-26
Add documentation generation script.

andrewsomething:
:books:  This PR adds a documentation generation script that can be invoked with `make docs` It produces YAML that will be consumed by the product docs team. It mostly relies on Cobra's built-in doc generation methods, but we need to expose some additional information.


Makefile
```diff
@@ -185,4 +186,9 @@ release:
 	@echo ""
 	@$(MAKE) _release
 
-
+.PHONY: docs
+docs:
+	@echo "=> Generate YAML documenation in ${DOCS_OUT}"
```
bentranter: Typo here in "documenation", should be "documentation"

scripts/gen-yaml-docs.go
```diff
@@ -0,0 +1,74 @@
+/*
+Copyright 2018 The Doctl Authors All rights reserved.
```
bentranter: Should the copyright year be 2019?

Makefile
andrewsomething: :man_facepalming:

PR #602, opened 2019-09-26
os.RemoveAll on temp dir is best effort

zachgersh:
Not really a huge deal if it fails as `OS` will remove temp dir at restart anyway.

### DIFF
--- a/integration/init_test.go
+++ b/integration/init_test.go
@@ -41,6 +41,7 @@ func TestMain(m *testing.M) {
 	if err != nil {
 		panic("failed to create temp dir")
 	}
+	defer os.RemoveAll(tmpDir) // yes, this is best effort only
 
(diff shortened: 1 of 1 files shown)


integration/init_test.go
```diff
@@ -57,10 +57,11 @@ func TestMain(m *testing.M) {
...
 
-	err = os.RemoveAll(tmpDir)
-	if err != nil {
-		panic("failed to cleanup the doctl acceptance artifacts")
-	}
+	// Fails only on windows sometimes for unknown reasons
+	// making best effort for now since worst problem is
+	// that you leave a temp dir around that will get
+	// cleaned up by the OS on a restart.
+	os.RemoveAll(tmpDir)
```
hilary: Makes sense to do this, but let's follow best practices and use [defer](https://golang.org/doc/effective_go.html#defer)

PR #611, opened 2019-09-28
cleanup betacmd and more

hilary:
this PR includes several small commits, each with a small, focused refactor:

* streamlining the global viper getters/setters (which will eventually go away, but at least we can have them readable in the meantime)
* inlining requiredKey, more of that viper
* removing the global effect beta system (later we'll add the ability to make a command beta, which will hide it in help unless that specific beta command is enabled)

### DIFF
--- a/commands/command_config.go
+++ b/commands/command_config.go
@@ -43,9 +43,3 @@ func hiddenCmd() cmdOption {
 	}
 }
 
-// betaCmd tags commands as beta.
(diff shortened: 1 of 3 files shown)


zachgersh:
just need to update the branch before the tests run 👯‍♂

hilary:
@zachgersh I updated, but you'll need to re-approve now that it has been updated. If you could then approve that would be spiffy

PR #613, opened 2019-09-30
adds droplet neighbors test

zachgersh:
- there is also a fix here for a bug that we had
in the neighbors code (arrays + pointers are
always tricky).
- don't actually think we should be using droplet pointers in any of our returns from droplets.go but I didn't want to make that large of a change without adding more tests first.

### DIFF
--- a/do/droplets.go
+++ b/do/droplets.go
@@ -321,7 +321,9 @@ func (ds *dropletsService) Neighbors(id int) (Droplets, error) {
 
 	var droplets Droplets
 	for _, d := range list {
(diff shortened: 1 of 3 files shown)


do/droplets.go
```diff
@@ -321,7 +321,9 @@ func (ds *dropletsService) Neighbors(id int) (Droplets, error) {
 
 	var droplets Droplets
 	for _, d := range list {
```
bentranter: This is probably easier read as,

```go
droplets := make(Droplets, len(list))
for i := range list {
	droplets[i] = Droplet{&list[i]}
}
```

PR #618, opened 2019-09-30
Add region list test

andrewsomething:


integration/region_list_test.go
```diff
@@ -0,0 +1,145 @@
...
+				"--format", "Slug",
+			)
+
+			output, err := cmd.CombinedOutput()
+			expect.NoError(err, fmt.Sprintf("received error output: %s", output))
+			expect.Equal(strings.TrimSpace(regionListFormatOutput), strings.TrimSpace(string(output)))
+		})
+	})
+
+	when("passing a no-header", func() {
```
zachgersh: when "passing no header" probably makes more sense here.

andrewsomething: doh!

PR #619, opened 2019-09-30
removes all traces of plugin

zachgersh:
- we believe that no one has actually written a plugin
and we'd like to simplify the CLI when users aren't
using a particular command
- did a quick look around for the datatype required
to write a plugin and it mostly looked like it
was being used by doctl forks (so no real plugins
published anywhere).


zachgersh:
we going to deprecate first.

PR #623, opened 2019-10-02
remove obsolete windows ssh code

hilary:
Works fine on windows without it; runs green on travis and on my x1.


zachgersh:
@andrewsomething - since you have a windows machine could you verify that this actually works before we merge? I am unsure if we actually have tests around this functionality.

zachgersh:
> Tested it out on my Windows VM, looks good!

as in SSH'ed or run the tests?

andrewsomething:
Only Linux boxes here in the office, but I do have access to a Windows machine at home. I can give it a try if still needed when I'm back home later.

hilary:
I'm going to go ahead and merge on the basis of @bentranter's tests. I'll add an comment to the release notes. If necessary, we can always revert and/or address specific issues as they arise.

PR #631, opened 2019-10-04
Make image name positional and add default image

bentranter:
Updates the new command to create custom images to turn the `--image-name` flag into a positional argument, and adds a default to the image distribution flag.

This is small follow-up to the excellent work done by @VictorAvelar 🎉

### DIFF
--- a/integration/image_create_test.go
+++ b/integration/image_create_test.go
@@ -48,8 +48,7 @@ func testImageCreate(t *testing.T, when spec.G, it spec.S) {
 				"-u", server.URL,
 				"compute",
 				"image",
-				"create",
-				"--image-name", "ubuntu-18.04-minimal",
+				"create", "ubuntu-18.04-minimal",
 				"--image-url", "http://cloud-images.ubuntu.com/minimal/releases/bionic/release/ubuntu-18.04-minimal-cloudimg-amd64.img",
 				"--region", "nyc3",
 			)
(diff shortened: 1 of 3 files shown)


integration/image_create_test.go
```diff
@@ -135,9 +125,7 @@ const imageCreateResponse = `{
 	  "error_message": "",
 	  "id": 38413969,
 	  "name": "ubuntu-18.04-minimal",
-	  "regions": [
-  
-	  ],
+	  "regions": [],
```
zachgersh: nit: should this actually match the same region we requested ie `nyc3`. It's just fake output so probably fine?

bentranter: Hmm I think I'll add the region. Better to have as close to real data as possible in these so it'll make our lives easier once we start on end-to-end tests.

andrewsomething: That's what the response to a create request actually looks like. It seems that the `region` attribute remains empty until the `status` changes to `available`.

bentranter: On second though, the API docs seem to suggest a region isn't returned here... time to test it out

bentranter: Regions is empty:

```json
{
  "image":{
    "created_at":"2019-10-04T17:03:09Z",
    "description":"Cloud-optimized image w/ small footprint",
    "distribution":"Ubuntu",
    "error_message":"",
    "id":53120417,
    "name":"ubuntu-18.04-minimal-test-2",
    "regions":[

    ],
    "status":"NEW",
    "tags":[

    ],
    "type":"custom",
    "updated_at":"2019-10-04T17:03:09Z"
  }
}
```

bentranter: Whoops, missed asb's comment, guess I didn't need to test 😛

PR #632, opened 2019-10-04
consolidate live and test flag semantics

hilary:
- IsSet behaved differently between live and test
- namespacing was done differently between live and test
- alone of all the config methods, IsSet was not namespaced,
  and therefore had different semantics from the other
  methods in that it did not reference viper

### DIFF
--- a/doit.go
+++ b/doit.go
@@ -22,7 +22,6 @@ import (
 	"log"
 	"net/http"
 	"os"
-	"regexp"
 	"strconv"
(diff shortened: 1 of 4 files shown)


doit.go
```diff
@@ -22,7 +22,6 @@ import (
 	"log"
 	"net/http"
 	"os"
-	"regexp"
```
zachgersh: wow is it nice to remove a regex.

PR #634, opened 2019-10-04
adds first database test to the mix

zachgersh:
different approach here using a template (I know you should template out json) since the response is fixed and we just want to double check / return the same values the server is accepting to render them back to the client side.


zachgersh:
please merge #633 first and then I will rebase this (that other commit will drop out).

PR #639, opened 2019-10-09
Add snapshot tests.

andrewsomething:
So I wasn't really able to get around to doing one a day...


zachgersh:
just needs a rebase here - like the new `suite` pattern

PR #641, opened 2019-10-10
adds projects related tests

zachgersh:
This is only missing a few of the tests under the `projects resources` subcommand. Those will follow on here shortly. All of this encouraged by how fast @andrewsomething is working on his integration test PRs 💃

### DIFF
--- a/integration/projects_list_test.go
+++ b/integration/projects_list_test.go
@@ -1,0 +1,143 @@
+package integration
+
+import (
+	"bytes"
+	"fmt"
+	"html/template"
+	"net/http"
+	"net/http/httptest"
+	"net/http/httputil"
+	"os/exec"
+	"strings"
+	"testing"
+
+	"github.com/sclevine/spec"
+	"github.com/stretchr/testify/require"
+)
+
+var _ = suite("projects/list", func(t *testing.T, when spec.G, it spec.S) {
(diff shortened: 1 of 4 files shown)


integration/projects_list_test.go
```diff
@@ -0,0 +1,143 @@
...
+  ],
+  "links": {
+    "pages": {
+      "first": "https://{{.Host}}/v2/projects?page=1",
+      "next": "https://{{.Host}}/v2/projects?page=2",
+      "last": "https://{{.Host}}/v2/projects?page=2"
+    }
+  },
+  "meta": {
+    "total": 1
```
andrewsomething: If this is passing, our pagination logic doesn't depend on it, but this should be `2`. It should be the full total, not just what is displayed on the page. E.g:

```
curl -X GET -H "Content-Type: application/json" -H "Authorization: Bearer $DO_TOKEN" "https://api.digitalocean.com/v2/projects?per_page=1" | jq .meta.total
4
```

Despite only showing the details of one project due to `per_page`, the total count is 4.

zachgersh: yeah, the meta in the response mostly seems to not matter at all. It focuses mainly on what the `links`say are the next few pages regardless of content size.

PR #646, opened 2019-10-10
Fixed doctl's configuration file location for OS X

StevenH1901:
After installing doctl with brew, the configuration file is not at `${XDG_CONFIG_HOME}/doctl/config.yaml` nor `~/.config/doctl/config.yaml` . It is placed in `${HOME}/Library/Application Support/doctl/config.yaml`


hilary:
While few of us are used to thinking of `${HOME}/Library/Application Support` as where application config files go, it is the canonical location for config files on Darwin.

Think of it this way: by default, applications should not have access to each other's configuration files. Apple has always done better when it comes to security and, by default, want apps not to go distributing themselves around your stuff willy-nilly. 😄 

See also: https://golang.org/pkg/os/#UserConfigDir

StevenH1901:
Sorry, I'm slightly confused. I just want to make an update to the documentation so that it's clear if you're installing via brew on OSX, the config files will be placed in `${HOME}/Library/Application Support`. Currently it reads as though it will be placed in `${XDG_CONFIG_HOME}/doctl/config.yaml` which is not correct.

hilary:
No, I'm the one who should apologize! Yup, a documentation update is just what we need. Thanks for catching that.

MattIPv4:
Hey, @StevenH1901 - Thanks a ton for this sweet PR! 😄

PR #650, opened 2019-10-11
this is a dependency bump

zachgersh:
bump dependencies by running `go get -u` for our mod.


bentranter:
As I mentioned, the bump from Viper v1.2.x to v1.4.x scares me a tiny bit, so I'll need to check that for any unexpected changes before I'm super confident merging.

zachgersh:
actually decided to close this until we get more tests around the rest of the commands.

PR #651, opened 2019-10-11
Allow users to switch contexts

bentranter:
Removes the default value from the flag parsing call that sets the global `Context` variable, because for some reason that's the value we use in the commands *if it's set* to decide which context to use to pull the authentication token from the user's config.

Also adds an integration test to check for this case in order to prevent future regressions.

Co-authored-by: Zach Gershman <zachgersh@gmail.com>


hilary:
Fixes https://github.com/digitalocean/doctl/issues/656

fabn:
Upgraded to latest brew version but the issue is still there

```
@mbp.local ➜ Downloads rvm:(ruby-2.6.2)  brew upgrade doctl
==> Upgrading 1 outdated package:
doctl 1.32.3 -> 1.32.3_1
```

bentranter:
Hey @fabn, I've just released v1.33.0 which contains the fix, so it should be updated on Brew within a day or two. If you'd like to use it now, you can grab a binary containing the latest release from https://github.com/digitalocean/doctl/releases/tag/v1.33.0.

fabn:
Ok thanks I'll try

PR #658, opened 2019-10-22
Add improved and expanded doc strings

andrewsomething:
This PR greatly improves and expands on `doctl`'s help output. I'm opening it as a draft as the docs team is still in the process of reviewing the copy, but I wanted to give more folks visibility into their work. I'm also looking for any general feedback on organization so we can land this once the copy is nailed down.

One technical bit to call out: The existing `CmdBuilder` method passes the same text to Cobra for both the long and short description for a command. (See: https://github.com/digitalocean/doctl/blob/master/commands/command.go#L54-L58) So this introduces a `CmdBuilderWithDocs` method to allow us to transition away from the current implementation incrementally.

---

CC: @mgibson323, @johndmulhausen, @hazel-nut


commands/cdns.go
```diff
@@ -29,34 +29,68 @@ func CDN() *Command {
 	cmd := &Command{
 		Command: &cobra.Command{
 			Use:   "cdn",
-			Short: "cdn commands",
-			Long:  "cdn is used to access cdn commands",
+			Short: "Provides access to commands that manage CDNs.",
```
andrewsomething:
```suggestion
			Short: "Create, access, and manage CDN endpoints.",
```

From a UI perspective for these top level short descriptions, it might be nice to be a bit more action orientated and focus on what the task the user wants to acomplish. I'm picking on `cdn` here, but it could be any command. Thoughts @johndmulhausen?

commands/command.go
```diff
@@ -48,14 +48,22 @@ func (c *Command) ChildCommands() []*Command {
...
-	return cmdBuilderWithInit(parent, cr, cliText, desc, out, true, options...)
+	return cmdBuilderWithInit(parent, cr, cliText, desc, desc, out, true, options...)
 }
 
-func cmdBuilderWithInit(parent *Command, cr CmdRunner, cliText, desc string, out io.Writer, initCmd bool, options ...cmdOption) *Command {
+// CmdBuilderWithDocs builds a new command with custom long descriptions.
+// This was introduced to transition away from the current CmdBuilder
+// implementation incrementally. When all commands are built using this
+// function, CmdBuilderWithDocs should replace it.
+func CmdBuilderWithDocs(parent *Command, cr CmdRunner, cliText, shortdesc string, longdesc string, out io.Writer, options ...cmdOption) *Command {
```
zachgersh: Was going to comment that I know the _builder_ pattern has ended up being a real pain as we try to refactor this codebase and get rid of it _but_ I think you took the best approach you could given the constraints of what we have today.

Docs are worlds and away better than what we have today. I don't really see introducing this additional function as making it any harder for us to get rid of this builder pattern and we get the benefit of having nice doc strings.

johndmulhausen: That's the hope, @zachgersh! <3

commands/cdns.go
johndmulhausen: My thinking here was that for the top-level "namespaces" that don't actually do anything on their own it created a sort of double-jeopardy to say that this command "creates... CDN endpoints" when it's really just a bucket of commands and `doctl compute cdn create` would actually do that. 

Having one place that promises x functionality seemed "cleanest" to me. Just my $0.02, happy to shift if that's not making sense to anyone but me

eliza411:
Hey folks! The product docs team is blocked on producing `doctl` content until the auto-generated reference materials from this PR are published. 

Is there anything specific blocking the merge here?

bentranter:
Was going to say LGTM, but it looks like Veronica's already got it 🙂

PR #661, opened 2019-10-22
Adds load balancer tests

zachgersh:
adding associated load balancer tests AND using the nice draft functionality that @andrewsomething reminded me existed. Silly me.

This PR features a slightly different play on the pattern of testing the command alias. If there are only 2 aliases, I have gone ahead and just made two tests. More than two, we do the table test pattern we have in a couple of our other tests.


hilary:
Is there a reason `lb_remove_forwarding_rules.go` is named differently?

PR #665, opened 2019-10-24
Add test for firewall creates

bentranter:
Adds a new, single successful test case for firewall creates.

---

I was going to do these all at once, but I realize I haven't added new test case before, so here's a warmup before adding a bunch 🙂

### DIFF
--- a/integration/firewall_create_test.go
+++ b/integration/firewall_create_test.go
@@ -2,6 +2,7 @@ package integration
 
 import (
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
 	"net/http/httputil"
@@ -25,11 +26,20 @@ var _ = suite("compute/firewall/create", func(t *testing.T, when spec.G, it spec
 		server = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 			switch req.URL.Path {
 			case "/v2/firewalls":
+				if req.Method != http.MethodPost {
+					w.WriteHeader(http.StatusMethodNotAllowed)
+					return
+				}
+
 				auth := req.Header.Get("Authorization")
 				if auth != "Bearer some-magic-token" {
 					w.WriteHeader(http.StatusUnauthorized)
 					return
 				}
+
+				body, err := ioutil.ReadAll(req.Body)
+				expect.NoError(err)
+				expect.JSONEq(firewallCreateRequestBody, string(body))
 
 				w.Write([]byte(firewallCreateResponse))
 			default:
@@ -52,7 +62,7 @@ var _ = suite("compute/firewall/create", func(t *testing.T, when spec.G, it spec
 				"firewall",
 				"create",
 				"--name", "test-firewall",
-				"--inbound-rules", `"protocol:tcp,ports:443"`,
+				"--inbound-rules", `protocol:tcp,ports:443`,
 			)
 
 			output, err := cmd.CombinedOutput()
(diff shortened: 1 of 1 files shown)


integration/firewall_create_test.go
```diff
@@ -0,0 +1,84 @@
...
+		server = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+			switch req.URL.Path {
+			case "/v2/firewalls":
+				auth := req.Header.Get("Authorization")
+				if auth != "Bearer some-magic-token" {
+					w.WriteHeader(http.StatusUnauthorized)
+					return
+				}
+
+				w.Write([]byte(firewallCreateResponse))
```
zachgersh: Can we make sure to double check the method of the request here?

integration/firewall_create_test.go
```diff
@@ -0,0 +1,84 @@
...
+	it.Before(func() {
+		expect = require.New(t)
+
+		server = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+			switch req.URL.Path {
+			case "/v2/firewalls":
+				auth := req.Header.Get("Authorization")
+				if auth != "Bearer some-magic-token" {
+					w.WriteHeader(http.StatusUnauthorized)
+					return
```
zachgersh: I've also started to use a `expect.JSONEq` to validate the post params for the creates. I need to roll back through and add that to the older tests.

integration/firewall_create_test.go
```diff
@@ -0,0 +1,84 @@
...
+			)
+
+			output, err := cmd.CombinedOutput()
+			expect.NoError(err, fmt.Sprintf("received error output: %s", output))
+			expect.Equal(strings.TrimSpace(firewallCreateOutput), strings.TrimSpace(string(output)))
+		})
+	})
+})
+
+const firewallCreateOutput = `ID                                      Name             Status       Created At              Inbound Rules              Outbound Rules    Droplet IDs    Tags    Pending Changes
```
zachgersh: nit - can we turn this into a block
```go
const (
  firewallCreateOutput = ``
  firewallCreateResponse = ``
)
```

I definitely didn't do this in my earlier tests :/

zachgersh:
@bentranter couple _super_ small things but overall great.

zachgersh:
One other small thing of note - make sure to cover the `c` alias if `firewall` establishes that. I've been trying to cover all the command aliases that way we don't slip up and remove one in the future.

integration/firewall_create_test.go
```diff
@@ -0,0 +1,113 @@
...
+			case "/v2/firewalls":
+				if req.Method != http.MethodPost {
+					w.WriteHeader(http.StatusMethodNotAllowed)
+					return
+				}
+
+				auth := req.Header.Get("Authorization")
+				if auth != "Bearer some-magic-token" {
+					w.WriteHeader(http.StatusUnauthorized)
+					return
```
zachgersh: nit: we've got these flipped around in the other tests.

integration/firewall_create_test.go
```diff
@@ -0,0 +1,113 @@
...
+				}
+
+				t.Fatalf("received unknown request: %s", dump)
+			}
+		}))
+	})
+
+	when("the minimum required flags are provided", func() {
+		it("creates a firewall", func() {
+			aliases := []string{"create", "c"}
```
zachgersh: excellent!

PR #674, opened 2019-10-28
bumps viper, along with all of its indirect deps

zachgersh:


zachgersh:
We should cut a release of doctl with the latest kube fix before merging this and moving on 💃

Issue #672, opened 2019-10-28
Create image command needs better help menu


andrewrynhard:
When trying to create a custom image, I got this error:

```
$ doctl compute image create \
                 --region sfo2 \
                 --image-name talos-digital-ocean-tutorial \
                 --image-url https://talos-tutorial.sfo2.digitaloceanspaces.com/digital-ocean.tar.gz \
                 --image-distribution talos
Error: (image.create) command is missing required arguments
```

But no indication in the help menu that an argument is expected (had to find this in the code):

```
$ doctl compute image create --help
Create custom image

Usage:
  doctl compute image create [flags]

Flags:
  -h, --help                        help for create
      --image-description string    Text to describe an image
      --image-distribution string   Custom image distribution (default "Unknown")
      --image-name string           The custom image name (required)
      --image-url string            Custom image retrieval URL (required)
      --region string               Region slug identifier (required)
      --tag-names strings           List of new or existing tags

Global Flags:
  -t, --access-token string   API V2 Access Token
  -u, --api-url string        Override default API V2 endpoint
  -c, --config string         config file (default "/root/.config/doctl/config.yaml")
      --context string        authentication context
  -o, --output string         output format [text|json] (default "text")
      --trace                 trace api access
  -v, --verbose               verbose output
```

PR #677, opened 2019-10-29
image create: Name is a required arg, not flag. (Fixes: #672).

andrewsomething:


andrewsomething:
Rebased on https://github.com/digitalocean/doctl/pull/678, the K8s diff will go away once that's merged.

PR #680, opened 2019-10-30
refactor alias test for floating-ip

zachgersh:
- also added an alias that we've been using but was
missing on floating-ip delete (other commands have set rm)


zachgersh:
as requested by @bentranter in a previous merge.

PR #687, opened 2019-10-31
compute floating-ip-action: adds all tests

zachgersh:
- We've got an inconsistency here on our get action. All other gets are taking the flag `action-id` this one has instead chosen to pass the `action-id` as another positional argument.


zachgersh:
@bentranter / @andrewsomething should we fix this inconsistency? Seems like the right thing to do?

PR #688, opened 2019-10-31
compute volume-action: adds all related tests

zachgersh:
- this also has a bit of a problem (there is a deprecated command here that we should probably remove)

### DIFF
--- a/integration/compute_volume_action_test.go
+++ b/integration/compute_volume_action_test.go
@@ -1,0 +1,175 @@
+package integration
+
+import (
+	"fmt"
+	"io/ioutil"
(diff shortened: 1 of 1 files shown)


zachgersh:
@andrewsomething / @bentranter can we remove one of the deprecated commands here as well? Especially since it has a replacement already in the CLI? 

Seems harmless to me but I'd like to add fixes while merging these tests :D

zachgersh:
@andrewsomething - we could, I'd prefer review it (especially since it's been deprecated for 2 years) especially since as you point out the implementation is duplicative.

PR #689, opened 2019-10-31
volume-action: removes command that was deprecated

zachgersh:
- this command has detach as its parallel already
- command that was removed has been deprecated for over 2 years
- users should have an easy fix (switch to detach) if they
are even using this action


zachgersh:
Discovered this while writing our integration tests. Easy to remove.

bentranter:
Going to wait to hear from @andrewsomething for this one

PR #690, opened 2019-10-31
floating-ip-action: adjust action-id for consistency

zachgersh:
most of our other commands have a flag for action-id, this move to this pattern.


zachgersh:
I've closed this for now - would constitute a breaking change if we were to merge it.

PR #691, opened 2019-10-31
integration: use http consts for methods

zachgersh:


zachgersh:
thank you @bentranter for the glorious suggestion.

PR #693, opened 2019-11-07
Updated looksLikeUUID to only match hyphenated UUIDs (Fixes #461)

### LINKED ISSUE #461: doctl k8s cluster get <clustername> not working for cluster with long name
when I am creating a cluster with a long name "doctl k8s cluster get $CLUSTER" not working, but in list command it is showing the cluster

doctl k8s cluster get $CLUSTER

Error: GET https://api.digitalocean.com/v2/kubernetes/clusters/f8291060b73f4fa7b60586fe51a1d862: 404 cluster not found

doctl k8s cluster list

ID                                      Name                                Region    Version        Status     Node Pools
c4fbbc63-9ebf-4b0e-8ec2-4df84549e3d4    f8291060b73f4fa7b60586fe51a1d862    blr1      1.14.1-do.3    running    f8291060b73f4fa7b60586fe51a1d862-default-pool

neufeldtech:
Cluster names that happened to be 32 characters were [getting treated as UUIDs](https://github.com/google/uuid/blob/16ca3eab7d2086fd5a82993a291cbf3b87fe38b7/uuid.go#L60-L61) which causes issue #461. 

Based on [The tests](https://github.com/digitalocean/doctl/blob/d055080faf145deac6ce9df725c8b27e1a2432fc/commands/kubernetes_test.go#L19), I'm hypothesizing that the DO api is assuming UUID's to always be represented in the hyphenated form (xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx).

This PR updates the `looksLikeUUID()` function to use a strict regex match rather than `uuid.Parse()`'s more lenient checking, which should address the issue in #461 .

### DIFF
--- a/commands/kubernetes.go
+++ b/commands/kubernetes.go
@@ -1679,7 +1679,7 @@ func nodeByName(name string, nodes []*godo.KubernetesNode) (*godo.KubernetesNode
 }
 
 func looksLikeUUID(str string) bool {
-	r := regexp.MustCompile("^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[14][a-fA-F0-9]{3}-[8|9|aA|bB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}$")
+	r := regexp.MustCompile("^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[1-4][a-fA-F0-9]{3}-[8|9|aA|bB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}$")
 	return r.MatchString(str)
 }
 
--- a/commands/kubernetes.go
+++ b/commands/kubernetes.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


bentranter:
Going to give the rest of the team some time in case they'd like to review as well, but I'll merge this before the end of the day.

commands/kubernetes.go
```diff
@@ -1679,8 +1679,8 @@ func nodeByName(name string, nodes []*godo.KubernetesNode) (*godo.KubernetesNode
 }
 
 func looksLikeUUID(str string) bool {
-	_, err := uuid.Parse(str)
-	return err == nil
+	r := regexp.MustCompile("^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-4[a-fA-F0-9]{3}-[8|9|aA|bB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}$")
```
zachgersh: excellent job - one bit, could we have this `4[a-fA-F0-9]{3}` not hardcoded to version 4 of the standard? maybe [1-4] would be better here?

bentranter: To follow up on this, we're not sure if across all of our different teams we're **only** using UUIDv4s here, or if we're using a combination of UUIDv1s and UUIDv4s. Allowing all versions as valid prevents us from accidentally blocking valid UUIDs.

neufeldtech: Updated to support versions 1 through 4

MattIPv4:
Hey, @neufeldtech - Thanks a ton for this sweet PR! 😄

PR #694, opened 2019-11-08
Accept a port of zero for SRV records

bentranter:
Updates the create and edit domain records commands to send a value of zero when the port is set to zero via a flag, and to not send it otherwise.

This required both of these commands not to use `godo`, in order to save us from having to make a breaking change in `godo` to update the port field to be a int pointer instead of just an int.

In the future, this commit should be reverted once we're able to release a v2 of `godo` that addresses this issue properly.

If merged, we should close https://github.com/digitalocean/godo/pull/269.

### DIFF
--- a/do/domains.go
+++ b/do/domains.go
@@ -21,6 +21,11 @@ import (
 	"github.com/digitalocean/godo"
 )
 
+const (
+	domainRecordsPath = "v2/domains/%s/records"
+	domainRecordPath  = "v2/domains/%s/records/%d"
+)
+
 // Domain wraps a godo Domain.
 type Domain struct {
 	*godo.Domain
@@ -192,7 +197,7 @@ func (ds *domainsService) EditRecord(domain string, id int, drer *DomainRecordEd
 		return nil, godo.NewArgError("editRequest", "cannot be nil")
 	}
 
-	path := fmt.Sprintf("v2/domains/%s/records/%d", domain, id)
+	path := fmt.Sprintf(domainRecordPath, domain, id)
 	req, err := ds.client.NewRequest(context.TODO(), http.MethodPut, path, drer)
 	if err != nil {
 		return nil, err
@@ -213,7 +218,7 @@ func (ds *domainsService) CreateRecord(domain string, drer *DomainRecordEditRequ
 		return nil, godo.NewArgError("createRequest", "cannot be nil")
 	}
 
-	path := fmt.Sprintf("v2/domains/%s/records", domain)
+	path := fmt.Sprintf(domainRecordsPath, domain)
 	req, err := ds.client.NewRequest(context.Background(), http.MethodPost, path, drer)
 	if err != nil {
 		return nil, err
--- a/do/domains.go
+++ b/do/domains.go
@@ -15,6 +15,8 @@ package do
 
 import (
 	"context"
+	"fmt"
+	"net/http"
 
 	"github.com/digitalocean/godo"
 )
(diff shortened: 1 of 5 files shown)


do/domains.go
```diff
@@ -157,20 +174,54 @@ func (ds *domainsService) DeleteRecord(domain string, id int) error {
...
-	dr, _, err := ds.client.Domains.CreateRecord(context.TODO(), domain, drer)
+func (ds *domainsService) CreateRecord(domain string, drer *DomainRecordEditRequest) (*DomainRecord, error) {
+	if len(domain) < 1 {
+		return nil, godo.NewArgError("domain", "cannot be empty string")
+	}
+	if drer == nil {
+		return nil, godo.NewArgError("createRequest", "cannot be nil")
+	}
+
+	path := fmt.Sprintf("v2/domains/%s/records", domain)
```
zachgersh: nit: can we pull this as a const?

do/domains.go
```diff
@@ -157,20 +174,54 @@ func (ds *domainsService) DeleteRecord(domain string, id int) error {
...
+		return nil, godo.NewArgError("domain", "cannot be an empty string")
+	}
+	if id < 1 {
+		return nil, godo.NewArgError("id", "cannot be less than 1")
+	}
+	if drer == nil {
+		return nil, godo.NewArgError("editRequest", "cannot be nil")
+	}
+
+	path := fmt.Sprintf("v2/domains/%s/records/%d", domain, id)
```
zachgersh: nit: const as well.

do/domains.go
```diff
@@ -157,20 +174,54 @@ func (ds *domainsService) DeleteRecord(domain string, id int) error {
...
+	req, err := ds.client.NewRequest(context.TODO(), http.MethodPut, path, drer)
 	if err != nil {
 		return nil, err
 	}
 
-	return &DomainRecord{DomainRecord: dr}, nil
+	root := new(domainRecordRoot)
```
zachgersh: nit: can we not use new here and just do a var?

do/domains.go
```diff
@@ -157,20 +174,54 @@ func (ds *domainsService) DeleteRecord(domain string, id int) error {
...
+	}
+
+	path := fmt.Sprintf("v2/domains/%s/records", domain)
+	req, err := ds.client.NewRequest(context.Background(), http.MethodPost, path, drer)
 	if err != nil {
 		return nil, err
 	}
 
-	return &DomainRecord{DomainRecord: dr}, nil
+	root := new(domainRecordRoot)
```
zachgersh: nit: var instead of new

do/domains.go
```diff
@@ -32,6 +34,21 @@ type DomainRecord struct {
...
+// godo due to Go's JSON encoding logic.
+type DomainRecordEditRequest struct {
+	Type     string `json:"type,omitempty"`
+	Name     string `json:"name,omitempty"`
+	Data     string `json:"data,omitempty"`
+	Priority int    `json:"priority"`
+	Port     *int   `json:"port,omitempty"`
+	TTL      int    `json:"ttl,omitempty"`
+	Weight   int    `json:"weight"`
+	Flags    int    `json:"flags"`
```
zachgersh: Let's make it so we hit all of the nullable fields here as pointers so we don't have to come back and change this again if someone decides to make `priority`, `weight`, or `flags` 0.

do/domains.go
bentranter: This was copied exactly from `godo` so I didn't want to change it

do/domains.go
bentranter: I don't think we should. The only user-facing problem was that SRV records (the only ones that can accept a port) cannot be set to 0, so I wanted this "hack" to have the most narrow scope possible. I think updating all these fields risks creating new issues that we don't want to deal with until we create a godov2.

do/domains.go
bentranter: What here would be a `const`?

do/domains.go
bentranter: Same question here as above.

do/domains.go
zachgersh: const domainsRecords = `v2/domains/%s/records`

do/domains.go
zachgersh: it's a nit, so fine either way 💃

do/domains.go
zachgersh: I could go either way on this - if we are going to hack it I feel like we make it the correct hack. I guess we can always change it again when/if we get more issues from users?

do/domains.go
bentranter: Gotcha, will do

do/domains.go
bentranter: Yeah, that's what I figured as well – keep the blast radius of the hack small for now, and increase only if we need to. If there're more issues, that's probably an indication that we should prioritize a godo v2 over expanding the hack. Beyond that, there's also some handling of sending a default TTL if the value is 0 that I don't want to mess with 😛

bentranter:
Pinging @YingCGooi for a review, since this is an alternative solution to our [godo PR](https://github.com/digitalocean/godo/pull/269).

YingCGooi:
Nice work. LGTM

PR #695, opened 2019-11-08
Add support for container registry

kamaln7:
This PR adds support for container registry.


zachgersh:
hey @kamaln7 - know this isn't ready for review but before you submit for review would you please make sure to add an integration test?

Whole bunch of examples to crib from here: https://github.com/digitalocean/doctl/tree/master/integration

We'd also be glad to offer any help you may need.

kamaln7:
Pushed work that depends on https://github.com/digitalocean/godo/pull/273

commands/registry.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+// Registry creates the registry command
+func Registry() *Command {
+	cmd := &Command{
+		Command: &cobra.Command{
+			Use:     "registry",
+			Aliases: []string{"reg", "r"},
+			Short:   "registry commands",
+			Long:    "registry is used to access container registry commands",
```
andrewsomething: As this is being released as a private beta which requires the user to first sign up and then be invited, can we label the descriptions with `[Beta]`? Additionally, it might make sense to hide this from menus at this point.

zachgersh: +1 on the hide

go.mod
```diff
@@ -6,40 +6,31 @@ require (
 	github.com/blang/semver v3.5.1+incompatible
 	github.com/cpuguy83/go-md2man v1.0.10 // indirect
 	github.com/creack/pty v1.1.7
-	github.com/digitalocean/godo v1.22.0
+	github.com/digitalocean/godo v1.25.1-0.20191114163902-6c8415f2abff
```
bentranter: Is this correct? This is pinned to a specific commit, could we just bump this to godo v1.26.0?

bentranter:
@kamaln7 I'm going to wait a bit to let @andrewsomething and @zachgersh have a chance to review before I merge, but we'll get this is today 🙂

go.mod
```diff
@@ -24,7 +24,7 @@ require (
 	github.com/spf13/cobra v0.0.3
 	github.com/spf13/jwalterweatherman v1.1.0 // indirect
 	github.com/spf13/viper v1.4.0
-	github.com/stretchr/testify v1.3.0
+	github.com/stretchr/testify v1.4.0
```
zachgersh: nit: could we not upgrade this in the same PR? I know it's fine but it makes the PR slightly bigger than it has to be.

integration/registry_create_test.go
```diff
@@ -0,0 +1,90 @@
+package integration
```
zachgersh: really appreciate you writing an integration test 💃

commands/registry.go
```diff
@@ -0,0 +1,244 @@
...
+			Long:    "[beta] registry is used to access container registry commands",
+			Hidden:  true,
+		},
+	}
+
+	CmdBuilder(cmd, RunRegistryCreate, "create <registry-name>", "create container registry", Writer)
+
+	CmdBuilder(cmd, RunRegistryGet, "get", "get the container registry", Writer, aliasOpt("g"), displayerType(&displayers.Registry{}))
+
+	cmdRunRegistryDelete := CmdBuilder(cmd, RunRegistryDelete, "delete", "delete the container registry", Writer, aliasOpt("del"))
```
zachgersh: nit: I think we need to add more aliases here as most command stick to `aliasOpt("d", "del", "rm")`

I think get being `g` is correct.

commands/registry.go
```diff
@@ -0,0 +1,244 @@
...
+// RunRegistryLogin logs in Docker to the registry
+func RunRegistryLogin(c *CmdConfig) error {
+	creds, err := c.Registry().DockerCredentials(&godo.RegistryDockerCredentialsRequest{
+		ReadWrite: true,
+	})
+	if err != nil {
+		return err
+	}
+
+	dc := &dockerConfig{}
```
zachgersh: nit: can this please be `var dockerConfig`. The declare and assign isn't actually needed. You would need to put the `&` below on your `dc` in json.Unmarshal.

go.mod
```diff
@@ -6,40 +6,31 @@ require (
...
 	github.com/creack/pty v1.1.7
-	github.com/digitalocean/godo v1.22.0
+	github.com/digitalocean/godo v1.26.0
 	github.com/dustin/go-humanize v1.0.0
 	github.com/fatih/color v1.7.0
 	github.com/gobwas/glob v0.2.3
 	github.com/golang/mock v1.3.1
-	github.com/google/gofuzz v0.0.0-20170612174753-24818f796faf // indirect
-	github.com/google/uuid v1.1.0
+	github.com/google/uuid v1.1.1
```
zachgersh: why did uuid get upgraded in this PR as well? Can we not upgrade this unless it is a direct dependency?

zachgersh:
@kamaln7 overall looks good - I had just a couple of really smallish things.

go.mod
kamaln7: I needed to upgrade it to get access to the YAMLEq function. It's the only change that was made since `v1.3.0`

integration/registry_create_test.go
kamaln7: of course!

go.mod
kamaln7: I'm not sure. I think this change was made when I added `k8s.io/api/core/v1` as a dependency. What do you suggest I do here?

commands/registry.go
kamaln7: 👍🏼

commands/registry.go
kamaln7: 👍🏼

kamaln7: 👍🏼

go.mod
zachgersh: if it is, we can leave it but I wanted to understand why it got upgraded.

commands/registry.go
```diff
@@ -0,0 +1,254 @@
...
+	if !force && AskForConfirm("delete registry") != nil {
+		return fmt.Errorf("operation aborted")
+	}
+
+	return c.Registry().Delete()
+}
+
+// DockerConfigProvider allows a user to read from a remote and local Kubeconfig, and write to a
+// local Kubeconfig.
+type DockerConfigProvider interface {
```
andrewsomething: Might be missing something, but I don't see this referenced anywhere else. I was able to build and use it without this.

kamaln7: You're right, this should've been deleted.

commands/registry.go
```diff
@@ -0,0 +1,254 @@
...
+		cmd.Stdout = c.Out
+		cmd.Stderr = c.Out
+
+		err = cmd.Run()
+		if err != nil {
+			return err
+		}
+	}
+
+	fmt.Println("logged docker in to the registry")
```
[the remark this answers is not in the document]
kamaln7: Good call 👍

eddiezane:
@kamaln7 nice work on this!

Right now your `doctl registry get` command functions as both a `get` and `list` (which as a command is absent). All of the other resources have distinct commands for `list`. Was this design intentional?

kamaln7:
Thanks @eddiezane! Yes, that was intentional. The reason is that an account cannot own more than one registry so `get`ting your registry or `list`ing all registries would have the same result.

Theoretically we could change it to match the other resources to have both `get <name>` and `list` commands, with `get` validating that the registry's name matches the input as some sort of future-proofing. Although right now the API is exposed at `/registry` and not `/registries` 🤷‍♂

PR #702, opened 2019-11-19
add `doctl auth run` subcommand

tdyas:
`doctl auth run COMMAND [args]` will run a subprocess with the current access token placed into the `DIGITALOCEAN_TOKEN` environment variable in the subprocess' environment.

This is useful to invoke `terraform` or other automation scripting that needs to make DigitalOcean API calls without having to reinvent how to securely pass the API token to the subprocess (including avoiding passing it on shell command-line).

Open question:

- Is there a way to mock out the subprocess invocation? I didn't want to start adding interfaces for spawning commands without asking first, but it make may sense to not have the test look for /bin/sh since as written this is really a functional test and not a unit test. Thoughts?

- My actual goal is to have the terraform provider automatically use `doctl's` configured token, so I should probably send a PR to that repo for that functionality. But this subcommand seems useful for now.


bentranter:
Hey @tdyas, thanks for this, it's a really interesting suggestion. I'll follow up with our team to get some proper feedback on this sometime this week, especially considering:
> My actual goal is to have the terraform provider automatically use doctl's configured token

Is something we've been thinking about lately in regards to our Terraform provider.

tdyas:
As background: My only use case thus far has been running Terraform to configure my droplet, so I'd be fine with a solution for Terraform versus this PR. I've been using the following script in lieu of having support in the DigitalOcean Terraform provider or what is proposed by this PR:

```
#!/usr/bin/env python3

import os
import subprocess
import sys

import yaml


def main(args):
    with open(os.path.expanduser('~/.config/doctl/config.yaml')) as f:
        digitalocean_config = yaml.safe_load(f)
    new_env = os.environ.copy()
    new_env['DIGITALOCEAN_TOKEN'] = digitalocean_config['access-token']
    os.execvpe(args[0], args, new_env)


if __name__ == '__main__':
    main(sys.argv[1:])
```

andrewsomething:
Hi @tdyas,

I'd love to know more about your use case. On the Terraform side, I've been a bit reluctant to implement this as the `doctl` config file was not really intended to be used outside `doctl` itself and could potentially change formats. There is an open issue there where I'm looking for feedback: https://github.com/terraform-providers/terraform-provider-digitalocean/issues/259

Also, here's an example of how you can use the `doctl` config file with Terraform today:

```
provider "digitalocean" {
  token = yamldecode(data.local_file.doctl_config.content).access-token
}

data "local_file" "doctl_config" {
  filename = pathexpand("~/.config/doctl/config.yaml")
}
```

PR #705, opened 2019-11-25
Remove github.com/google/uuid dependency

bentranter:
As of https://github.com/digitalocean/doctl/pull/693, we can identify valid UUID's without needing to call `uuid.Parse`. The only thing that our UUID dependency did after that was generate UUIDs in our tests, so this PR removes that dependency and updates any test that used that functionality to use a hardcoded UUID.


zachgersh:
THIS IS PREFECT.

bentranter:
_p r e f e c t_

PR #706, opened 2019-11-25
Add integration test for DIGITALOCEAN_CONTEXT

bentranter:
Adds an integration test for `DIGITALOCEAN_CONTEXT`, an environment variable used to set the name of the current context for a command. This is essentially what the `--context` flag already does.

### DIFF
--- a/integration/auth_test.go
+++ b/integration/auth_test.go
@@ -212,4 +212,33 @@ context: default
 			expect.NoError(err)
 		})
 	})
+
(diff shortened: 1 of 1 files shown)


bentranter:
@zachgersh Given that the contexts are essentially a key value pair between the context name and token value, I wanted to check that setting `DIGITALOCEAN_CONTEXT` loads the correct key – the value doesn't really matter since that's covered by the other tests.

PR #707, opened 2019-11-26
Add OS info to user agent string

bentranter:
Adds the OS info to the user agent string that we set on our `godo` client. The OS info is pulled from the runtime's `GOOS` and `GOARCH` variables. The new user agent string will now look like,

```
doctl/1.22.3 (darwin amd64)
```

or

```
doctl/0.0.0-dev (windows 386)
```

---

If the regex in the test is too much, I can remove it.

### DIFF
--- a/doit.go
+++ b/doit.go
@@ -23,6 +23,7 @@ import (
 	"net/http"
 	"os"
 	"regexp"
+	"runtime"
 	"strconv"
 	"strings"
 
@@ -207,7 +208,7 @@ func (c *LiveConfig) GetGodoClient(trace bool, accessToken string) (*godo.Client
 }
 
 func userAgent() string {
-	return "doctl/" + DoitVersion.String()
+	return fmt.Sprintf("doctl/%s-%s/%s", runtime.GOOS, runtime.GOARCH, DoitVersion.String())
 }
 
 // SSH creates a ssh connection to a host.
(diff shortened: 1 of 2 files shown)


doit.go
```diff
@@ -207,7 +208,7 @@ func (c *LiveConfig) GetGodoClient(trace bool, accessToken string) (*godo.Client
 }
 
 func userAgent() string {
-	return "doctl/" + DoitVersion.String()
+	return fmt.Sprintf("doctl/%s-%s/%s", runtime.GOOS, runtime.GOARCH, DoitVersion.String())
```
andrewsomething:
```suggestion
	return fmt.Sprintf("doctl/%s (%s %s)", DoitVersion.String(), runtime.GOOS, runtime.GOARCH)
```

I think the version should come before the platform. Browsers tend to use a format like: `<product>/<version> (<platform>)`

See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent

bentranter: That sounds okay to me.

I'd originally had it as `<product>/<version>/<platform>` to match the file format for Go binary downloads, but @zachgersh suggested the format in this PR. I'm happy to do anything, so if other folks are happy with your suggestion, I'll do that 🙂

zachgersh: if only @andrewsomething had changed the test as well 💃

andrewsomething: If only @andrewsomething was better with regexp!

PR #712, opened 2019-12-07
[WIP] Add custom bash completion

### LINKED ISSUE #699: Add tab completion for resource names
Example: `doctl kubernetes cluster kubeconfig save [tab]` would suggest existing clusters to tab complete.

`kubectl` does this for all existing resources. Looks like the implementation is [here](https://github.com/kubernetes/kubernetes/blob/f2469a7a2c75a78b8c0b0514652511d105d2b2e7/pkg/kubectl/cmd/cmd.go#L155-L166).

Figuring out how this works should only be difficult for the first resource we implement it on.

eddiezane:
Figured out how to implement this. It required updating Cobra.

Adding the remaining resources won't be too difficult - just tedious.

Still need to figure out how to handle nested completion like:

```
doctl kubernetes cluster node-pool get <cluster-id|cluster-name> <pool-id|pool-name>
```

Closes #699


chrisshroba:
I see that Cobra supports [auto-generating completion functions for multiple shells (bash, zsh, fish, powershell)](https://github.com/spf13/cobra/blob/master/shell_completions.md). Would you consider modifying to use that method, so we get completion in multiple shells for free?  If not, I can try my hand at it.

eddiezane:
@chrisshroba please feel free to take over!

It will be good to hear your experience working with the new Cobra features. We've been looking into this for kubectl.

Hopefully some good guidance for you here:
https://github.com/kubernetes/kubectl/issues/882
https://github.com/kubernetes/kubernetes/pull/93714

PR #714, opened 2019-12-10
ssh: fix for snap

### LINKED ISSUE #475: Provide instructions with permission denied error when using Snap without SSH key access
**Problem**

When installed as an Ubuntu snap, doctl needs to be explicitly given permission to access SSH keys for use with the `doctl compute ssh` sub-command. For details, see:

https://github.com/digitalocean/doctl#using-doctl-compute-ssh

Attempting to SSH to a Droplet before granting those permissions results in an error:

    Error: fork/exec /usr/bin/ssh: permission denied

Unless you have read the README, this error message can be confusing and does not provide a path forward for the user.

**Proposed Solution**

When `doctl compute ssh` is run in a snap environment (hint: https://docs.snapcraft.io/environment-variables) without first being granted access to the ssh-keys interface, we should catch the permission denied error (or check first if possible) and present the user with an error message with a clear path to resolving the problem. For example:

    Error: fork/exec /usr/bin/ssh: permission denied

    Using the doctl Snap? Grant access to the ssh-keys interface to use this command.

        sudo snap connect doctl:ssh-keys :ssh-keys

zachgersh:
fixes #475

### DIFF
--- a/pkg/ssh/ssh.go
+++ b/pkg/ssh/ssh.go
@@ -74,8 +74,8 @@ func (r *Runner) Run() error {
 		_, isSnap := os.LookupEnv("SNAP")
 
 		if os.IsPermission(err) && isSnap {
-			msg := "Using the doctl Snap? Grant access to the ssh-keys interface with this command: sudo snap connect doctl:ssh-keys :ssh-keys"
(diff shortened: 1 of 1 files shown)


zachgersh:
@bentranter - thank you sir. Going to rebuild and build a candidate one moer time before I merge.

PR #731, opened 2020-01-29
Support kubernetes node pool labels

timoreimann:
~The PR depends on the latest (yet unreleased) master version of godo to set labels on Kubernetes requests. **It is only meant to allow the PR to be shared and gather early feedback; the dependency will be reverted to the next godo release once it is out (and block merging until that has happend).**~ (godo has been released and this PR updated accordingly.)

### DIFF
--- a/commands/kubernetes_test.go
+++ b/commands/kubernetes_test.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"fmt"
+	"sort"
 	"testing"
 
 	"github.com/digitalocean/doctl"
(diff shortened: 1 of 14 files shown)


commands/kubernetes_test.go
```diff
@@ -677,110 +698,137 @@ func TestKubernetesNodePool_Create(t *testing.T) {
...
 func TestKubernetesNodePool_Update(t *testing.T) {
-	// by cluster ID
-	withTestClient(t, func(config *CmdConfig, tm *tcMocks) {
-		r := godo.KubernetesNodePoolUpdateRequest{
-			Name:  testNodePool.Name,
-			Count: &testNodePool.Count,
-			Tags:  testNodePool.Tags,
-		}
-		tm.kubernetes.EXPECT().UpdateNodePool(testCluster.ID, testNodePool.ID, &r).Return(&testNodePool, nil)
+	t.Run("by cluster ID", func(t *testing.T) {
```
timoreimann: As a drive-by improvement, I wrapped the test cases into `t.Run()` calls as I found it super hard to tell which tests failed in the old structure.

The individual tests did not change fundamentally. Let me know though if you'd like to me to split the `t.Run()`-related modification out into a separate commit.

Verolop: I think this makes sense for now. If that changes in the future, we can always revisit it. 
Thanks for the clarification!

zachgersh:
I don’t have additional feedback just that we need to ship godo first as Timo mentioned. You got this!

timoreimann:
PR has been updated to use godo v1.30.0 and should be ready for merge. :)

PR #741, opened 2020-02-14
vendor changes (duh) remember to use make (duh)

hilary:


hilary:
closing - need to pull out the original change and figure out what's up

PR #745, opened 2020-02-18
Resolve dependency issues on Go 1.13

bentranter:
Resolves the dependency issue raised in #729.

On Go 1.13, several of the `x/crypto/*` packages have been promoted to the standard library. When you vendor those packages, it looks like `go mod tidy` is removing packages you don't need, so instinctively you'd delete them. Those files are actually for compatibility with the same packages that're now in the standard library, so we've now checked in those files.


Verolop:
Yayy! 🎉🎉🎉

PR #760, opened 2020-03-06
Add expanded docs for registry commands.

andrewsomething:


andrewsomething:
Suggestions applied. Thanks @kamaln7!

PR #761, opened 2020-03-09
Replace CmdBuilderWithDocs with new CmdBuilder.

andrewsomething:
As the comment for `CmdBuilderWithDocs` states, it was introduced to transition away from the current `CmdBuilder` implementation incrementally. The original `CmdBuilder` took one description argument and passed it on to Cobra as both the long description and the short description. Now that all commands are using `CmdBuilderWithDocs`, we can get rid of one of the two separate implementations. This PR:

* Adds a second description argument to the few commands (mostly deprecated ones) left that didn't already have them.
* It gives renames `CmdBuilderWithDocs` to `CmdBuilder` and removes old implementation of the latter.
* Replace all usage of `CmdBuilderWithDocs` with `CmdBuilder`.

### DIFF
--- a/commands/account.go
+++ b/commands/account.go
@@ -30,7 +30,7 @@ For example, ` + "`" + `doctl account get` + "`" + ` retrieves account profile d
 		},
 	}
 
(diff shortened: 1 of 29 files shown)


andrewsomething:
@bentranter Thanks! Rebased this on master and accounted for #736.

PR #764, opened 2020-03-16
auth init: print instructions on how to generate a token

kamaln7:
Before:

```
$ doctl auth init
DigitalOcean access token: 
```

After:

```
$ doctl auth init
Please authenticate doctl for use with your DigitalOcean account. You can generate a token in the control panel at https://cloud.digitalocean.com/account/api/token

Enter your access token:
```


kamaln7:
Thanks, Ben!

PR #765, opened 2020-03-17
use skip_cleanup for deploy in travis

hilary:
as dplv1 uses skip_cleanup


hilary:
Merging to get this fixed today. Review pending.

PR #769, opened 2020-03-20
load-balancers: add enable-proxy-protocol and enable-backend-keepalive flags

anitgandhi:
`enable-proxy-protocol` is/has been GA, but seems like it just wasn't bubbled up to `doctl` from `godo` previously. 

`enable-backend-keepalive` is a new feature, currently in beta.

This also bumps `godo` to v1.33.0.

### DIFF
--- a/args.go
+++ b/args.go
@@ -227,6 +227,8 @@ const (
 	ArgRedirectHttpToHttps = "redirect-http-to-https"
 	// ArgEnableProxyProtocol is a flag that indicates whether PROXY protocol should be enabled on the load balancer.
 	ArgEnableProxyProtocol = "enable-proxy-protocol"
+	// ArgEnableBackendKeepalive is a flag that indicates whether keepalive connections should be enabled to target droplets from the load balancer.
+	ArgEnableBackendKeepalive = "enable-backend-keepalive"
 	// ArgStickySessions is a list of sticky sessions settings for the load balancer.
 	ArgStickySessions = "sticky-sessions"
 	// ArgHealthCheck is a list of health check settings for the load balancer.
--- a/args.go
+++ b/args.go
@@ -225,6 +225,8 @@ const (
 	ArgLoadBalancerAlgorithm = "algorithm"
 	// ArgRedirectHttpToHttps is a flag that indicates whether HTTP requests to the load balancer on port 80 should be redirected to HTTPS on port 443.
 	ArgRedirectHttpToHttps = "redirect-http-to-https"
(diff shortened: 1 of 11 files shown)


args.go
```diff
@@ -225,6 +225,10 @@ const (
 	ArgLoadBalancerAlgorithm = "algorithm"
 	// ArgRedirectHttpToHttps is a flag that indicates whether HTTP requests to the load balancer on port 80 should be redirected to HTTPS on port 443.
 	ArgRedirectHttpToHttps = "redirect-http-to-https"
+	// ArgEnableProxyProtocol is a flag that indicates whether PROXY protocol should be enabled on the load balancer
```
bentranter: Add a period at the end of the doc string so it's consistent with other entries.

```suggestion
	// ArgEnableProxyProtocol is a flag that indicates whether PROXY protocol should be enabled on the load balancer.
```

args.go
```diff
@@ -225,6 +225,10 @@ const (
...
+	ArgEnableProxyProtocol = "enable-proxy-protocol"
+	// ArgEnableBackendKeepalive is a flag that indicates whether keepalive connections should be enabled to target droplets from the load balancer
```
bentranter: Same here.

```suggestion
	// ArgEnableBackendKeepalive is a flag that indicates whether keepalive connections should be enabled to target droplets from the load balancer.
```

go.sum
```diff
@@ -40,6 +40,10 @@ github.com/dgrijalva/jwt-go v3.2.0+incompatible/go.mod h1:E3ru+11k8xSBh+hMPgOLZm
 github.com/dgryski/go-sip13 v0.0.0-20181026042036-e10d5fee7954/go.mod h1:vAd38F8PWV+bWy6jNmig1y/TA+kYO4g3RSRF0IAv0no=
 github.com/digitalocean/godo v1.32.0 h1:ljfhYi/IqDYiZcBYV7nUPzw1Q7NlmPSFDtI69UeRThk=
 github.com/digitalocean/godo v1.32.0/go.mod h1:iJnN9rVu6K5LioLxLimlq0uRI+y/eAQjROUmeU/r0hY=
+github.com/digitalocean/godo v1.32.1-0.20200320152931-d8bc2d10f11e h1:xUc6id4Kfm7j4A7FRlijPLWP4Zj8gTyQ/UVJwa0LDrI=
```
bentranter: Is this 1.32.1 stuff leftover from when you were developing against an unreleased `godo` version, or is this right?

anitgandhi: yup i must have forgotten to `go mod tidy`, i ran `go mod vendor` directly, just noticed that there's a `make vendor` for this. will fix

PR #775, opened 2020-04-02
Add the doctl mascot

bentranter:
Adds the `doctl` mascot. I've updated the README to display the `doctl` project name, the badges, and the mascot so that they're centred. I copied the formatting used by React Native's README, so that we're using a formatting approach that is reliable.

I've saved the mascot in our team's shared Space with the CDN enabled, so that we don't need to unnecessarily increase `doctl`'s total file size with an image.


bentranter:
For reference, here's what this would look like if merged:

image

PR #776, opened 2020-04-05
Fix host architecture detection

benjamb:
The `-p` option to uname is non-portable, and in many cases (such as
mine) this will return 'unknown'. Instead use the POSIX compliant `-m`.

Without this patch, on my x86_64 machine GOARCH ends up being set as 386, thus the build fails as follows:
```
$ make native
=> building doctl via go build

# github.com/digitalocean/doctl/cmd/doctl
loadinternal: cannot find runtime/cgo
/usr/lib/go/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
/usr/bin/ld: skipping incompatible /usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/../../../libgcc_s.so.1 when searching for libgcc_s.so.1
/usr/bin/ld: skipping incompatible /usr/lib/libgcc_s.so.1 when searching for libgcc_s.so.1
/usr/bin/ld: cannot find libgcc_s.so.1
collect2: error: ld returned 1 exit status

make: *** [Makefile:44: _build] Error 2
```


hilary:
Very nice, thanks!

MattIPv4:
Hey, @benjamb - Thanks a ton for this sweet PR! 😄

PR #783, opened 2020-04-14
Ntate/feature/docr ea support

nicktate:

Updates doctl to support current routes that will be offered at the EA release.

#### Dependencies

* [x] https://github.com/digitalocean/godo/pull/325
* [x] https://github.com/digitalocean/godo/pull/326

### DIFF
--- a/commands/registry.go
+++ b/commands/registry.go
@@ -388,10 +388,15 @@ func RunRepositoryDeleteTag(c *CmdConfig) error {
 		return fmt.Errorf("operation aborted")
 	}
 
+	var errors []string
 	for _, tag := range tags {
 		if err := c.Registry().DeleteTag(registry.Name, repository, tag); err != nil {
-			return err
+			errors = append(errors, err.Error())
 		}
+	}
+
+	if len(errors) > 0 {
+		return fmt.Errorf("failed to delete all repository tags: \n%s", strings.Join(errors, "\n"))
 	}
 
 	return nil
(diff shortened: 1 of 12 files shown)


commands/registry.go
```diff
@@ -259,9 +329,123 @@ func RunRegistryLogout(c *CmdConfig) error {
...
+	repository := c.Args[0]
+	tags := c.Args[1:]
+
+	if !force && AskForConfirm(fmt.Sprintf("delete %d repository tag(s)", len(tags))) != nil {
+		return fmt.Errorf("operation aborted")
+	}
+
+	for _, tag := range tags {
+		if err := c.Registry().DeleteTag(registry.Name, repository, tag); err != nil {
+			return err
```
adamwg, line 393: Might be more user-friendly to aggregate errors and return them at the end, so that one deletion failing in the middle of a larger set of tags doesn't block deleting the rest. Not sure if there's any pattern/precedent to follow from other subcommands.

commands/registry.go
```diff
@@ -259,9 +329,123 @@ func RunRegistryLogout(c *CmdConfig) error {
...
+
+	registry, err := c.Registry().Get()
+	if err != nil {
+		return fmt.Errorf("failed to get registry: %w", err)
+	}
+
+	repository := c.Args[0]
+	digests := c.Args[1:]
+
+	if !force && AskForConfirm(fmt.Sprintf("delete %d repository manifest(s) by digest", len(digests))) != nil {
```
adamwg, line 419: At risk of making the confirmation too wordy, it would be nice to include a warning here that any associated tags will also be deleted.

commands/registry.go
nicktate: I had considered this when initially implementing but after a quick (randomized) survey of other commands and deletion functionality, they all exit on first encounter of a failure without processing the rest.

Thoughts @hilary ?

bentranter: How likely is it that a deletion will fail? If it's pretty rare, I think it's okay to keep this as-is.

nicktate: Deletion failure would be pretty rare. 

I agree the experience would be a little nicer not blocking on failures. Since it sounds like we are okay with that process, I opted to update.

PR #788, opened 2020-04-23
Update Container Registry to EA status

nicktate:

Moves Container Registry Commands to EA

> Draft PR until 04/28/20

### DIFF
--- a/commands/registry.go
+++ b/commands/registry.go
@@ -42,9 +42,8 @@ func Registry() *Command {
 		Command: &cobra.Command{
 			Use:     "registry",
 			Aliases: []string{"reg", "r"},
-			Short:   "[Beta] Display commands for working with container registries",
-			Long:    "[Beta] The subcommands of `doctl registry` create, manage, and allow access to your private container registry.",
-			Hidden:  true,
+			Short:   "[EA] Display commands for working with container registries",
+			Long:    "[EA] The subcommands of `doctl registry` create, manage, and allow access to your private container registry.",
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


andrewsomething:
@nicktate  - One thing I just realized is is that we probably want to remove `Hidden:  true,` from the repository sub-command:

https://github.com/digitalocean/doctl/blob/af0cac536acf293d8ffeb4a05a4ac74d2b3cebbe/commands/registry.go#L99

As it is a sub-command of registry, it is already not displayed in the top level help output. So it just being hidden from the list of registry sub-commands:

```
doctl registry -h
[Beta] The subcommands of `doctl registry` create, manage, and allow access to your private container registry.

Usage:
  doctl registry [command]

Aliases:
  registry, reg, r

Available Commands:
  create              Create a private container registry
  delete              Delete a container registry
  get                 Retrieve details about a container registry
  kubernetes-manifest Generate a Kubernetes secret manifest for a registry
  login               Log in Docker to a container registry
  logout              Log out Docker from a container registry
```

nicktate:
@andrewsomething Yep 👍 , this draft PR is removing both hidden fields. 

You're right though it could have been left unhidden previously since it was a subcommand of another hidden field. Are you okay with just waiting for this PR to go through for EA or do you want a separate one to push out a new version with the sub repository command unhidden.

andrewsomething:
doh! Ignore me @nicktate - Somehow I only noticed the copy change when looking at the diff.  Definetly okay with just waiting for this PR to go through for EA. Sorry for the noise.

PR #798, opened 2020-04-29
registry: Add a command to get a Docker config

adamwg:
Add a command, `doctl registry docker-config`, that prints a Docker config. This is useful for users who want to configure a third-party tool such as a build system to access their registry, and mirrors the functionality offered in the control panel.

### DIFF
--- a/args.go
+++ b/args.go
@@ -292,4 +292,7 @@ const (
 	ArgVPCDescription = "description"
 	// ArgVPCIPRange is a VPC range of IP addresses in CIDR notation.
 	ArgVPCIPRange = "ip-range"
+
+	// ArgReadWrite indicates a generated token should be read/write.
+	ArgReadWrite = "read-write"
 )
--- a/commands/registry.go
+++ b/commands/registry.go
@@ -70,7 +70,7 @@ func Registry() *Command {
 	CmdBuilder(cmd, RunRegistryLogout, "logout", "Log out Docker from a container registry",
 		logoutRegDesc, Writer)
 
-	kubeManifestDesc := `This command outputs a YAML-formated Kubernetes secret manifest that can be used to grant a Kubernetes cluster pull access to your private container registry.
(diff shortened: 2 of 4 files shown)


adamwg:
@andrewsomething Good idea, added a test. Will leave one comment inline about something I wasn't sure about with the spec framework.

integration/registry_docker_config.go
```diff
@@ -0,0 +1,104 @@
...
+					t.Fatal("failed to dump request")
+				}
+
+				t.Fatalf("received unknown request: %s", dump)
+			}
+		}))
+	})
+
+	it("prints the returned read-only docker config", func() {
+		expectReadWrite = false
```
adamwg, line 73: Is this safe? And, is there a better way to test both the read-only and read/write cases? (This doesn't feel great, but since the expectations are set up in the `Before` it wasn't clear there was a nicer way to go.)

andrewsomething: I think generally we don't want the test behavior to vary based on something defined outside of the command that is being called. I might have the server do something like this:

```
readWriteParam := req.URL.Query().Get("read_write")
if readWriteParam != "true" {
	w.Write([]byte(registryDockerCredentialsReadOnlyResponse))
} else {
	w.Write([]byte(registryDockerCredentialsReadWriteResponse))
}
```

The contents of the auth string in the response bodies aren't particularly important.

integration/registry_docker_config.go
```diff
@@ -0,0 +1,104 @@
+package integration
```
andrewsomething, line 1: This file should get renamed from `integration/registry_docker_config.go`  to `integration/registry_docker_config_test.go`. That will make the GuardRails scan happy too.

integration/registry_docker_config.go
adamwg: Good idea - done.

Issue #709, opened 2019-12-03
permissions error with doctl snap accessing docker


JayKayy:
doctl is unable to access docker to authenticate to the registry when installed via snap.

```
$ doctl registry login
Error: unable to find the Docker CLI binary. Make sure docker is installed
```
But docker is installed via snap

```
$ docker --version
Docker version 18.09.9, build 1752eb3
```

I believe a slot is needed for doctl snap to support :docker-support interface

zachgersh:
@JayKayy - thank you so much for this issue! Hopefully we'll have a fix for you soon.

swetank01:
I am also facing same problem while running "doctl registry login"
@zachgersh  do we have any progress/solution ?

swetank01:
@JayKayy I found a custom solution :  https://stackoverflow.com/questions/60221651/doctl-is-unable-to-find-docker-binary/60237742#60237742 
Hope it will help you..

gibiansky:
@zachgersh 

I just hit this issue and wrote up my experience debugging it:

https://gist.github.com/gibiansky/f4d7167b67eb88422151ff5d4e5fbf9c

The core issue is that snaps can't access `/usr/bin/docker`, so using `exec.LookPath` to check for Docker via an executable is an issue. I had to get rid of `doctl` snap and install it just as an executable. It seems like there's some sort of special Snap docker support (the `docker` interface / connection) but I couldn't figure out if this would help -- I assume it would let you programmatically act like a Docker client, but still wouldn't let you `exec.LookPath` for the docker executable. 

 It seems like until then just installing Releases from Github is maybe preferred to the the snap install? 

Thanks! Hope my little experience write-up helps someone.

andrewsomething:
I wanted to provide a bit of an update here: As @gibiansky explained so well, the confinement provided by the Snap security model prevents `doctl` from calling out to the Docker binary. Additionally, the `docker` Snap interface only gives access to the Docker daemon, which does not provide the login mechanism. In order to address this, we have rewritten the command to not depend on the Docker binary being present (see: https://github.com/digitalocean/doctl/pull/805). This will still require us to get approval from the Canonical Snap team to get write access to the Docker configuration file (`~/.docker/config.json`). We're hopeful that this will be granted.

In the meantime, installing `doctl` directly is a workaround:

https://github.com/digitalocean/doctl#downloading-a-release-from-github

braedon:
Thanks for the update @andrewsomething.

[...]

PR #805, opened 2020-05-19
registry: Remove dependency on Docker binary for login/logout

andrewsomething:
This removed the runtime dependency on the Docker binary for the registry login/logout commands. It also adds integration tests for them.

This does not resolve https://github.com/digitalocean/doctl/issues/709 on its own, but it should provide a path forward for getting the correct permissions approved by Canonical.


commands/registry.go
```diff
@@ -247,21 +245,20 @@ func RunRegistryLogin(c *CmdConfig) error {
...
 		}
-		cmd := execCommand("docker", args...)
-		cmd.Stdin = strings.NewReader(pass)
-		cmd.Stdout = c.Out
-		cmd.Stderr = c.Out
 
-		err = cmd.Run()
+		cf := dockerconf.LoadDefaultConfigFile(os.Stderr)
+		dockerCreds := cf.GetCredentialsStore(authconfig.ServerAddress)
+		err = dockerCreds.Store(authconfig)
```
bentranter, line 256: I think these two operations call cgo. I'm going to pull down your branch right now to check, but if so, it might make our release process more complicated since we may not be able to use Go's built-in cross compilation as part of goreleaser.

bentranter: Okay – they do call cgo, but it seems like it does not affect our ability to use Go's usual cross compilation tools, so I think this is fine.

PR #810, opened 2020-05-27
Added vpc-uuid arg to lb create and update cmds

### LINKED ISSUE #803: load balancers: Add support for specifying a VPC

When I create a new load balancer using `doctl`, I want to be able to specify which VPC it should belong to.

The `doctl compute load-balancer create` command should accept a `--vpc-uuid` flag that specifies which VPC it should belong to.

**Additional context**

https://godoc.org/github.com/digitalocean/godo#LoadBalancerRequest
https://developers.digitalocean.com/documentation/v2/#create-a-new-load-balancer

scotchneat:
closes #803


integration/projects_resources_get_test.go
```diff
@@ -206,14 +206,14 @@ IP             Region    Droplet ID    Droplet Name
       "sizes": [ "s-1vcpu-1gb" ],
       "features": [ "metadata" ],
       "available": true
-    },
```
bentranter, line 209: Looks like you might need to run `gofmt` on this, or maybe GitHub is rendering this diff weird.

bentranter: Or wait, this is a JSON string, `gofmt` won't help. I think this just needs to tweaked so the whitespace is consistent.

integration/projects_resources_get_test.go
```diff
@@ -246,7 +246,8 @@ ID                                      IP                 Name             Stat
       ],
       "features": [ "install_agent" ],
       "available": true
-    },
+	},
+	"vpc_uuid": "00000000-0000-4000-8000-000000000000",
```
bentranter, line 250: Same here, this is indented slightly more than the other lines.

integration/projects_resources_get_test.go
scotchneat: Yea, that's actually part of a multiline string.  Looks like I had indentation set to use tabs.

integration/projects_resources_get_test.go
scotchneat: Yea, that's actually part of a multiline string.  Looks like I had indentation set to use tabs.

PR #829, opened 2020-06-24
Install kubernetes one click apps on cluster creation

keladhruv:
**WHAT**
This PR allows users to install one click applications when they create a kubernetes cluster

**CHANGES**

- Added --1-click argument to kubernetes cluster create
- Updated unit and integration test cases
- Use Godo to install 1-click apps

**SAMPLE USAGE**
`kubernetes cluster create testcluster --1-clicks metrics-server,linkerd2`


commands/kubernetes.go
```diff
@@ -613,6 +614,20 @@ func (s *KubernetesCommandService) RunKubernetesClusterCreate(defaultNodeSize st
 			s.tryUpdateKubeconfig(kube, cluster.ID, clusterName, setCurrentContext)
 		}
 
+		oneClickApps, err := c.Doit.GetStringSlice(c.NS, doctl.ArgOneClicks) 
+		if err != nil {
+			return err
+		}
+		if len(oneClickApps) > 0 {
```
andrewsomething, line 621: What happens if `wait` is false here? e.g. `doctl kubernetes cluster create --wait=false new-cluster-01 --1-clicks loki`
Is it the install job queued or does the cluster need to be completely up first?

keladhruv: The install job is queued, it waits for the cluster to be up and then installs the application.

andrewsomething: :ok_hand:  Perfect!

keladhruv:
@andrewsomething updated integration test.

PR #835, opened 2020-07-14
apps: add basic apps support

nanzhong:
This build on [godo's v1.39.0 release](https://github.com/digitalocean/godo/releases/tag/v1.39.0) by adding the `apps` subcommand to enable basic crud support for the app platform.

```
> doctl apps --help
[Beta] The subcommands of `doctl app` manage your apps.

Usage:
  doctl apps [command]

Aliases:
  apps, a

Available Commands:
  create            Create an app
  create-deployment Create a deployment
  delete            Deletes an app
  get               Get an app
  get-deployment    Get a deployment
  list              List all apps
  list-deployments  List all deployments
  logs              Get logs
  update            Update an app

Flags:
  -h, --help   help for apps

Global Flags:
  -t, --access-token string   API V2 access token
  -u, --api-url string        Override default API endpoint
  -c, --config string         Specify a custom config file (default "/home/nzhong/.config/doctl/config.yaml")
      --context string        Specify a custom authentication context name
  -o, --output string         Desired output format [text|json] (default "text")
      --trace                 Show a log of network activity while performing a command
  -v, --verbose               Enable verbose output

Use "doctl apps [command] --help" for more information about a command.
```


nanzhong:
🤔 Mhmm, the suggested fix from guardrails doesn't really make sense to me.

We don't want to use `filepath.Base()` here because we want the full path. Checking the file extension also provides little value here.

The spec file that is being opened gets parsed as either json or yaml and then used to construct an api request, so there really isn't a risk of arbitrary execution.

Any suggestions?

andrewsomething:
@nanzhong  - Agreed, that doesn't seem very relevant. Think this should allow the test to pass:

```
specFile, err := os.Open(specPath)  // guardrails-disable-line
```

commands/apps.go
```diff
@@ -0,0 +1,393 @@
...
+
+		io.Copy(c.Out, resp.Body)
+	} else {
+		warn("No logs found for app component")
+	}
+
+	return nil
+}
+
+func parseAppSpec(spec []byte) (*godo.AppSpec, error) {
```
andrewsomething, line 380: Just a suggestion:

Might be nice to do some additional validation here with some feedback for the user. This will send and valid yaml or json as `{"spec":{"name":""}}` to the API.

I could even imagine spec validate sub-command being useful.

nanzhong: The validation currently happens on the server side, so errors will actually flow though. One the reasons we don't want to have client side validation is to allow that spec to change while letting the backend handle all the backwards compatibility. Though there will need to be some sort of coordination between the backend and the generated spec types in godo.

The spec validate subcommand is a good idea and something I can follow up once we add support for an explicit validation endpoint on the backend.

nanzhong:
🤔 I pushed changes and the branch has them but this PR isn't updating....

nanzhong:
Going to try re-creating the PR.

PR #842, opened 2020-07-22
apps: fix yaml spec handling and add support for log following

nanzhong:
This depends on https://github.com/digitalocean/godo/pull/358 and will need to be updated with a vendor change once a godo release is cut; it's currently pointing directly at that commit.

This PR fixes incorrect handling of yaml specs and also adds support for following logs.

### DIFF
--- a/commands/apps_test.go
+++ b/commands/apps_test.go
@@ -311,9 +311,9 @@ func Test_parseAppSpec(t *testing.T) {
 		StaticSites: []godo.AppStaticSiteSpec{
 			{
 				Name: "static",
-				GitHub: godo.GitHubSourceSpec{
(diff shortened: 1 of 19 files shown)


commands/apps_test.go
```diff
@@ -294,3 +295,84 @@ func TestRunAppsGetLogs(t *testing.T) {
 		})
 	}
 }
+
+func Test_parseAppSpec(t *testing.T) {
+	expectedSpec := &godo.AppSpec{
+		Name: "test",
+		Services: []godo.AppServiceSpec{
```
kamaln7, line 302: can we add a component with a generic git source?

PR #851, opened 2020-09-09
apps: add defaults for deployment id and component name

bouk:
This defaults the deployment ID to the current deployment, and the component name to '' so aggregate logs are used.

APPS-1547

### DIFF
--- a/commands/apps.go
+++ b/commands/apps.go
@@ -147,7 +147,7 @@ Only basic information is included with the text output format. For complete app
 	logs := CmdBuilder(
 		cmd,
 		RunAppsGetLogs,
-		"logs <app id> <deployment id> <component name>",
+		"logs <app id> <component name (default 'web')>",
 		"Get logs",
 		`Get component logs for a deployment of an app.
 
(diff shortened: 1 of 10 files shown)


commands/apps.go
```diff
@@ -328,12 +329,26 @@ func RunAppsListDeployments(c *CmdConfig) error {
...
 // RunAppsGetLogs gets app logs for a given component.
 func RunAppsGetLogs(c *CmdConfig) error {
-	if len(c.Args) < 3 {
+	if len(c.Args) < 1 {
 		return doctl.NewMissingArgsErr(c.NS)
 	}
 	appID := c.Args[0]
-	deploymentID := c.Args[1]
-	component := c.Args[2]
+	component := "web"
```
snormore, line 336: Should we default this to the first component in the active deployment instead?

bouk: Yes, that would be better.

kamaln7:
I think we should default to aggregate logs if a component name isn’t set

snormore (dismissed):

bouk:
I've updated it to use godo 1.44

snormore (approved):

nicktate (approved):

andrewsomething (approved):
:+1: LGTM!

PR #852, opened 2020-09-09
apps: add spec commands: get and validate

kamaln7:
This adds two commands:

1. `doctl apps spec get <app-id>` prints an app spec as YAML (default) or JSON
    * an optional deployment ID may be passed to get that specific deployment's spec
    * I added this because `doctl apps get <app id> -o json | jq .[0].spec` is not very convenient plus it requires `jq` or similar to be installed
1. `doctl apps spec validate <filename>` checks whether a given app spec is valid
    * we are working on an API endpoint that validates specs beyond just checking if they're valid JSON or YAML, the plan is to use this endpoint here once it's ready
    * I didn't add integration tests because the API endpoints are covered by `apps get` and `apps get-deployment`. LMK if I should still add them


andrewsomething (changes requested):
Looks like this is failing due to godo changes introduced in your previous PR.

> I didn't add integration tests because the API endpoints are covered by apps get and apps get-deployment. LMK if I should still add them

We would definitely appreciate integration tests for new commands. We'd like to keep coverage here good. Beyond testing the endpoints, its the only real test of the UI itself.

kamaln7:
Addressed the godo issues and added integration tests :+1: Also updated app spec parsing to surface the actual error

andrewsomething (approved):
LGTM!

🙏 Thanks for the tests!

PR #854, opened 2020-09-21
Kubernetes: add support for node pool taints

timoreimann:
This change allows to set node pool taints during node pool create and update operations.

Refs digitalocean/DOKS#3

### DIFF
--- a/commands/kubernetes_test.go
+++ b/commands/kubernetes_test.go
@@ -43,6 +43,7 @@ var (
 			Count:  3,
 			Tags:   []string{"hello", "bye"},
 			Labels: map[string]string{},
+			Taints: []godo.Taint{},
 			Nodes:  testNodes,
 		},
 	}
@@ -337,6 +338,7 @@ func TestKubernetesList(t *testing.T) {
 
 func TestKubernetesCreate(t *testing.T) {
 	testNodePool := testNodePool
+
 	testNodePool.Labels = map[string]string{
 		"key1": "value1",
 		"key2": "value2",
@@ -346,6 +348,24 @@ func TestKubernetesCreate(t *testing.T) {
 		inputLabels = append(inputLabels, fmt.Sprintf("%s=%s", key, val))
 	}
 	sort.Strings(inputLabels)
+
+	testNodePool.Taints = []godo.Taint{
+		{
+			Key:    "key1",
+			Value:  "value1",
+			Effect: "NoSchedule",
+		},
+		{
+			Key:    "key2",
+			Value:  "value2",
+			Effect: "NoExecute",
+		},
+	}
+	var inputTaints []string
+	for _, taint := range testNodePool.Taints {
+		inputTaints = append(inputTaints, taint.String())
+	}
+
 	testNodePool.AutoScale = true
 	testNodePool.MinNodes = 1
 	testNodePool.MaxNodes = 10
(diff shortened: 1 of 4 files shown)


commands/kubernetes_test.go
```diff
@@ -337,6 +338,7 @@ func TestKubernetesList(t *testing.T) {
 
 func TestKubernetesCreate(t *testing.T) {
 	testNodePool := testNodePool
```
MorrisLaw, line 340: What's the point of `testNodePool := testNodePool`?

commands/kubernetes_test.go
```diff
@@ -337,6 +338,7 @@ func TestKubernetesList(t *testing.T) {
...
+
```
MorrisLaw, line 341: extra whitespace?

go.mod
```diff
@@ -52,3 +52,5 @@ require (
 replace github.com/stretchr/objx => github.com/stretchr/objx v0.2.0
 
 replace golang.org/x/crypto => golang.org/x/crypto v0.0.0-20200420201142-3c4aac89819a
+
+replace github.com/digitalocean/godo => github.com/timoreimann/godo v1.7.4-0.20200921135118-82e120c0a69c
```
MorrisLaw: is adding `github.com/timoreimann/godo` just for testing?

commands/kubernetes_test.go
timoreimann: It was already there when I touched the code :) I think it's meant to make a copy of `testNodePool` to work with that does not interfere with other tests. Although this only has limited effect since `testNodePool` includes slices and maps which are always copied by reference.

Probably something that can be addressed in a future PR (presumably by creating a function that Returns a fresh, new `testNodePool` struct on each invocation).

commands/kubernetes_test.go
timoreimann: I added it on purpose to provide for a better visual separation between initializations of more "complex" structs (namely labels and taints).

go.mod
timoreimann: Yes -- see the PR description about the temporary nature of the associated commit.

commands/kubernetes_test.go
MorrisLaw: Yeah, I was just curious. But yeah, I guess that makes sense. Thank you for the clarification!

commands/kubernetes_test.go
MorrisLaw: ok, sgtm!

timoreimann:
CI is failing because of the breaking change in `godo` at digitalocean/godo#367. My suggestion would be that we submit another PR to update the Apps consumer in `doctl` to the `godo` changes once `godo` is released and bumped in `doctl`; afterwards, this PR can be rebased and should go green.

commands/kubernetes_test.go
kamaln7: I added https://github.com/mitchellh/copystructure as a dependency in a recent PR that's been merged already so you could use that to do a deep copy if you wanted to!

timoreimann:
PR has been rebased onto the latest `doctl` master which has godo updated to v1.45.0.

Thanks for the help @kamaln7 👏

andrewsomething (approved):
LGTM!

PR #855, opened 2020-09-23
Add multi-stage build, non-root user, multi-arch builds in Dockerfile

BobyMCbobs:
# Goals

Adjust container build to

- allow the CLI to be built with-in the container build stage
- run as non-root, so it's less privileged
- allow building for different architectures

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,22 @@
 # docker build -t doctl_local --build-arg DOCTL_VERSION=1.23.1 .
 #
-# This Dockerfile exists so casual uses of `docker build` and `docker run` do something sane.
-# We don't recommend using it: If you want to develop in docker, please use `make docker_build`
-# instead.
-
-FROM alpine:3.8
-
-ARG DOCTL_VERSION
-ENV DOCTL_VERSION=$DOCTL_VERSION
-
-RUN apk add --no-cache curl
-
-RUN mkdir /lib64 && ln -s /lib/libc.musl-x86_64.so.1 /lib64/ld-linux-x86-64.so.2
-
+# This Dockerfile builds in the first stage and adds it into a clean base in the second.
+# If you want to develop in docker, please use `make docker_build` instead.
+FROM golang:1.14.6-alpine3.11 AS build
 WORKDIR /app
+COPY . /app/
+ARG GOARCH=""
+RUN CGO_ENABLED=0 GOARCH="$GOARCH" go build \
+  -a \
+  -installsuffix cgo \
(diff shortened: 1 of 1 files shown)


andrewsomething:
Hi @BobyMCbobs. Thanks for the PR!

That specific `Dockerfile` was written back before we published official images to [Docker Hub](https://hub.docker.com/r/digitalocean/doctl). It was intended to allow people to build images with the official released binary. That's why it downloads the binary from GitHub. Now that we provide the official images, I do think it would be good to simplify things in this repo and improve the default `Dockerfile`. We should probably collapse collapse `Dockerfile`, `Dockerfile.build`, and `Dockerfile.cntr` into one file.

The one issue with this PR is that it builds things differently than we do elsewhere. I'd love to keep the builds more consistent so there is less to debug. E.g. It doesn't set [the ldflags needed](https://github.com/digitalocean/doctl/blob/master/scripts/_build.sh) for version on commit info in `doctl version` output.

Though in order to use `make native` and allow for building with other arches, the `GOARCH` variable in the makefile needs to be overridable: 

https://github.com/digitalocean/doctl/blob/master/Makefile#L35-L38

Let us know if you're interested in making those changes. If not, I can probably prioritize getting doing so.

BobyMCbobs:
I would recommend having a Dockerfile like this around, as it makes the builds highly portable.

I'd prefer not to make the changes in the Makefile, but it would be great to have more multi-arch builds (such as _arm64_).

One thing that it's important to address is, running as non-root. The container should not need to run as root. In the changes, it runs as the user _user_ / _1000_.

andrewsomething:
Totally understandable! Though I think this approach will be more maintainable for us in the long run. I'd appreciate your feedback on https://github.com/digitalocean/doctl/pull/860. It includes the non-root user and should allow building multi-arch with the same interface as this PR: 

    docker build --build-arg GOARCH=arm64 . -t doctl

BobyMCbobs:
> Totally understandable! Though I think this approach will be more maintainable for us in the long run. I'd appreciate your feedback on #860. It includes the non-root user and should allow building multi-arch with the same interface as this PR:
> 
> ```
> docker build --build-arg GOARCH=arm64 . -t doctl
> ```

@andrewsomething, this looks pretty good.

Perhaps one last thing would be to make sure that the Go binary is statically compiled as I couldn't see any args in the [build script](https://github.com/digitalocean/doctl/blob/master/scripts/_build.sh#L29) for this. e.g: https://github.com/digitalocean/doctl/pull/855/files#diff-3254677a7917c6c01f55212f86c57fbfR9-R12

PR #862, opened 2020-10-01
commands/doit: Prevent creation of config home when using `--config`

### LINKED ISSUE #833: doctl tries to create ~/.config even if -c provided
**Describe the Issue:**

I'm calling doctl as `doctl -c /etc/doctl/config.yaml`, since it's running in a cronjob, but it still refuses to run because it can't create `/root/.config`. 

It should not care about `$HOME/.config` when `-c` is provided.

(That it fails to create it when installed by snap is #591 -- _this_ bug report is that it shouldn't try, regardless of whether it's installed as a snap or anything else.)

**Environment:**

 - doctl version: 

```
# doctl version
Error: mkdir /root/.config: permission denied
```

Haha, just kidding 🤪

```
doctl version 1.45.1-release
Git commit hash: 1041a0e5
```

 - OS: Ubuntu 18.04
 - Installation method: snap

nywilken:
This change adds a PersistentPreRun command hook to only create the default configuration directory if no custom config path is set.

I see that the oconfigHome() is used by the Kubernetes command for the cache directory so I opted to leave the function as is to not break any other commands that might be relying on the function to create the config directory if it doesn't already exist. 

The test were left unmodified as there appears to be coverage for using a custom cfg file path already, but happy to add a test to confirm that the directory is not created when specifying a custom cfg file. 

Closes #833

### DIFF
--- a/commands/auth.go
+++ b/commands/auth.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"sort"
 	"strings"
 	"syscall"
@@ -217,8 +218,17 @@ func writeConfig() error {
 	return nil
 }
 
+// defaultConfigFileWriter returns a writer to a newly created config.yaml file in the default config home.
+// When using the default config file path the default config home directory will be created; Otherwise
+// the custom config home directory must exist and be writable to the user issuing the auth command.
 func defaultConfigFileWriter() (io.WriteCloser, error) {
 	cfgFile := viper.GetString("config")
+
+	defaultCfgFile := filepath.Join(defaultConfigHome(), defaultConfigName)
+	if cfgFile == defaultCfgFile {
+		configHome()
+	}
+
 	f, err := os.Create(cfgFile)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


andrewsomething:
This looks great @nywilken!

I wonder if we sould just create the directory when it is actually needed and remove the need for the `PersistentPreRun` hook? 

https://github.com/digitalocean/doctl/blob/df1536f7728fdbe375b061a70f68eb7d7cbe6671/commands/auth.go#L220

As pointed out in the issue, the current behavior doesn't allow commands that don't need auth to run and obscure the hint to run `doctl auth intit` for commands that do when it can't write to the directory.

```
$ XDG_CONFIG_HOME=/root/ ./builds/doctl version 
Error: mkdir /root/doctl: permission denied
```

nywilken:
> I wonder if we sould just create the directory when it is actually needed and remove the need for the `PersistentPreRun` hook?

I think it does make sense to go that route over the PresistentPreRun if it doesn't break anything. I took the path of least resistance. I pushed up a few new commits after looking further into how configHome() is being used. 

I think it should cover everything including the example use case you provided 
`XDG_CONFIG_HOME=/root ./builds/doctl version`. 

With that said, maybe the docs, if it doesn't already, should state the the use of a custom config file path must exist and be writable to the user before running `doctl auth init`. Let me know what you think of the updates. Cheers!

andrewsomething (approved):
:+1:  Looks good.

Thanks for this great contribution!

nywilken:
> :+1:  Looks good.
> 
> 
> 
> Thanks for this great contribution!

@andrewsomething thank you! Thanks for the review, and for the help to get the fix in.

PR #864, opened 2020-10-02
Fix typos

devtty1er:
This PR fixes some typos in vendored dependencies. **I'm guessing that you don't want these changes, but maybe this PR will prompt a much-needed migration away from vendoring.** I'm not a Golang developer, but don't Go Modules (which are [already implemented](https://github.com/digitalocean/doctl/blob/master/go.mod)) improve upon vendoring?


andrewsomething:
We do use Go modules for vendoring. So when `go mod vendor` is run, all these changes would be overwritten. Not everyone may agree, but there are a number of reasons to keep the vendor directory in source control. For example, it protects against a dependency being removed completely from the internet. I personally find the contributor experience a bit nicer if I can build the project right away after cloning it regardless of say having `GOPATH` set up.

PR #873, opened 2020-10-05
Fixed golint check

### LINKED ISSUE #869: Top-level doctl package fails golint check
### Overview

The top-level `doctl` package fails the `golint` check.

### How to fix it

1. Clone this repo.
2. If you haven't already, install Go through their [official installer](https://golang.org/dl/).
3. If you haven't already, install the official `golint` tool by following the instructions outlined in their [repo](https://github.com/golang/lint).
4. Run `golint` on the top-level `doctl` package in this repo. To do that, you can run, `golint ./`, and it will report style errors within that package.
5. Fix the issue shown by `golint`. Some are easy enough, but others might require some context. Here are some tips for some of the current issues,
    1. The issue around the `ArgRedirectHttpToHttps` arg will require that you also update where that variable is used after you fix it. This is used only in `doctl/commands/load_balancers.go`.
    2. The issue around the `Major` variable is actually suggesting that the exported variables `Build, Major, Minor, Patch, Label string` all need their own definitions with their own doc strings, instead of being declared on one line. To fix this, you'll need to know an accurate doc string for each value. Here they are,
```go
// Build is set at build time. It defines the git SHA for the current
// build.
Build string

// Major is set at build time. It defines the major semantic version of
// doctl.
Major string

// Minor is set at build time. It defines the minor semantic version of
// doctl.
Minor string

// Patch is set at build time. It defines the patch semantic version of
// doctl.
Patch string

// Label is set at build time. It defines the string that comes after the
// version of doctl, ie, the "dev" in v1.0.0-dev.
Label string
``` 

Once you've solved the `golint` issues, open a pull request to this repo with your changes. If you want, mention this issue in your pull request by writing something like, "Closes #869" which will close this issue upon being merged.

kadern0:
Signed-off-by: Pablo Caderno <kaderno@gmail.com>

Closes #869

### DIFF
--- a/args.go
+++ b/args.go
@@ -239,8 +239,8 @@ const (
 	ArgLoadBalancerName = "name"
 	// ArgLoadBalancerAlgorithm is a load balancing algorithm.
 	ArgLoadBalancerAlgorithm = "algorithm"
-	// ArgRedirectHttpToHttps is a flag that indicates whether HTTP requests to the load balancer on port 80 should be redirected to HTTPS on port 443.
-	ArgRedirectHttpToHttps = "redirect-http-to-https"
+	// ArgRedirectHTTPToHTTPS is a flag that indicates whether HTTP requests to the load balancer on port 80 should be redirected to HTTPS on port 443.
+	ArgRedirectHTTPToHTTPS = "redirect-http-to-https"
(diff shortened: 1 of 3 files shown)


bentranter (approved):
Looks great! Thanks again @kadern0!

PR #874, opened 2020-10-05
Improved apps docstrings

johndmulhausen:
Just nips and tucks and removing the `[Beta]` tag


andrewsomething (approved):
:+1:  Thanks John!

Issue #878, opened 2020-10-08
`doctl apps logs ${ID}` panics (until logs become available)


DazWilkin:
**Describe the Issue:**

I'm a fan of `doctl` and decided to start using it from the get-go as I explore `doctl apps`. I deployed  a `Dockerfile `based app and noticed that, before the deployment completed (and before logs became available), `doctl apps  logs ${ID}` panics. Once logs become available, the command succeeds.

**Environment:**

 - doctl version:
```
doctl version 1.48.0-release
Git commit hash: 03410cbf
```
 - OS:  Linux
```
Description:	Ubuntu 20.04.1 LTS
Release:	20.04
Codename:	focal
```

 - Installation method: Snap

**Additional Details:**

```bash
doctl apps logs ${ID}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xd1da34]

goroutine 1 [running]:
github.com/digitalocean/doctl/commands.RunAppsGetLogs(0xc000416000, 0x0, 0x0)
	/build/parts/doctl/build/commands/apps.go:360 +0x7e4
github.com/digitalocean/doctl/commands.cmdBuilderWithInit.func1(0xc0002ab900, 0xc000384450, 0x1, 0x1)
	/build/parts/doctl/build/commands/command.go:69 +0x146
github.com/spf13/cobra.(*Command).execute(0xc0002ab900, 0xc000384410, 0x1, 0x1, 0xc0002ab900, 0xc000384410)
	/build/parts/doctl/build/vendor/github.com/spf13/cobra/command.go:766 +0x29d
github.com/spf13/cobra.(*Command).ExecuteC(0x17ae5a0, 0xc00032fcc0, 0xc00000e020, 0xc00016df48)
	/build/parts/doctl/build/vendor/github.com/spf13/cobra/command.go:852 +0x2ea
github.com/spf13/cobra.(*Command).Execute(...)
	/build/parts/doctl/build/vendor/github.com/spf13/cobra/command.go:800
github.com/digitalocean/doctl/commands.Execute()
	/build/parts/doctl/build/commands/doit.go:125 +0x34
main.main()
	/build/parts/doctl/build/cmd/doctl/main.go:24 +0x47
```
Once deployment completes:

```bash
doctl apps logs ${ID}
do-apps-rust 2020-10-08T18:22:02.033405233Z Configured for staging.
[...]

PR #879, opened 2020-10-08
app logs: Protect against panic when no deployment is found.

andrewsomething:
This checks that `app.ActiveDeployment` is not nil before assigning `deploymentID`. It also looks for an in progress deployment which is useful in the case someone is running `doctl apps logs --type build $ID` on a newly created app.

Closes: #878


bentranter (approved):
LGTM!

PR #888, opened 2020-10-19
kubernetes - add commands related to docr integration

varshavaradarajan:
cc @adamwg

### DIFF
--- a/commands/kubernetes.go
+++ b/commands/kubernetes.go
@@ -218,6 +218,9 @@ func kubernetesCluster() *Command {
 
 	cmd.AddCommand(kubernetesNodePools())
 
+	cmd.AddCommand(kubernetesRegistryIntegration())
+
+
 	nodePoolDeatils := `- A list of node pools available inside the cluster`
 	clusterDetails := `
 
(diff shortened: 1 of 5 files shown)


commands/kubernetes.go
```diff
@@ -477,6 +480,37 @@ This command deletes the specified node in the specified node pool, and then cre
...
+
+	k8sCmdService := kubernetesCommandService()
+
+	cmdRegistryAdd := CmdBuilder(cmd, k8sCmdService.RunKubernetesRegistryAdd,
+		"add", "Add registry for clusters", `
+This command add registries to the specified cluster(s).`,
+		Writer, aliasOpt("a"))
+
+	AddStringSliceFlag(cmdRegistryAdd, doctl.ArgClusterUUID, "", nil,
+		"uuid of cluster to add registry for; repeat to specify additional clusters.")
```
andrewsomething, line 501: A couple thoughts here.

It might be a bit nicer UX to use arguments rather than flags here. Eg:

```
doctl kubernetes cluster registry add <uuid> <uuid>
```

rather than:
```
doctl kubernetes cluster registry add --cluster-uuid <uuid> --cluster-uuid <uuid>
```



Additionally, other Kubernetes commands take either a name or UUID. Might be nice be consistent with them here. There is a `clusterIDize` helper:

https://github.com/digitalocean/doctl/blob/51e58c7cf399ba128622352642892a1000725813/commands/kubernetes.go#L1808

commands/kubernetes.go
```diff
@@ -1557,6 +1611,15 @@ func buildNodePoolCreateRequestFromArgs(c *CmdConfig, r *godo.KubernetesNodePool
 	return nil
 }
 
+func buildClusterRegistryRequestFromArgs(c *CmdConfig, r *godo.KubernetesClusterRegistryRequest) error {
+	uuids, err := c.Doit.GetStringSlice(c.NS, doctl.ArgClusterUUID)
```
andrewsomething, line 1615: We should either make the UUID flag required or (if we use an arg like suggested above) check the user input here and return an error if no UUID has been passed.

varshavaradarajan: @andrewsomething - decided to go with args.

commands/kubernetes.go
```diff
@@ -477,6 +480,31 @@ This command deletes the specified node in the specified node pool, and then cre
...
+			Short:   "Display commands for integrating clusters with docr",
+			Long:    "The commands under `registry` are for managing DOCR integration with Kubernetes clusters. You can use these commands to add or remove registry from one or more clusters.",
+		},
+	}
+
+	k8sCmdService := kubernetesCommandService()
+
+	CmdBuilder(cmd, k8sCmdService.RunKubernetesRegistryAdd,
+		"add <cluster-id|cluster-name> <cluster-id|cluster-name>", "Add registry for clusters", `
+This command add registries to the specified cluster(s).`,
```
andrewsomething, line 497:
```suggestion
		"add <cluster-id|cluster-name> <cluster-id|cluster-name>", "Add container registry support to Kubernetes clusters", `
This command adds container registry support to the specified Kubernetes cluster(s).`,
```

commands/kubernetes.go
```diff
@@ -477,6 +480,31 @@ This command deletes the specified node in the specified node pool, and then cre
...
+		Writer, aliasOpt("a"))
+
+	CmdBuilder(cmd, k8sCmdService.RunKubernetesRegistryRemove,
+		"remove <cluster-id|cluster-name> <cluster-id|cluster-name>", "Add registry for clusters", `
+This command add registries to the specified cluster(s).`,
```
andrewsomething, line 502:
```suggestion
		"remove <cluster-id|cluster-name> <cluster-id|cluster-name>", "Remove container registry support from Kubernetes clusters", `
This command removes container registry support from the specified Kubernetes cluster(s).`,
```

andrewsomething (changes requested):
Code changes look great. A couple suggestions inline around the doc strings.

varshavaradarajan:
@andrewsomething - thanks for catching that! Addressed your comments. :)

andrewsomething (approved):
:+1:  LGTM, thanks!

PR #896, opened 2020-10-23
SSH Key: make public key accessible

### LINKED ISSUE #894: SSH key public_key not accessible via doctl
When `list`ing or `get`ting a ssh-key, with the `doctl compute ssh-key` subcommand, it's possible to see the fingerprint of the key, but not possible to see the actual public key, which would be useful

an optional format for the ssh-key field, e.g. `--format PublicKey` that would show the public key string `ssh-rsa ...`. The string is usually very long so showing it should be optional

**Additional context**
The value of that string is already in the API request, at the `public_key` attribute

kowshikRoy:
Displays public key. Previously it's not showing public key information.
Screen Shot 2020-10-24 at 3 03 20 AM

With this PR, it shows Public key as well.
Screen Shot 2020-10-24 at 3 04 41 AM

Fixes #894

### DIFF
--- a/commands/displayers/key.go
+++ b/commands/displayers/key.go
@@ -31,13 +31,13 @@ func (ke *Key) JSON(out io.Writer) error {
 
 func (ke *Key) Cols() []string {
 	return []string{
-		"ID", "Name", "FingerPrint",
+		"ID", "Name", "FingerPrint", "PublicKey",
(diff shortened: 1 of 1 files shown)


bentranter (approved):
Thanks @kowshikRoy!

PR #897, opened 2020-10-23
Perf: Make use of cached binary instead of downloading

kowshikRoy:
Previously, while building docker image, any change in this repo will create new intermediate image and all the binary after that need to be downloaded again. 
`ADD . /go/src/github.com/digitalocean/doctl/`

To avoid this, we can download the binary first and add the code later. 

Pros: This saves significant amount of time where network is slow and doing frequent code changes


andrewsomething (approved):
:+1:  Nice improvement!

Thanks for the contribution!

Issue #853, opened 2020-09-21
--auto-upgrade flag requires '=' and documentation doesn't describe usage well


JayKayy:
**Describe the Issue:**

the autoupgrade flag should be used like `--auto-upgrade=false`and not `--auto-upgrade false`. Having a space, instead of '=' appears to accept the input but does not actually afftect the setting. I have provided output below. The `--help` output does not specify to use a space or `=` and can be misleading, especially when an error isn't thrown.

```
doctl kubernetes cluster update lab --auto-upgrade=false
Notice: Cluster updated, fetching new credentials
Notice: Adding cluster credentials to kubeconfig file found in "/home/jkwiatko/.kube/config"
Notice: Setting current-context to do-fra1-lab
ID                                      Name    Region    Version        Auto Upgrade    Status     Node Pools
8dcf2be8-955e-4c8f-850b-5ba5241096fa    lab     fra1      1.18.8-do.0    false           running    pool-eoogxuz0u
```

```
doctl kubernetes cluster update lab --auto-upgrade false
Notice: Cluster updated, fetching new credentials
Notice: Adding cluster credentials to kubeconfig file found in "/home/jkwiatko/.kube/config"
Notice: Setting current-context to do-fra1-lab
ID                                      Name    Region    Version        Auto Upgrade    Status     Node Pools
8dcf2be8-955e-4c8f-850b-5ba5241096fa    lab     fra1      1.18.8-do.0    true            running    pool-eoogxuz0u
```

**Environment:**

 - doctl version: doctl version 1.46.0-release

 - OS: Fedora 32
 - Installation method: Binary downloaded from releases

andrewsomething:
Thanks for the feedback John!

`--auto-upgrade=false` and `--auto-upgrade=true` can be used when you want to be explicit about the values, but passing the value is not necessary with boolean flags. Just excluding the flag altogether should be the same as using `--auto-upgrade=false`, and just passing `--auto-upgrade` alone should be the same as passing `--auto-upgrade=true`.

Any suggestions on how we could make this more clear in the help output?

JayKayy:
oh, perhaps that is clear. Just when I was trying to do it i was surprised that the affect didnt take yet the command seemed to succeed with the ' '.

Your description above makes sense. I instead read it as needing a flag.

I would suggest either adding `--auto-upgrade=<bool>` to the help usage. Or changing the description to:

[...]

PR #905, opened 2020-10-29
Describe --auto-upgrade flag more clearly

mamercad:
Might help #853. The description I'm using is almost verbatim from the [API doc](https://developers.digitalocean.com/documentation/v2/#create-a-new-kubernetes-cluster), which feels like the right thing to do. Also, this helps by clearly stating the default.


andrewsomething (approved):
:+1: Thanks @mamercad!

PR #902, opened 2020-10-27
registry: implement garbage collection subcommands

waynr:
In support of the new garbage collection feature for DOCR I am introducing the following new subcommands:

* `doctl registry garbage-collection start [<registry-name>]`
  * initiate garbage collection on a registry
  * will fail if a garbage collection has already been started
* `doctl registry garbage-collection get-active [<registry-name>]`
  * returns information about the currently-active garbage collection for a registry
* `doctl registry garbage-collection list [<registry-name>]`
  * lists all garbage collections that have been requested for a registry
* `doctl registry garbage-collection cancel [<registry-name>] <gc-uuid>`
  * cancels the specified garbage collection
  * requires an explicit gc to be specified to avoid accidentally cancelling the wrong garbage collection; unlikely currently, but may become an issue in the future when we support multiple registries.

I am opening this PR before finishing the tests so as to provide an opportunity as early as possible for reviewers to provide feedback while I work on the tests.

TODO:

* unit tests
  * [x] RunStartGarbageCollection
  * [x] RunGetGarbageCollection
  * [x] RunListGarbageCollection
  * [x] RunUpdateGarbageCollection
* integration tests
  * [x] `doctl registry garbage-collection start`
  * [x] `doctl registry garbage-collection get-active`
  * [x] `doctl registry garbage-collection list`
  * [x] `doctl registry garbage-collection cancel`
* [ ] deploy new API endpoints on the backend after it is clear that no more changes will be necessary in response to doctl/godo feedback
* [x] merge [associated godo PR](https://github.com/digitalocean/godo/pull/400)

PR #906, opened 2020-10-29
Improve error messaging when more args than expected are passed.

andrewsomething:
When a user provides multiple arguments to a command then only accepts one, we provide unclear error messaging. For example:

```
$ doctl databases sql-mode get aaa-bbb-ccc ddd-eee-fff
Error: (sql-mode.get) command is missing required arguments
```

This PR adds a new helper method, `ensureOneArg`, that returns a `MissingArgsErr` if no args are passed or a `TooManyArgsErr` (introduced in https://github.com/digitalocean/doctl/pull/902) if more than one is passed. This allows for clearer messaging:

```
$ ./builds/doctl databases sql-mode get aaa-bbb-ccc ddd-eee-fff
Error: (sql-mode.get) command contains unsupported arguments

$ ./builds/doctl databases sql-mode get
Error: (sql-mode.get) command is missing required arguments
```

### DIFF
--- a/commands/actions.go
+++ b/commands/actions.go
@@ -189,8 +189,9 @@ func isZeroTime(t time.Time) bool {
 
 // RunCmdActionGet runs action get.
 func RunCmdActionGet(c *CmdConfig) error {
-	if len(c.Args) != 1 {
-		return doctl.NewMissingArgsErr(c.NS)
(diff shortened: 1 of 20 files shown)


bentranter (approved):
Looks good (once Travis passes in about two weeks 😴)

Issue #908, opened 2020-10-30
`doctl auth init` not working on Windows 10 powershell


JohnathanWeisner:
**Describe the Issue:**

Not able to authorize doctl after installing doctl via the instructions https://www.digitalocean.com/docs/apis-clis/doctl/how-to/install/

After following the instructions in the link above and getting to the step when you must run:
`doctl auth init --context <NAME>`
And supplying a newly created valid API key I'm met with this error:
```
doctl auth init --context batman
Please authenticate doctl for use with your DigitalOcean account. You can generate a token in the control panel at https://cloud.digitalocean.com/account/api/tokens

Enter your access token:
Validating token... invalid token

Error: Unable to use supplied token to access API: Get "https://api.digitalocean.com/v2/account": net/http: invalid header field value "Bearer \x16" for key Authorization
```

**Environment:**

 - doctl version: ```doctl version 1.48.1-release
Git commit hash: a5b1cd2
release 1.49.0 is available, check it out!```
 - OS: Windows 10 Pro Version 1903
 - Installation method: `Invoke-WebRequest https://github.com/digitalocean/doctl/releases/download/v1.48.1/doctl-1.48.1-windows-amd64.zip -OutFile ~\doctl-1.48.1-windows-amd64.zip`

**Additional Details:**

Trace
```
Validating token... doctl: 2020/10/29 21:47:56 -> "GET /v2/account HTTP/1.1\r\nHost: api.digitalocean.com\r\nUser-Agent: doctl/1.48.1-release (windows amd64) godo/1.46.0\r\nAccept: application/json\r\nContent-Type: application/json\r\nAccept-Encoding: gzip\r\n\r\n"
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x48 pc=0x86458a]

goroutine 1 [running]:
net/http/httputil.DumpResponse(0x0, 0xc0003db201, 0x0, 0x10e3620, 0xc0003e4650, 0xc7, 0x0)
        /home/travis/.gimme/versions/go1.14.9.linux.amd64/src/net/http/httputil/dump.go:305 +0x4a
github.com/digitalocean/doctl.(*recorder).RoundTrip(0xc000395cc0, 0xc0003db200, 0xc000395cc0, 0x0, 0x0)
        /home/travis/gopath/src/github.com/digitalocean/doctl/recorder.go:47 +0x183
[...]

PR #909, opened 2020-10-30
Check error in RoundTrip.

andrewsomething:
In https://github.com/digitalocean/doctl/issues/908, when the reporter tried to produce a trace, it resulted in a panic. We're not checking this error.


bentranter (approved):
👍 pending tests

PR #911, opened 2020-11-03
Support subscription tiers for DOCR

adamwg:
As of GA, creating a container registry requires specifying a subscription tier. Add support for listing available subscription tiers, and add a `--subscription-tier` option for create, defaulting to the "basic" plan.


bentranter (approved):
LGTM!

PR #912, opened 2020-11-09
kubernetes: Stop using "master" to describe the control plane

adamwg:
Along with the broader Kubernetes community and industry, we're moving away from using the word "master" to describe the Kubernetes control plane. We've never explicitly referred to either in our API or control panel, but we do mention it in the docs for `doctl k8s cluster get`. Update those docs to be more accurate by describing the k8s API server rather than the nebulous "master node".


andrewsomething (approved):
:+1:  LGTM

Issue #916, opened 2020-11-14
zsh autocompletion uses non portable sed extensions


m-242:
**Describe the Issue:**

We trying to use the zsh completion generated by doctl with a sed version not supporting the `_filedir` extension (so busybox sed, that is used by default on alpine linux)

**Environment:**

 - doctl version:
both the last release:
```doctl version 1.52.0-release
Git commit hash: af5bae4
```
and the one packaged on alpine:
```
doctl version 0.0.0-dev
```
 - OS: alpine linux edge and 3.1.12
 - Installation method: using the [distribution package](https://pkgs.alpinelinux.org/package/edge/testing/x86_64/doctl) and downloading from [the release page](https://github.com/digitalocean/doctl/releases/tag/v1.52.0).

**Additional Details:**
Here is a detailed how to to reproduce the bug. It assumes you have docker installed and running, and your containers can access internet.

### Running a clean alpine container with docker:
`docker run -it alpine /bin/sh`

### Setup in the container
```shell
# apk add zsh
# zsh
# autoload -Uz compinit
# compinit
# wget https://github.com/digitalocean/doctl/releases/download/v1.52.0/doctl-1.52.0-linux-amd64.tar.gz
# tar -C /usr/local/bin -zxvf doctl-1.52.0-linux-amd64.tar.gz
```

The bug in itself
```shell
# source <(doctl completion zsh)
```

You should get the following output
```
sed: bad regex '[[:<:]]_filedir[[:>:]]': Unknown character class name
```

Note: If we used the alpine provided [doctl-zsh-completion](https://pkgs.alpinelinux.org/package/edge/testing/x86_64/doctl-zsh-completion) package, the same bug would appear when trying to use the autocomplete (ie, `doctl [tab]`)

It seems the bug comes from [line 600 of the completion.go file](https://github.com/digitalocean/doctl/blob/5d55b2d49a8bc47043d64c5ccd6f879c6cf10535/commands/completion.go#L600).

PR #918, opened 2020-11-16
completion/zsh: Use correct word boundary pattern for BusyBox sed

andrewsomething:
To test from this branch, run:

```
make build
docker run -it --mount type=bind,source="$(pwd)"/builds,target=/app alpine /bin/sh
apk add zsh
zsh
autoload -Uz compinit
compinit
source <(./app/doctl completion zsh)
```

No error should be reported and completion should function as expected.

Closes: https://github.com/digitalocean/doctl/issues/916

### DIFF
--- a/commands/completion.go
+++ b/commands/completion.go
@@ -405,7 +405,7 @@ complete -c doctl -n '__fish_seen_subcommand_from projects' -a update    -d "Upd
 # Completions for the 'doctl projects create' command
 complete -c doctl -n '__fish_seen_subcommand_from projects create' -l description -d "Dscription of the project"
 complete -c doctl -n '__fish_seen_subcommand_from projects create' -l environment -d "Environment in which your project resides."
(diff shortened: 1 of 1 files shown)


commands/completion.go
```diff
@@ -585,7 +585,8 @@ autoload -U +X bashcompinit && bashcompinit
 # use word boundary patterns for BSD or GNU sed
 LWORD='[[:<:]]'
 RWORD='[[:>:]]'
-if sed --help 2>&1 | grep -q GNU; then
+SEDHELP=$(sed --help 2>&1)
+if echo ${SEDHELP} | grep -q GNU || echo ${SEDHELP} | grep -q BusyBox; then
```
andrewsomething, line 589: This could potentially be simplified, but I'm not certain if `grep -qE "GNU|BusyBox"` is portable to FreeBSD sed (or Mac).

bentranter (approved):
LGTM!

PR #921, opened 2020-11-19
Migrate tests to GitHub Actions

bentranter:
Migrates the tests from Travis to GitHub Actions.

Also fixes a `gofmt` failure so that those checks would pass.

Tested at https://github.com/bentranter/doctl/pull/1.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,8 @@ DOCS_OUT = $(shell echo $${DOCS_OUT:-$(my_d)/builds/docs/yaml})
 UNAME_S := $(shell uname -s)
 UNAME_M := $(shell uname -m)
 
+GOFILES_NOVENDOR = $(shell find . -type f -name '*.go' -not -path "./vendor/*")
+
 GOOS = linux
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -103,6 +105,12 @@ shellcheck:
 	@echo ""
 	@scripts/shell_check.sh
 
+.PHONY: gofmt_check
+gofmt_check:
```
andrewsomething, line 109: :+1:  Thanks for this! We should probably add `go lint` at some point as well. Though the results aren't exactly pretty right now...

andrewsomething (approved):
:shipit:  LGTM!

bentranter:
Travis seems stuck, so I'm gonna merge this.

Makefile
bentranter: Definitely, I thought about adding a ticket to address it during Hacktoberfest, but realized there were way too many warnings to ask anyone from the community to work on fixing it.

PR #922, opened 2020-11-24
registry: add support for garbage collection types

waynr:
The intent here is to provide users of DOCR a way to optionally garbage collection untagged manifests while retaining the previous default behavior of only garbage collecting unreferenced blobs.


bentranter (approved):
LGTM!

PR #926, opened 2020-12-09
apps: add `apps propose` command, expand spec validation

kamaln7:
1. Added `doctl apps propose`
2. Updated `doctl apps spec validate`
3. Updated other commands to accept reading app specs from stdin by setting the path to `-`

### `doctl apps propose`

API docs: https://developers.digitalocean.com/documentation/v2/#propose-an-app-spec

Usage: `doctl apps propose --spec spec.yaml [--app app-id]`

#### available app name
```
App Name Available?    Is Static?    Free Static Site Usage    $/month    $/month on higher tier    $/month on lower tier
yes                    no            3 of 3                    5.00       17.00                     n/a
```

#### unavailable app name

```
App Name Available?    Suggested App Name     Is Static?    Free Static Site Usage    $/month    $/month on higher tier    $/month on lower tier
no                     sample-dockerfile-2    no            3 of 3                    5.00       17.00                     n/a
```

### `doctl apps spec validate`

Added a `--schema-only` flag. Defaults to false. If set, the old behavior of only validating that the spec marshals into a `godo.AppSpec` object is retained.

If not set, it calls the `ProposeApp` API endpoint.

In both cases the command now prints the validated spec instead of a `The app spec is valid` message.

#### some file that's not a spec
```
Error: Failed to parse app spec: json: cannot unmarshal string into Go value of type godo.AppSpec
exit status 1
```

#### valid yaml but invalid spec
```
Error: POST https://api.digitalocean.com/v2/apps/propose: 400 error validating app spec field "services.git.repo_clone_url": services.git.repo_clone_url in body is required
exit status 1
```

#### happy path - valid spec

The spec I passed in doesn't set the `routes` field
```
name: sample-hugo
region: nyc
static_sites:
- build_command: hugo
  environment_slug: hugo
  envs:
  - key: HUGO_VERSION
    scope: BUILD_TIME
    value: 0.74.3
  git:
    branch: main
    repo_clone_url: https://github.com/digitalocean/sample-hugo.git
  name: hugo
  routes:
  - path: /
```

### DIFF
--- a/commands/displayers/apps.go
+++ b/commands/displayers/apps.go
@@ -302,7 +302,7 @@ func (r AppProposeResponse) Cols() []string {
 
 	cols = append(cols, []string{
 		"AppIsStatic",
-		"StaticSites",
+		"StaticApps",
 		"AppCost",
 		"AppTierUpgradeCost",
 		"AppTierDowngradeCost",
@@ -316,7 +316,7 @@ func (r AppProposeResponse) ColMap() map[string]string {
 		"AppNameAvailable":     "App Name Available?",
 		"AppNameSuggestion":    "Suggested App Name",
 		"AppIsStatic":          "Is Static?",
-		"StaticSites":          "Free Static Site Usage",
+		"StaticApps":           "Static App Usage",
 		"AppCost":              "$/month",
 		"AppTierUpgradeCost":   "$/month on higher tier",
 		"AppTierDowngradeCost": "$/month on lower tier",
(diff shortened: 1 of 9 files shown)


ChiefMateStarbuck (dismissed):
lgtm! I'll need an approval from another app engineer to merge this

commands/displayers/apps.go
```diff
@@ -284,3 +284,75 @@ func (is AppInstanceSizes) JSON(w io.Writer) error {
...
+		"AppIsStatic":          "Is Static?",
+		"StaticSites":          "Free Static Site Usage",
+		"AppCost":              "$/month",
+		"AppTierUpgradeCost":   "$/month on higher tier",
+		"AppTierDowngradeCost": "$/month on lower tier",
+	}
+}
+
+func (r AppProposeResponse) KV() []map[string]interface{} {
+	staticSites := fmt.Sprintf("%s of %s", r.Res.ExistingStaticApps, r.Res.MaxFreeStaticApps)
```
andrewsomething, line 327: Looks like the API does not return the `existing_static_apps` field at all if no static sites are in use. Since it is a string, not an int, the default value is blank rather than `0`. So you end up with:

```
Free Static Site Usage
of 3 
```

You could handle it here, but you might want to fix that on the API side.

andrewsomething (commented):
Looks good! One comment in line. Let me know how you want to approach it. If you're fixing on the API side, I'll go ahead and merge.

commands/displayers/apps.go
kamaln7: Good catch @andrewsomething! I updated the representation of free static apps

andrewsomething (approved):
:shipit:

PR #929, opened 2020-12-17
Better Error Messaging in Droplet Actions

### LINKED ISSUE #928: Improve error message when providing non-numeric droplet id
**Describe the Issue:**

As a new user of `doctl`, I mistakenly tried to use using the droplet's name rather than the id when triggering actions. Since the error isn't clear, it took me longer than I'd like to admit to figure out what I was doing wrong.

For example

```shell
% doctl compute droplet-action power-off not-an-id
Error: strconv.Atoi: parsing "not-an-id": invalid syntax
```

An error such as this would have been helpful:

```shell
% doctl compute droplet-action power-off not-an-id
Error: expected <droplet-id> to be numeric, got "not-an-id"
```

**Environment:**
 - doctl version: doctl version 1.54.0-release
 - OS: macOS Big Sur
 - Installation method: Homebrew

ChiefMateStarbuck:
fixes #928


commands/droplet_actions.go
```diff
@@ -544,3 +548,15 @@ func RunDropletActionSnapshot(c *CmdConfig) error {
 
 	return performAction(c, fn)
 }
+
+// contextualAtoi cleans the error output of Atoi calls
+func contextualAtoi(s, resource string) (int, error) {
+	n, err := strconv.Atoi(s)
+	if err == nil {
+		return n, nil
```
scotchneat, line 556: You should check if `n` is positive before returning here since the returned error on line 559 indicates a positive integer is expected.

commands/droplet_actions.go
```diff
@@ -544,3 +548,15 @@ func RunDropletActionSnapshot(c *CmdConfig) error {
 
 	return performAction(c, fn)
 }
+
+// contextualAtoi cleans the error output of Atoi calls
+func contextualAtoi(s, resource string) (int, error) {
```
andrewsomething, line 553: Looks like this would be generally useful where ever we're using `strconv.Atoi` with user input. Might make sense to put it somewhere that reflects that. Maybe something like `commands/validations.go`?

commands/droplet_actions.go
scotchneat: I think you can ignore my suggestion as that's not really the scope of the issue reported, and it would make sense to implement that across the entire cli, not just for droplet actions.

Other than @andrewsomething request bellow, I think this looks good.

commands/validations.go
```diff
@@ -0,0 +1,31 @@
...
+*/
+
+package commands
+
+import (
+	"fmt"
+	"strconv"
+)
+
+// contextualAtoi cleans the error output of Atoi calls
```
bentranter, line 21: This comment needs to start with a capital letter for GoDoc and golint to be happy.
```suggestion
// ContextualAtoi cleans the error output of Atoi calls
```

commands/validations.go
```diff
@@ -0,0 +1,31 @@
...
+func ContextualAtoi(s, resource string) (int, error) {
+	n, err := strconv.Atoi(s)
+	if err == nil {
+		return n, nil
```
bentranter, line 25: It's worth adding a check that `n` is positive, maybe something like,
```suggestion
		if n < 0 {
		    return 0, fmt.Errorf("expected %d to be a postive integer", n)
		}
		return n, nil
```

bentranter (changes requested):
Two small comments around the `ContextualAtoi` function.

commands/validations.go
ChiefMateStarbuck: 🤔  i never receive errors like this on my machine. I'll have to look into that. Thank you!

commands/validations.go
```diff
@@ -0,0 +1,34 @@
...
+	"fmt"
+	"strconv"
+)
+
+// ContextualAtoi cleans the error output of Atoi calls
+func ContextualAtoi(s, resource string) (int, error) {
+	n, err := strconv.Atoi(s)
+	if err == nil {
+		if n < 0 {
+		    return 0, fmt.Errorf("expected %d to be a postive integer", n)
```
bentranter, line 26: Typo in `postive`, should be `positive`:
```suggestion
		    return 0, fmt.Errorf("expected %d to be a positive integer", n)
```

commands/validations.go
```diff
@@ -0,0 +1,34 @@
...
+		}
+		return n, nil
+	}
+	if _, ok := err.(*strconv.NumError); ok {
+		return 0, fmt.Errorf(`expected %s to be a postive integer, got "%s"`, resource, s)
```
bentranter, line 31: Same here:
```suggestion
		return 0, fmt.Errorf(`expected %s to be a positive integer, got "%s"`, resource, s)
```

bentranter (changes requested):
Small typo, after that LGTM

ChiefMateStarbuck:
fixes #932

bentranter:
@ChiefMateStarbuck Since this updates that crypto dependency, could you also update the vendored dependencies? You can do so by running `go mod vendor`.

ChiefMateStarbuck:
> @ChiefMateStarbuck Since this updates that crypto dependency, could you also update the vendored dependencies? You can do so by running `go mod vendor`.

Updated 👍

commands/validations.go
bentranter: @ChiefMateStarbuck Not sure if the diff is coming out weird for me or not, but it looks like this typo is still here. Could you fix the typo here as well (assuming it's not already fixed?)

bentranter (changes requested):
I think there's still one typo – could you confirm @ChiefMateStarbuck? Thanks!

ChiefMateStarbuck:
> I think there's still one typo – could you confirm @ChiefMateStarbuck? Thanks!

thought i caught that. Fixed 👍

bentranter (approved):
LGMT! Thanks @ChiefMateStarbuck!

PR #931, opened 2020-12-17
Bump k8s.io dependencies to v0.20.0.

andrewsomething:
Closes: https://github.com/digitalocean/doctl/issues/927


bentranter (approved):
Seems good, not going to even try to load the "Files changed" tab when there's 526 files changed 😛

PR #940, opened 2021-01-20
Add GET operations to volume-actions doctl

danaelhe:
added subcommand 'compute volume-action get' to check on status of action.

### DIFF
--- a/commands/volume_actions.go
+++ b/commands/volume_actions.go
@@ -59,6 +59,22 @@ func VolumeAction() *Command {
 		},
 	}
 
+	actionDetail := `
+
+	- The unique numeric ID used to identify and reference a volume action.
+	- The status of the volume action. This will be either ` + "`" + `in-progress` + "`" + `, ` + "`" + `completed` + "`" + `, or ` + "`" + `errored` + "`" + `.
+	- A time value given in ISO8601 combined date and time format that represents when the action was initiated.
+	- A time value given in ISO8601 combined date and time format that represents when the action was completed.
+	- The resource ID, which is a unique identifier for the resource that the action is associated with.
+	- The type of resource that the action is associated with.
+	- The region where the action occurred.
+	- The slug for the region where the action occurred.
+	`
+
+	cmdVolumeActionsGet := CmdBuilder(cmd, RunVolumeActionsGet, "get <image-id>", "Retrieve the status of an image action", `Use this command to retrieve the status of an image action, including the following details:`+actionDetail, Writer,
+		displayerType(&displayers.Action{}))
+	AddIntFlag(cmdVolumeActionsGet, doctl.ArgActionID, "", 0, "action id", requiredOpt())
+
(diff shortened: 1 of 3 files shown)


commands/volume_actions.go
```diff
@@ -59,6 +59,22 @@ func VolumeAction() *Command {
...
+	- A time value given in ISO8601 combined date and time format that represents when the action was completed.
+	- The resource ID, which is a unique identifier for the resource that the action is associated with.
+	- The type of resource that the action is associated with.
+	- The region where the action occurred.
+	- The slug for the region where the action occurred.
+	`
+
+	cmdVolumeActionsGet := CmdBuilder(cmd, RunVolumeActionsGet, "get <image-id>", "Retrieve the status of an image action", `Use this command to retrieve the status of an image action, including the following details:`+actionDetail, Writer,
+		displayerType(&displayers.Action{}))
+	AddIntFlag(cmdVolumeActionsGet, doctl.ArgActionID, "", 0, "action id", requiredOpt())
```
andrewsomething, line 76: Here you add an `--action-id` flag, but in `RunVolumeActionsGet` it looks like you expect the action ID to be a positional argument. 

    doctl compute volume-action get aaa-111-bbb-222-ccc  --action-id 12345

vs

    doctl compute volume-action get aaa-111-bbb-222-ccc 12345

We should probably stick to using the flag to stay consistent with the existing `doctl compute droplet-action get` command.

https://github.com/digitalocean/doctl/blob/67904a7b84498c0e34bbe5b24ef0340155c9cadf/commands/droplet_actions.go#L69

andrewsomething (changes requested):
Looking good! One comment in-line about using a flag vs a positional argument for the action ID.

commands/volume_actions.go
danaelhe: Makes sense- have applied changes in most recent commit.

andrewsomething: Looks like there are some merge conflicts that need to be cleaned up.

commands/volume_actions.go
```diff
@@ -59,6 +59,22 @@ func VolumeAction() *Command {
...
+	- The status of the volume action. This will be either ` + "`" + `in-progress` + "`" + `, ` + "`" + `completed` + "`" + `, or ` + "`" + `errored` + "`" + `.
+	- A time value given in ISO8601 combined date and time format that represents when the action was initiated.
+	- A time value given in ISO8601 combined date and time format that represents when the action was completed.
+	- The resource ID, which is a unique identifier for the resource that the action is associated with.
+	- The type of resource that the action is associated with.
+	- The region where the action occurred.
+	- The slug for the region where the action occurred.
+	`
+
+	cmdVolumeActionsGet := CmdBuilder(cmd, RunVolumeActionsGet, "get <volume-action-id>", "Retrieve the status of an volume action", `Use this command to retrieve the status of an volume action, including the following details:`+actionDetail, Writer,
```
andrewsomething, line 74:
```suggestion
	cmdVolumeActionsGet := CmdBuilder(cmd, RunVolumeActionsGet, "get <volume-id>", "Retrieve the status of an volume action", `Use this command to retrieve the status of an volume action, including the following details:`+actionDetail, Writer,
```

andrewsomething (approved):
:shipit:  Looks great!

PR #944, opened 2021-01-21
apps: disallow unknown fields during spec parsing

nicktate:

* Updates decoding for `app spec` to ignore unknown fields. 

#### Additional Context

This should help alleviate confusion where the user is on an older version of doctl before the relevant godo models were updated for a particular feature. Previously, the create/update would succeed but the values would be stripped from the payload.

### DIFF
--- a/commands/apps.go
+++ b/commands/apps.go
@@ -14,6 +14,7 @@ limitations under the License.
 package commands
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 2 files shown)


kamaln7 (approved):

bentranter (approved):
LGTM pending CI, thanks again @nicktate!

ChiefMateStarbuck (approved):

PR #945, opened 2021-01-22
Adding compute volume-action list cli

danaelhe:
extending the volume-action cli with a "list" sub command.


andrewsomething (changes requested):
Looks like it needs a rebase on `main`

andrewsomething (approved):
:+1:  Looks great!

PR #955, opened 2021-02-15
kubernetes - enable surge upgrade by default during cluster creation

varshavaradarajan:
Description: Now that surge upgrades is stable, enable surge upgrades by default as it is a better user experience during upgrades. 

cc @adamwg


adamwg (approved):
lgtm.

andrewsomething (approved):
:+1:

Issue #954, opened 2021-02-14
doctl apps logs follow mode unexpectedly exits


Craytor:
**Describe the Issue:**
When I run `doctl apps logs c2ad307c-6f44-43f5-9b89-a80909b03e3d -f`, doctl will get all the current logs and print them to the terminal. However, doctl does not follow the logs as intended with `-f`. doctl will wait about a minute and then exit the log listening process.

**Environment:**
 - doctl version: doctl version 1.56.0-release
 - OS: macOS Catalina
 - Installation method: Homebrew

**Additional Details:**
Not sure of any - if there are any questions please let me know and I can assist.

andrewsomething:
Thanks for raising this issue @Craytor We've just release v1.57.0 which should resolve this issue.

fedebaier:
I'm running v1.57.0 on macOS Big Sur, installed with Homebrew, and this is still happening.

PR #959, opened 2021-02-20
apps logs: Use websocket connection for reading logs.

andrewsomething:
Using the HTTP connection to retrieve app logs runs into issues with Cloudflare buffering. Instead, we should be listening on the websocket connection. This adds a new listen package for listening to a websocket and outputting its messages. The apps logs command now uses it.

Fixes: https://github.com/digitalocean/doctl/issues/954


kamaln7 (approved):
Thanks for taking care of this! 🙌🏼

ChiefMateStarbuck (approved):
LGTM

jcodybaker (approved):
This is great. Thanks for addressing this.

PR #960, opened 2021-02-22
whitesource: set vulnerableCheckRunConclusionLevel to success

andrewsomething:
Whitesource consitently reports false positives for Kubernetes related libraries.
It does not understand its versioning scheme and has some issues with Go mod's
dependency handling in geneneral. For example, it is currently reporting vulnerablities
for kubernetes-1.15.1-beta.0 while we have k8s.io/client-go v0.20.0. It does not
know that:

> For each v1.x.y Kubernetes release, the major version (first digit) would remain 0.

https://github.com/kubernetes/client-go#versioning

See for example: #843, #812. and #811

This causes confusion in PR reviews and leads to alert fatigue.

Setting vulnerableCheckRunConclusionLevel to success means:

> the conclusion status of a WhiteSource Security Check will always be
> 'Success', even if the check fails.

https://whitesource.atlassian.net/wiki/spaces/WD/pages/556007950/WhiteSource+Bolt+for+GitHub#WhiteSourceBoltforGitHub-vulnerableCheckRunConclusionLevel

WhiteSource will continue to open new issues when it finds a _new_
vulnerablity, but it will no longer cause every PR to have a failed
check for things we've already determined are false positives. I think
this is a good trade off.


ChiefMateStarbuck (approved):
NICE

PR #966, opened 2021-03-08
registry - generate a DOSecret operator friendly secret manifest

varshavaradarajan:
The Kubernetes secret manifest  we generate makes it easy for users to pull private images into their DOKS cluster. Now that we have the dosecret-operator to make the secret available in all namespaces, generate operator friendly manifests for users to apply into their DOKS cluster. Applying the manifest generated here to a DOKS cluster is the same as doing:

```
doctl kubernetes cluster registry add
```

cc @adamwg , @timoreimann

PR #967, opened 2021-03-09
Fix integration test focus.

andrewsomething:
Reviewing https://github.com/digitalocean/doctl/pull/966, I realized that the integration tests were not running. They were inadvertently set to focus on one test, preventing the others from running:

https://github.com/digitalocean/doctl/blob/9b40e1a24c0b12a02c2fd971320d23e4ce3a2db1/integration/database_firewall_remove_test.go#L50

This PR fixes that and updates tests that should have seen changes.


bentranter (approved):
LGTM, thanks for catching this!

Issue #964, opened 2021-03-02
Add watch option to Apps Platform create-deployment


sjparkinson:
**What is the problem this feature would solve? Please describe.**

I would like a means to wait for an Apps Platform deployment to complete and get a relevent status code back from the command line, to inform a job as part of a build pipeline in GitHub Actions.

**Describe the solution you'd like**

When calling `doctl apps create-deployment` I'd like to pass a `--watch` option so that I can get the command to return once the deployment is completed, with a relevent return status code.

There's a few ways to cut this for sure, so whatever makes the most sense.

**Additional context**

`kubectl` supports this by calling `kubectl rollout status --watch <name>` after applying a change.

https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#-em-status-em-

sjparkinson:
Feel free to get an idea of what I'm trying to achieve in GitHub Actions from https://github.com/sjparkinson/isitup.org/blob/73f9585569060b1de4d4415f5f09f2152f26f37b/.github/workflows/deploy.yaml#L40-L132.

That's my old Kubernetes deployment job, recreating it to tie togther a Apps Platform deployment to a deployment in GitHub is the main aim.

At the moment the auto-deploy from GitHub feature is a bit too keen. In Heroku there is a good option to wait for status checks to pass first before starting a deployment. I figure a `--watch` option is a good interim solution until App Platform is more fully integrated with GitHub and all it's deployment and checks APIs.

andrewsomething:
This sounds like something we should have. It would be consistent with other resources that can take awhile to become ready. Both `doctl compute droplet create` and `doctl kubernetes cluster create` have `--wait` flags that block until the resource has been successfully created.

We have similar logic to what will be needed here in the Terraform provider:

https://github.com/digitalocean/terraform-provider-digitalocean/blob/1f777298a6e1d91b2cfccb6c877fd0970d7d74cf/digitalocean/resource_digitalocean_app.go#L159

Shanduur:
Is there any progrss on this issue? If no one started the work yet, I could try to implement it.

andrewsomething:
@Shanduur It is not actively being worked on. If you're interested in implementing it, I'd be very happy to review/merge. Let me know if you have any questions!

Shanduur:
@andrewsomething should the wait flag be passed during the integration testing?

PR #969, opened 2021-03-18
feat: wait flag for apps create-deployment command

Shanduur:
In this pull request issue #964 is addressed:
- `--wait` flag was added to the `doctl apps create-deployment`
- defaults to `false`
- waits for deployment to enter `ACTIVE` phase
- on phases `ERROR`, `CANCELED` and `UNKNOWN` returns error

### DIFF
--- a/commands/apps.go
+++ b/commands/apps.go
@@ -120,7 +120,7 @@ Creating an app deployment will pull the latest changes from your repository and
 		displayerType(&displayers.Deployments{}),
 	)
 	AddBoolFlag(deploymentCreate, doctl.ArgAppForceRebuild, "", false, "Force a re-build even if a previous build is eligible for reuse")
-	AddBoolFlag(deploymentCreate, doctl.ArgCommandWait, "", true,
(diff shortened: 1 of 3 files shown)


andrewsomething (commented):
This is great stuff @Shanduur! Thanks for the contribution!

To answer your question from  https://github.com/digitalocean/doctl/issues/964, it would be good to include this in the integration testing. Are you up for it?

Shanduur:
Maybe in future PR, but I would have to check how are you doing the integration testing - I tried passing it straight away, but (as I suppose) this caused output mismatch. Additionally the response time is not deterministic, so this should probably be also took into account. As `doctl kubernetes cluster create` does not include `--wait` flag in integration testing, I skipped this part for now.

andrewsomething (approved):
:+1:  I pushed some additional commits adding basic tests that exercise the feature a bit. Take a look if you want to see the ways we are currently testing. 

Thanks again for this great contribution!

MattIPv4:
Hey, @Shanduur - Thanks a ton for this sweet PR! 😄

PR #971, opened 2021-03-22
Upgrade to Go 1.16

scotchneat:
Creating a draft PR for feedback while I continue testing the changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [ 1.15.x ]
+        go-version: [ 1.15.x, 1.16.x]
         os: [ ubuntu-latest, macos-latest, windows-latest ]
     runs-on: ${{ matrix.os }}
     steps:
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 builds
 doit
 .vscode
+.idea/
 project.json
 .cover
 dist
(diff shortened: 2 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -10,7 +10,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [ 1.15.x ]
+        go-version: [ 1.15.x, 1.16.x]
```
scotchneat, line 13: Should I remove `1.15.x` here?

ChiefMateStarbuck (commented):
Looks good. 🥇

ChiefMateStarbuck (approved):
i have no objections to the removal of 1.15.x.
It would cut the test suite down.

.github/workflows/ci.yml
scotchneat: I'm leaving it and updating the CONTRIBUTION.md to reflect 1.15 as the minimal version of Go.

scotchneat:
I tested the snap build locally. It doesn't actually use local code (it clones the repo instead) but it is using the updated snapcraft.yaml so it's built with Go 1.16.

```
root@cesar:~# gdb /snap/doctl/current/bin/doctl.real
...
(gdb) p 'runtime.buildVersion'
$1 = 0xf4e6fd "go1.16.2"
```

Initially, I was concerned the build scripts weren't pulling local changes but the only change included is to use Go 1.16 for the build, which this confirms. 

We'll need to rework the scripts to support building a local snap using local code changes (instead of cloning) to be able to include this as a proper step in the contribution process but that's not required for this change.

cc: @andrewsomething

andrewsomething (approved):
:+1:  LGTM!

Issue #975, opened 2021-04-09
Can't (un)set result limit count


WoLfulus:
I made some automation using `doctl` but I just found out (when things broke 😆) that the results are getting limited to 25 and there's no flag to "unlimit" it, thus I assumed it would return all of them.

Basically I have ~40 apps on DigitalOcean App Platform and `doctl apps ls` is showing me only those first 25.

Am I missing something? --help doesn't show any flags to set the return limit

Using latest version on windows and downloaded from releases.

andrewsomething:
DigitalOcean API responses are paginated. Unfortunately, it looks like this was not handled when apps support as initially added to doctl. We'll need to make a few changes to support this.

Thanks for flagging this issue!

rijkvanzanten:
@andrewlouis93 Do you happen to have a clearer overview of the changes that are needed to support this? I'm down to give it a shot in a PR, but I'm not (at all) familiar with the internal of `doctl` (yet) 🙂

andrewsomething:
Hey @rijkvanzanten! Thanks for the offer, but the work is already in progress. Take a look at the linked PRs to see what this involves. If there's another issue that piques your interest, please feel free to tag me. I'd be very happy to break down any needed changes for you.

rijkvanzanten:
So fast! Thanks 👍🏻

andrewsomething:
The fix for this should now be available: https://github.com/digitalocean/doctl/releases/tag/v1.59.0

PR #976, opened 2021-04-09
apps: Support pagination.

andrewsomething:
Requires a new godo release for: https://github.com/digitalocean/godo/pull/443

Fixes: https://github.com/digitalocean/doctl/issues/975

### DIFF
--- a/do/apps.go
+++ b/do/apps.go
@@ -74,11 +74,32 @@ func (s *appsService) Get(appID string) (*godo.App, error) {
 }
 
 func (s *appsService) List() ([]*godo.App, error) {
-	apps, _, err := s.client.Apps.List(s.ctx, nil)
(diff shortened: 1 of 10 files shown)


scotchneat:
When I test `go run cmd/doctl/main.go apps ls`, I get the following response:

```
Error: GET https://api.digitalocean.com/v2/apps?page=1&per_page=200: 400 cannot retrieve more than 25 items per page
```

It looks like the default `perPage` value is 200 and can't be configured when calling `PaginateResp()`.

andrewsomething:
@scotchneat Yup, there is still a change that needs to happen on the API side to make Apps pagination consistent with the rest of the API.

scotchneat (approved):
lgtm!

PR #977, opened 2021-04-13
Fix fish documentation

2bithacker:
Update the fish documentation to correct for calling it zsh, as well as using the correct startup file, and changing the bash-only `<()` syntax to fish's `(|psub)`


andrewsomething (approved):
:+1:  Thanks Chip!

PR #982, opened 2021-04-27
Update README with the appropriate command to remove configuration

### LINKED ISSUE #979: Fix the configuration removing command
**Describe the Issue:**

```
rm -rf $HOME/Library/Application Support/doctl
```

this is won't work as expected, because of the space in `Application Support`. It will be treated as two different arguments, and this command will completely remove `$HOME/Library/Application` directory if that exists on the user machine, without asking.

Should be quoted:
```
rm -rf "$HOME/Library/Application Support/doctl"
```

scotchneat:
This updates the command listed in the README.md to completely remove the configuration (on MacOS) with quotes around the path so the space in `Application Support` doesn't get confused for an argument separation.

```
rm -rf "$HOME/Library/Application Support/doctl"
```
Fixes #979

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,32 +54,31 @@ Use "doctl [command] --help" for more information about a command.
 
 See the [full reference documentation](https://www.digitalocean.com/docs/apis-clis/doctl/reference/) for information about each available command.
 
-- [doctl](#doctl---)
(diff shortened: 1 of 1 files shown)


andrewsomething (approved):
:shipit:

PR #983, opened 2021-04-30
go.mod: remove uneeded replace

peebs:
`go mod why -m github.com/strechr/objx` reveals that this package doesn't
actually get built or included in the doctl command. 

```
# github.com/strechr/objx
(main module does not need module github.com/strechr/objx)
```

This replace was added to force the objx package to v0.2.0 because of security vuln from issue: https://github.com/digitalocean/doctl/issues/735

 `go list -mod=mod -m github.com/stretchr/objx` reveals that v0.2.0 is the resolved version that would be used anyway even if the package import graph included it in a build of doctl in the future.

Removing this replace statement also enables building of doctl via the new `go install pkg@version` command which won't build replace statements. This command is available in go 1.16.

To make this change I deleted the replace. Then I ran `go mod vendor` and then `go mod tidy`.


andrewsomething (approved):
:+1:  Looks good. Thanks for the contribution!

MattIPv4:
Hey, @peebs - Thanks a ton for this sweet PR! 😄

PR #990, opened 2021-06-08
Allow node-pool args to be empty

Retler:
This PR adds the capability of ignoring empty node-pool arguments. Some background to motivate this:

The documentation doesn't mention that the --num-nodes string should not terminate with a `;` character. Therefore, when trying to use an argument like `--num-nodes:"name=pool-name;count=3;"` the command fails with an error, because the trailing `;` character results in an empty arg.

### DIFF
--- a/commands/kubernetes.go
+++ b/commands/kubernetes.go
@@ -1720,7 +1720,8 @@ func parseNodePoolString(nodePool, defaultName, defaultSize string, defaultCount
 		Labels: map[string]string{},
 		Taints: []godo.Taint{},
 	}
-	for _, arg := range strings.Split(nodePool, argSeparator) {
+	trimmedPool := strings.TrimSuffix(nodePool, argSeparator)
+	for _, arg := range strings.Split(trimmedPool, argSeparator) {
 		if arg == "" {
 			continue
 		}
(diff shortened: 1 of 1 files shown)


commands/kubernetes.go
```diff
@@ -1721,6 +1721,9 @@ func parseNodePoolString(nodePool, defaultName, defaultSize string, defaultCount
 		Taints: []godo.Taint{},
 	}
 	for _, arg := range strings.Split(nodePool, argSeparator) {
```
andrewsomething, line 1723: How about trimming any trailing `;`?

```go
	trimmedPool := strings.TrimSuffix(nodePool, argSeparator)
	for _, arg := range strings.Split(trimmedPool, argSeparator) {
```

It's a bit of a corner case, but it provides a proper error if someone were to run:

    doctl kubernetes cluster create test --node-pool ";"

andrewsomething (commented):
Thanks for this change! I left a comment in-line with a suggestion.

commands/kubernetes.go
Retler: That seems reasonable. Thanks for the review, I have now modified the code.

andrewsomething (approved):
:+1:  LGTM; thanks again!

MattIPv4:
Hey, @Retler - Thanks a ton for this sweet PR! 😄

Issue #970, opened 2021-03-19
Add wait option to app create and update commands


andrewsomething:
**What is the problem this feature would solve? Please describe.**

In CI and/or scripting it is often useful for commands that create resources to block until they become ready. For example, both `doctl compute droplet create` and `doctl kubernetes cluster` create have `--wait` flags that block until the resource has been successfully created. Recently, the `doctl apps create-deployment` command has grown a `--wait` flag as well.

Creating or updating an App can also start a new deployment and should have the same behavior. 

**Describe the solution you'd like**

`doctl apps create` and  `doctl apps update` should accept a  `--wait` flag that will block until any new deployment they trigger reaches the `ACTIVE` phase.

**Additional context**

This is a continuation of the theme from https://github.com/digitalocean/doctl/issues/964 (PR: https://github.com/digitalocean/doctl/pull/969).

PR #991, opened 2021-06-10
adding the --wait flag to the apps create command

danaelhe:
per github issue #970 - adding a wait flag for doctl apps create command. The integration test is taking 30 seconds to complete. This causes a timeout error- if timeout is removed it works. Looking for a second pair of eyes to help determine why the test takes so long.


bentranter:
I've only had a quick look, but it seems like this only hits the API twice, so you'd think that this should take about 20 seconds with the configured timeout.

Because the timeout is at the beginning of the loop, it looks like what's happening is roughly,
1. Call the function in `doctl`, wait 10s
2. Perform the first API request, wait another 10s,
3. Perform the next API request, wait another 10s.

So because it's at the beginning, you pay a 10s penalty before the test can start. This (I think) is a limitation of the approach of ranging over a channel created by a ticker. You might be able to switch to a regular `for` loop and use `time.Sleep` instead of having to tick, since you don't need the concurrency benefits of a ticker as these code paths are single-threaded. This could look something roughly like,

```go
// waitForAppCreateRunning waits for a app to be running.
func waitForAppCreateRunning(apps do.AppsService, appID string) (*godo.App, error) {
	// If we check the status every ten seconds, and we want a timeout of
	// 30 minutes, we can compute the number of timeouts before we reach ~30
	// minutes using the math below.
	const maxAttempts = 10 * 6 * 30
	attempts := 0

	printNewLineSet := false
	var deploymentID string
	var app *godo.App

	for i := 0; i < maxAttempts; i++ {
		if attempts != 0 {
			fmt.Fprint(os.Stderr, ".")
			if !printNewLineSet {
				printNewLineSet = true
				defer fmt.Fprintln(os.Stderr)
			}
		}

		if deploymentID == "" {
			app, err := apps.Get(appID)
			if err != nil {
				return nil, fmt.Errorf("Error trying to read app deployment state: " + err.Error())
			}

			if app.InProgressDeployment != nil {
				deploymentID = app.InProgressDeployment.ID
			}
		} else {
			deployment, err := apps.GetDeployment(appID, deploymentID)
			if err != nil {
				return nil, errors.New("Error trying to read app deployment state: " + err.Error())
			}

			allSuccessful := deployment.Progress.SuccessSteps == deployment.Progress.TotalSteps
			if allSuccessful {
				app, _ = apps.Get(appID)
				return app, nil
			}

			if deployment.Progress.ErrorSteps > 0 {
				return nil, fmt.Errorf(fmt.Sprintf("error deploying app (%s) (deployment ID: %s):\n%s", appID, deployment.ID, godo.Stringify(deployment.Progress)))
			}
		}

		attempts++
		time.Sleep(10 * time.Second)
	}

	return nil, fmt.Errorf(fmt.Sprintf("timeout waiting to app (%s) deployment", appID))
}
```

With all that said, I took allergy medicine a bit ago and feel really messed up now haha, so please keep that in mind when considering this feedback. I'll have a closer look on Monday when I'm feeling normal 😄

andrewsomething:
> With all that said, I took allergy medicine a bit ago and feel really messed up now haha, so please keep that in mind when considering this feedback. 

@bentranter You're definitely correct. :smile:  We discussed this right after it was pushed up. Should have updated here!

danaelhe:
> > With all that said, I took allergy medicine a bit ago and feel really messed up now haha, so please keep that in mind when considering this feedback.
> 
> @bentranter You're definitely correct. 😄 We discussed this right after it was pushed up. Should have updated here!

that's a good point! I agree with your suggested approach, will apply it.

danaelhe:
I'm never sure if I should commit the changes made to the vendor files. Is this kosher?

go.mod
```diff
@@ -17,27 +16,28 @@ require (
...
 	github.com/opencontainers/image-spec v1.0.1 // indirect
 	github.com/opencontainers/runc v1.0.0-rc90 // indirect
 	github.com/pelletier/go-toml v1.6.0 // indirect
+	github.com/russross/blackfriday/v2 v2.1.0 // indirect
 	github.com/sclevine/spec v1.3.0
 	github.com/shiena/ansicolor v0.0.0-20151119151921-a422bbe96644
-	github.com/sirupsen/logrus v1.7.0 // indirect
-	github.com/spf13/cobra v0.0.3
+	github.com/sirupsen/logrus v1.8.1 // indirect
+	github.com/spf13/cobra v1.1.3
```
andrewsomething, line 32: Since Cobra is such a core piece of our app, I'd be reluctant to do a major upgrade with out a closer review of what has changed and how that impacts us. I think that would be better done in a PR by itself to make it easier to revert if needed.

andrewsomething:
> I'm never sure if I should commit the changes made to the vendor files. Is this kosher?

Committing the changes to the vendor files is correct if you are intending to update the dependencies. Unless that's needed for this PR, I'd rather not.

danaelhe:
> > I'm never sure if I should commit the changes made to the vendor files. Is this kosher?
> 
> Committing the changes to the vendor files is correct if you are intending to update the dependencies. Unless that's needed for this PR, I'd rather not.

I agree, I'll keep it out of scope for this PR and revert the changes.

commands/apps_test.go
```diff
@@ -189,109 +189,6 @@ func TestRunAppsDelete(t *testing.T) {
...
-				SourceCommitHash: "commit",
-			}},
-			Cause: "Manual",
-			Phase: godo.DeploymentPhase_Active,
-			Progress: &godo.DeploymentProgress{
-				PendingSteps: 1,
-				RunningSteps: 0,
-				SuccessSteps: 0,
-				ErrorSteps:   0,
-				TotalSteps:   1,
```
andrewsomething, line 272: Can't add a code suggestion to deleted code apparently, but this should still pass. Just need to update the response to have:  `SuccessSteps: 1,` due to the change in how we determine a successfully finished deployment,

commands/apps_test.go
```diff
@@ -189,109 +189,6 @@ func TestRunAppsDelete(t *testing.T) {
...
-					Status:    "pending",
-					StartedAt: time.Now(),
-				}},
-			},
-			CreatedAt: time.Now(),
-			UpdatedAt: time.Now(),
-		}
-
-		tm.apps.EXPECT().CreateDeployment(appID, false).Times(1).Return(deployment, nil)
-		tm.apps.EXPECT().GetDeployment(appID, deployment.ID).Times(1).Return(activeDeployment, nil)
```
andrewsomething, line 285: This would now be expected to be called twice:

```
tm.apps.EXPECT().GetDeployment(appID, deployment.ID).Times(2).Return(activeDeployment, nil)
```

commands/apps.go
```diff
@@ -331,59 +373,57 @@ func RunAppsCreateDeployment(c *CmdConfig) error {
...
-			failCount = 0
-		} else {
-			// Allow for transient API failures
-			failCount++
-			if failCount >= maxAPIFailures {
-				return nil, err
+		if deploymentID == "" {
+			app, err := apps.Get(appID)
+			if err != nil {
+				return fmt.Errorf(("Error trying to read app deployment state: " + err.Error()))
```
andrewsomething, line 400:
```suggestion
				return err
```

We could just return the `err` directly here and handle the messaging further up the stack. Looking at how the error is checked in `RunAppsCreateDeployment`, the output becomes a bit repetitive: 

     App deployment couldn't enter `running` state: Error trying to read app deployment state: <actual error>

commands/apps.go
```diff
@@ -331,59 +373,57 @@ func RunAppsCreateDeployment(c *CmdConfig) error {
...
-			continue
-		}
+			if app.InProgressDeployment != nil {
+				deploymentID = app.InProgressDeployment.ID
+			}
+		} else {
+			deployment, err := apps.GetDeployment(appID, deploymentID)
+			if err != nil {
+				ticker.Stop()
+				return fmt.Errorf(fmt.Sprintf("Error trying to read app deployment state: " + err.Error()))
```
andrewsomething, line 410:
```suggestion
				return err
```

Same thing here. Also one thing to note is that you don't need to use `fmt.Sprintf` with `fmt.Errorf`, you could just use:

    return fmt.Errorf("Error trying to read app deployment state: %s", err.Error())

andrewsomething (changes requested):

commands/apps_test.go
```diff
@@ -189,109 +189,6 @@ func TestRunAppsDelete(t *testing.T) {
 	})
 }
 
-func TestRunAppsCreateDeployment(t *testing.T) {
```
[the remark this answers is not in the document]
danaelhe: ok cool, i'll bring it back the in my next commit along with the rest of the changes.

commands/apps.go
danaelhe: thanks for the tip!

go.mod
danaelhe: i agree, gonna keep this out for now.

danaelhe:
going to open another PR without the 400+ dependencies

PR #994, opened 2021-06-15
adding the --wait flag to the apps create command

danaelhe:
(reopened PR to disregard 400+ vendor files committed)

per github issue #970 - adding a wait flag for doctl apps create command. The integration test is taking 30 seconds to complete. This causes a timeout error- if timeout is removed it works. Looking for a second pair of eyes to help determine why the test takes so long.

applies changes suggested from previous pr such as:
- replaced ticker with for loop (thanks @bentranter)
- clearer error messages

### DIFF
--- a/commands/apps.go
+++ b/commands/apps.go
@@ -55,6 +55,8 @@ func Apps() *Command {
 		displayerType(&displayers.Apps{}),
 	)
 	AddStringFlag(create, doctl.ArgAppSpec, "", "", `Path to an app spec in JSON or YAML format. Set to "-" to read from stdin.`, requiredOpt())
+	AddBoolFlag(create, doctl.ArgCommandWait, "", false,
+		"Boolean that specifies whether to wait for an app to complete before returning control to the terminal")
 
 	CmdBuilder(
 		cmd,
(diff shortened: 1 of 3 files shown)


commands/apps.go
```diff
@@ -331,59 +373,48 @@ func RunAppsCreateDeployment(c *CmdConfig) error {
 	return c.Display(displayers.Deployments{deployment})
 }
 
-// waitForAppDeploymentRunning waits for a app deployment to be running.
-func waitForAppDeploymentRunning(apps do.AppsService, appID string, deploymentID string) (*godo.Deployment, error) {
-	failCount := 0
+func waitForActiveDeployment(apps do.AppsService, appID string, deploymentID string) error {
+	const maxAttempts = 10 * 6 * 30
```
andrewsomething, line 377: Think we might want something a bit shorter here. 1800 attempts at 10 seconds each is 5 hours. Maybe something like 30 minutes?

andrewsomething (commented):
Looks great, but I did have one question/comment inline about the length of the timeout.

commands/apps.go
danaelhe: oh 30 minutes is more reasonable hahaha. Will do.

commands/apps.go
```diff
@@ -331,59 +373,48 @@ func RunAppsCreateDeployment(c *CmdConfig) error {
 	return c.Display(displayers.Deployments{deployment})
 }
 
-// waitForAppDeploymentRunning waits for a app deployment to be running.
-func waitForAppDeploymentRunning(apps do.AppsService, appID string, deploymentID string) (*godo.Deployment, error) {
-	failCount := 0
+func waitForActiveDeployment(apps do.AppsService, appID string, deploymentID string) error {
+	const maxAttempts = 1800
```
andrewsomething, line 377: In the version using a ticker, 1800 was the max number of seconds (it breaks on `i * interval > timeout`).  Moving to a loop, this is now the number of times that we go through the loop.  With the 10 second sleep at the end of the loop, `maxAttempts` should be `1800 / 10` for a half hour timeout.

danaelhe: Oh! I get it. Thanks for explaining.

andrewsomething (approved):
:+1:  LGTM!

PR #998, opened 2021-06-28
Fix auth key mangling when switching contexts

bentranter:
"Fixes" a bizarre issue where running `doctl auth switch` mangles all auth contexts that contain a `.` character.

Prior to this change, when the command was run, it appears that Viper decides that anything containing a `.` indicates that the value should be split at the `.`, and then anything after it should become the key of a nested entry in a map in the config YAML file.

I have no idea why this fix works, but it does. I'm still looking into the root cause, but I figure that we might as well get this change out.

### DIFF
--- a/commands/auth.go
+++ b/commands/auth.go
@@ -217,6 +217,13 @@ func RunAuthSwitch(c *CmdConfig) error {
 		context = viper.GetString("context")
 	}
 
+	// The two lines below aren't required for doctl specific functionality,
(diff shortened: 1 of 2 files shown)


andrewsomething (commented):
Any way we can test this? Since the behavior is so strange, it's definitely something we want to know has changed if we say upgrade Viper.

andrewsomething (approved):
:+1:

Issue #1001, opened 2021-07-05
1.61.0 release tarball changed?


Jehops:
FreeBSD package maintainer here.  Our package building system does a checksum on the release tarballs and we are now seeing a mismatch.  It doesn't look like the release was re-rolled.  Do you know anything about the change?

I'm told that these GitHub generated tarballs are not guaranteed to be static, although this is a first for me in years of maintaining dozens of GitHub hosted packages.
 
```=> Attempting to fetch https://codeload.github.com/digitalocean/doctl/tar.gz/v1.61.0?dummy=/digitalocean-doctl-v1.61.0_GH0.tar.gz
fetch: https://codeload.github.com/digitalocean/doctl/tar.gz/v1.61.0?dummy=/digitalocean-doctl-v1.61.0_GH0.tar.gz: size unknown
fetch: https://codeload.github.com/digitalocean/doctl/tar.gz/v1.61.0?dummy=/digitalocean-doctl-v1.61.0_GH0.tar.gz: size of remote file is not known
digitalocean-doctl-v1.61.0_GH0.tar.gz                 5078 kB 5605 kBps    01s
=> Fetched file size mismatch (expected 5200371, actual 5200375)
```

andrewsomething:
Hi @Jehops. Thanks for maintaining the FreeBSD package! This does seem odd.

We build and upload the binary tarballs using goreleaser and we include the SHA256 checksums for them as part of the release.. The source tarballs are generated by GitHub automatically. We don't keep a record of what the checksum for those are at release.

What was the original checksum you recorded?

Jehops:
Hi @andrewsomething.  My pleasure. 

That the GitHub automatically generated tarballs can change is odd and unfortunate, but [described elsewhere](https://github.com/keybase/client/issues/10800).

The original SHA256 was [5334fb1d2c169064d2a7584186bc4b3c48952e8af0a139447e890e204531356e](https://cgit.freebsd.org/ports/tree/sysutils/doctl/distinfo).  It's now 571ae4f408a267d9824b56b5ed43604c867be951c09d8fd9b2b0db7cb4635078.

andrewsomething:
Looks like it should be pretty trivial for us to start including our own source archive as part of the release process:

https://goreleaser.com/customization/source/

That should at least prevent this in the future.

PR #1002, opened 2021-07-06
goreleaser: Release stable source tarball (fixes: #1001).

andrewsomething:
When GitHub generates source tarballs on the fly, there is no guarantee of byte-for-byte equivalence. This can cause issues for distribution packagers and others that wish to verify their downloads. 

This can be tested locally using: `goreleaser --snapshot ` The resulting `dist/` directory should include a tarball with the project source and the resulting SHA256 file should have the checksum of that file.


scotchneat (approved):
lgtm

PR #1004, opened 2021-07-07
apps: Add --tail flag to support tailing live application logs

ZachEddy:
This PR introduces a new flag on `apps logs` so users can specify the number of live log lines they want to fetch from an application. Currently, running `apps logs` will return all log lines.

It behaves as `docker logs` and `kubectl logs` would:

```sh
# fetch the last 10 logs
doctl apps logs <app-id> --tail 10

# fetch all the logs
doctl apps logs <app-id>
```

Note: I ran `make mocks` and most of the mock files changed, so I only committed changes to `mocks/AppsService.go` to keep the PR small. I tried using several different versions of `mockgen` but they all resulted in a large diff. Let me know if there's a specific version of `mockgen` I should use, or if I should commit all the mock changes in this PR.

### DIFF
--- a/args.go
+++ b/args.go
@@ -44,6 +44,8 @@ const (
 	ArgAppDeployment = "deployment"
 	// ArgAppLogFollow follow logs.
 	ArgAppLogFollow = "follow"
+	// ArgAppLogTail tail logs.
+	ArgAppLogTail = "tail"
 	// ArgAppForceRebuild forces a deployment rebuild
 	ArgAppForceRebuild = "force-rebuild"
 	// ArgClusterName is a cluster name argument.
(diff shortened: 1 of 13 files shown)


scotchneat (approved):
Looks good to me.
You did the right thing with the mocks @ZachEddy. Looks like we'll have to get those other files caught up (separately).

andrewsomething:
Good stuff! One question, should `--follow` and `--tail` be mutually exclusive options?

ZachEddy:
> Good stuff! One question, should `--follow` and `--tail` be mutually exclusive options?

IMO they should not be mutually exclusive because behavior and functionality-wise they don't conflict. Using them together works as one might expect: grab some specified number of logs with `--tail` and stream new logs with the `--follow` flag.

For what it's worth, both `kubectl` and `docker` let users specify both flags.

andrewsomething:
> Using them together works as one might expect: grab some specified number of logs with --tail and stream new logs with the --follow flag.

:ok_hand: Perfect, thanks!

PR #941, opened 2021-01-20
travis: Log into Docker before pulling images.

andrewsomething:
Docker is rate-limiting pulls. Log into the sammytheshark user before building snaps so that we are not an anonymous user being rate-limited by the shared Travis IP address. 

https://blog.travis-ci.com/docker-rate-limits
https://www.docker.com/increase-rate-limits

PR #1005, opened 2021-07-12
Adds release workflow

scotchneat:
This is to support releases using GH Action workflow (to migrate from TravisCI)

* migrates changelog generator to use `goreleaser`
* removes unused make target
* adds workflow to build and publish a snap release candidate on push to main
* adds job to the release.yml workflow to build a publish a snap stable release on push of a tag


.github/workflows/release.yml
```diff
@@ -0,0 +1,58 @@
...
+    runs-on: 'ubuntu-latest'
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Buld snap
+        id: build
+        run: |
+          make _build_snap && \
+          find doctl_v*.snap -print -exec echo ::set-output name=snap::{} \;
```
scotchneat, line 49: I'm sure there's a more efficient way to get the file name and am open to suggestions

ChiefMateStarbuck (dismissed):
How would i test this?

ChiefMateStarbuck:
should we make 1.16 required?

.github/workflows/release.yml
andrewsomething: I think that works for now. If we can dump the customer build image, it looks like the `snapcore/action-build` action handles this nicely.

.github/workflows/release.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      - uses: snapcore/action-publish@v1
```
andrewsomething, line 51: :ok_hand: Nice, hadn't seen that there was an official action for this.

.github/workflows/release.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Install Go
+        uses: actions/setup-go@v2
+      - name: Restore cache
+        uses: actions/cache@v2
+        with:
+          path: ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+      - name: Run GoReleaser
```
andrewsomething, line 29:
```suggestion
      - name: Login to Docker Hub
        uses: docker/login-action@v1
        with:
          username: sammytheshark
          password: ${{ secrets.DOCKER_PASSWORD }}

      - name: Run GoReleaser
```

Some time before the goreleaser step, we need to log into Docker Hub

.github/workflows/release.yml
```diff
@@ -0,0 +1,58 @@
...
+          make _build_snap && \
+          find doctl_v*.snap -print -exec echo ::set-output name=snap::{} \;
+
+      - uses: snapcore/action-publish@v1
+        with:
+          store_login: ${{ secrets.SNAP_TOKEN }}
+          snap: ${{ steps.build.outputs.snap }}
+          release: stable
+        env:
+          DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
```
andrewsomething, line 57: I don't think this is needed here.

scotchneat:
> How would i test this?

@ChiefMateStarbuck other than using a 3rd party tool like [act](https://github.com/nektos/act) to test locally, the respective event would need to be triggered - push to master/merge a pr, or tag a release

scotchneat:
> should we make 1.16 required?

@ChiefMateStarbuck I think that'd be a good, separate change

.github/workflows/snapcraft-candidate.yml
```diff
@@ -0,0 +1,29 @@
...
+    branches: [main]
+
+jobs:
+  build-and-publish:
+    name: 'Build and publish'
+    runs-on: 'ubuntu-latest'
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
```
andrewsomething, line 15: Even though we are not publishing a Docker image in this one, we still need to do the Docker login dance. `make _build_snap` pull the build container and Docker is now rate limiting pulls. Since Travis shares IPs across builds we were hitting this even though we never come close to that ourselves.

See: https://github.com/digitalocean/doctl/pull/941

Looks like this is a concern for GitHub Actions as well:

https://docs.docker.com/docker-hub/download-rate-limit/#github-actions

.github/workflows/snapcraft-candidate.yml
```diff
@@ -0,0 +1,29 @@
...
+          make _build_snap && \
+          find doctl_v*.snap -print -exec echo ::set-output name=snap::{} \;
+
+      - uses: snapcore/action-publish@v1
+        with:
+          store_login: ${{ secrets.SNAP_TOKEN }}
+          snap: ${{ steps.build.outputs.snap }}
+          release: candidate
+        env:
+          DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
```
andrewsomething, line 28: Don't think this is needed in the publish action, but we do need to log into Docker Hub first.

andrewsomething (changes requested):

.github/workflows/release.yml
scotchneat: I noticed I left this too

ChiefMateStarbuck:
> > How would i test this?
> 
> @ChiefMateStarbuck other than using a 3rd party tool like [act](https://github.com/nektos/act) to test locally, the respective event would need to be triggered - push to master/merge a pr, or tag a release

act doesnt seem to work so smoothly out the box. eh, i trust you. good work.  👍🏾

scotchneat:
I'm not sure what's going on with my local git repos lately. I did something to jack up this change so I'll open a new PR

PR #1006, opened 2021-07-19
monitoring: add CRUD commands for alert policies

bsnyder788:


commands/monitoring.go
```diff
@@ -0,0 +1,350 @@
...
+	"github.com/digitalocean/godo"
+	"github.com/spf13/cobra"
+)
+
+// Monitoring creates the monitoring commands hierarchy.
+func Monitoring() *Command {
+	cmd := &Command{
+		Command: &cobra.Command{
+			Use:   "monitoring",
+			Short: "Display commands to manage monitoring",
```
andrewsomething, line 33: Do we want to label these commands as beta for now? 

```suggestion
			Short: "[Beta] Display commands to manage monitoring",
```

Adding `Hidden:  true,` to the top level command will hide it from the main list of commands if you want to further limit its discoverability for now.

commands/monitoring.go
```diff
@@ -0,0 +1,350 @@
...
+	AddStringFlag(cmdAlertPolicyCreate, doctl.ArgAlertPolicyDescription, "", "", "A description of the alert policy.")
+	AddStringFlag(cmdAlertPolicyCreate, doctl.ArgAlertPolicyType, "", "", "The type of alert policy.")
+	AddStringFlag(cmdAlertPolicyCreate, doctl.ArgAlertPolicyCompare, "", "", "The comparator of the alert policy. Either `GreaterThan` or `LessThan`")
+	AddStringFlag(cmdAlertPolicyCreate, doctl.ArgAlertPolicyWindow, "", "5m", "The window to apply the alert policy conditions against.")
+	AddIntFlag(cmdAlertPolicyCreate, doctl.ArgAlertPolicyValue, "", 0, "The value of the alert policy to compare against.")
+	AddBoolFlag(cmdAlertPolicyCreate, doctl.ArgAlertPolicyEnabled, "", true, "Whether the alert policy is enabled.")
+	AddStringSliceFlag(cmdAlertPolicyCreate, doctl.ArgAlertPolicyEmails, "", nil, "Emails to send alerts to.")
+	AddStringSliceFlag(cmdAlertPolicyCreate, doctl.ArgAlertPolicyTags, "", nil, "Tags to apply the alert against.")
+	AddStringSliceFlag(cmdAlertPolicyCreate, doctl.ArgAlertPolicyEntities, "", nil, "Entities to apply the alert against.")
```
andrewsomething, line 64:
```suggestion
	AddStringSliceFlag(cmdAlertPolicyCreate, doctl.ArgAlertPolicyEntities, "", nil, "Entities to apply the alert against (e.g. a Droplet's ID).")
```

Think an example might help here since "entities" isn't a very common term in our services.

commands/monitoring.go
```diff
@@ -0,0 +1,350 @@
...
+	if err != nil {
+		return err
+	}
+
+	tags, err := c.Doit.GetStringSlice(c.NS, doctl.ArgAlertPolicyTags)
+	if err != nil {
+		return err
+	}
+
+	enabled, err := c.Doit.GetBoolPtr(c.NS, doctl.ArgAlertPolicyEnabled)
```
andrewsomething, line 125:
```suggestion
	enabled, err := c.Doit.GetBool(c.NS, doctl.ArgAlertPolicyEnabled)
```

Looks like `GetBoolPtr` ignores defaults and and excluding the enabled flag results in `nil`. Not sure if that's actually the intended behavior. 

https://github.com/digitalocean/doctl/blob/c7bf8d427e73e395f59f743cee05f6747392b677/doit.go#L292-L298

commands/monitoring.go
```diff
@@ -0,0 +1,350 @@
...
+	if len(slackURLs) != len(slackChannels) {
+		return errors.New("must provide the same number of slack channels as slack URLs")
+	}
+
+	slacks := make([]godo.SlackDetails, len(slackChannels))
+	for i, channel := range slackChannels {
+		slacks[i] = godo.SlackDetails{Channel: channel, URL: slackURLs[i]}
+	}
+
+	apcr := &godo.AlertPolicyCreateRequest{
```
andrewsomething, line 164: Is there any other input validation we should be doing before making the request?

commands/monitoring.go
```diff
@@ -0,0 +1,350 @@
...
+		Compare:     compare,
+		Value:       float32(value),
+		Window:      window,
+		Entities:    entities,
+		Tags:        tags,
+		Alerts: godo.Alerts{
+			Slack: slacks,
+			Email: emails,
+		},
+		Enabled: enabled,
```
andrewsomething, line 176:
```suggestion
		Enabled: &enabled,
```

andrewsomething (commented):
This is looking great! A couple very minor questions in line. It would be great to get some integration tests as well. See:

https://github.com/digitalocean/doctl/tree/main/integration
https://github.com/digitalocean/doctl/blob/main/CONTRIBUTING.md#integration-tests

I did have trouble testing the create command, but that might be expected right now?

```
./builds/doctl monitoring alerts create --compare GreaterThan --value 50 --type v1/insights/droplet/cpu --description "Test alert" --enabled  --tags personal
Enabled: trueError: POST https://api.digitalocean.com/v2/monitoring/alerts: 500 (request "4fec9302-c09c-4da5-ab86-7dea6a8ec900") Server Error
```

commands/monitoring.go
bsnyder788: I think we can let it be in terms of hidden, since we should be entering open beta period next week. I will add the [Beta] marker though!

commands/monitoring.go
bsnyder788: good call out!

bsnyder788:
> This is looking great! A couple very minor bits in line. It would be great to get some integration tests as well. See:
> 
> https://github.com/digitalocean/doctl/tree/main/integration
> https://github.com/digitalocean/doctl/blob/main/CONTRIBUTING.md#integration-tests
> 
> I did have trouble testing the create command, but that might be expected right now?
> 
> ```
> ./builds/doctl monitoring alerts create --compare GreaterThan --value 50 --type v1/insights/droplet/cpu --description "Test alert" --enabled  --tags personal
> Error: POST https://api.digitalocean.com/v2/monitoring/alerts: 500 (request "4fec9302-c09c-4da5-ab86-7dea6a8ec900") Server Error
> ```

It looks like you didn't supply any emails or slacks to send the alert to in this one. I will add some better validation so that we can stop it from sending requests without all the required params being passed.

commands/monitoring.go
bsnyder788: I went ahead and added validation on the actual window supplied, the alert policy type, and that there are emails and/or slacks supplied!

bsnyder788: I did that for both create and update.

bsnyder788:
> This is looking great! A couple very minor bits in line. It would be great to get some integration tests as well. See:
> 
> https://github.com/digitalocean/doctl/tree/main/integration
> https://github.com/digitalocean/doctl/blob/main/CONTRIBUTING.md#integration-tests
> 
> I did have trouble testing the create command, but that might be expected right now?
> 
> ```
> ./builds/doctl monitoring alerts create --compare GreaterThan --value 50 --type v1/insights/droplet/cpu --description "Test alert" --enabled  --tags personal
> Error: POST https://api.digitalocean.com/v2/monitoring/alerts: 500 (request "4fec9302-c09c-4da5-ab86-7dea6a8ec900") Server Error
> ```

I went ahead and added more integration tests.

commands/monitoring.go
bsnyder788: I added some more input validation for both the update and create commands!

commands/monitoring.go
bsnyder788: good call, I went ahead and clarified it with an example!

PR #1015, opened 2021-08-09
workflow/github-release: Build with Go 1.16.x

andrewsomething:
This should build Darwin ARM64 releases again. 

https://github.com/digitalocean/doctl/runs/3284170730?check_suite_focus=true#step:6:29

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,6 +26,9 @@ jobs:
 
       - name: Install Go
         uses: actions/setup-go@v2
+        with:
+          go-version: 1.16.x
+
       - name: Restore cache
         uses: actions/cache@v2
         with:


ChiefMateStarbuck (approved):

bentranter (approved):

ChiefMateStarbuck:
The CI Test Suite is requiring 1.15 . Should we update this to 1.16 and remove 1.15?

andrewsomething:
@ChiefMateStarbuck CI uses both 1.15.x and 1.16.x: https://github.com/digitalocean/doctl/blob/main/.github/workflows/ci.yml#L13

Think there is value in ensuring we build with both for users building their own in environments without the latest and greatest. That matches the upstream policy around supported releases, 1.15.x is supported until 1.17.x is released:  https://golang.org/doc/devel/release#policy

Issue #1020, opened 2021-08-26
domain record update unexpectedly changes record TTL


rcj4747:
**Describe the Issue:**

I have an existing domain record where I want to update one field, RecordData, but if I don't look up the existing TTL value and specify it during an update it will be changed to [1800](https://github.com/digitalocean/doctl/blob/84fb230735f6e6d0f7636aae8d4cbeb74d931ac7/commands/domains.go#L87).

I would like to run `doctl compute domain records update ${ZONE} --record-id ${REC_ID} --record-data ${NEW_IP}` and have the TTL remain unchanged.

The API documentation for [update_domain_record](https://docs.digitalocean.com/reference/api/api-reference/#operation/update_domain_record) does not indicate that the field is nullable :-( but a test with a value of 0 shows that the existing TTL is preserved.

**Environment:**

 - doctl version: doctl version 1.64.0-release, Git commit hash: 84fb2307
 - OS: Ubuntu 21.04
 - Installation method: Snap package version v1.63.1+git1.84fb2307

**Additional Details:**

Recreate steps:

```
ZONE=do.examplezone.com
RECORD=foo
DATA=192.168.1.1
TTL=600

# Create a record with a custom TTL
$ doctl compute domain records create $ZONE --record-type A --record-ttl $TTL --record-name $RECORD --record-data $DATA
ID           Type    Name    Data           Priority    Port    TTL    Weight
169333525    A       foo     192.168.1.1    0           0       600    0

# Updating a records without specifying TTL changes it to the default of 1800
$ doctl compute domain records update $ZONE --record-id 169333525 --record-data 192.168.1.2
ID           Type    Name    Data           Priority    Port    TTL     Weight
169333525    A       foo     192.168.1.2    0           0       1800    0

# Updating the record requires specification of the desired TTL to maintain its value
$ doctl compute domain records update $ZONE --record-id 169333525 --record-data 192.168.1.2 --record-ttl 600
ID           Type    Name    Data           Priority    Port    TTL    Weight
169333525    A       foo     192.168.1.2    0           0       600    0

# Patch the source for doctl changing the default TTL on updates from 1800 to 0
$ git diff
diff --git a/commands/domains.go b/commands/domains.go
index d5ac511b..24936095 100644
--- a/commands/domains.go
+++ b/commands/domains.go
@@ -84,7 +84,7 @@ func Domain() *Command {
[...]

PR #1021, opened 2021-08-26
Update domain record without changing TTL (#1020)

rcj4747:
Given an existing domain record with an non-default TTL, a user can not
update a field on the record without TTL changing to 1800 unless the
user remembers to specify the desired TTL on each update.  This patch
changes the default value of the TTL argument from 1800 to 0 for domain
record updates which preserves the current TTL of the record.


danaelhe:
Solved in #1027

PR #1027, opened 2021-08-31
Domain records update ttl put to patch per #1020

danaelhe:
Note- in order to completely solve [this issue](https://github.com/digitalocean/doctl/issues/1020) #1020 you must also edit your [Configuring Default Values](https://github.com/digitalocean/doctl#configuring-default-values) and change the default values for records from 1800 to 0


commands/domains.go
```diff
@@ -371,6 +371,7 @@ func RunRecordUpdate(c *CmdConfig) error {
 	if err != nil {
 		return err
 	}
+	fmt.Println(rTTL)
```
ChiefMateStarbuck, line 374: was this line for testing purposes or is this the line that prints proper output?

danaelhe: for testing- it's been removed in the next commit- thanks!

ChiefMateStarbuck (approved):

PR #1033, opened 2021-09-23
load-balancers: Add size-unit to commands.

wez470:
Adds new sizing field to load balancers create and update commands.

I noted that the comma separated value fields of health-checks and forwarding-rules didn't work with spaces so I also updated the descriptions around those.

Also, making size no longer required seemed to cause some issues for me locally. The cli was still complaining that it was required. Turns out my cli config file stores that + the default value and the cli was loading those even though the code for the command itself no longer required the field. Not sure if there's a good way around this. I had to manually fix up my config file for it to work properly.


commands/load_balancers.go
```diff
@@ -74,8 +74,10 @@ With the load-balancer command, you can list, create, or delete load balancers,
 		"The load balancer's name", requiredOpt())
 	AddStringFlag(cmdRecordCreate, doctl.ArgRegionSlug, "", "",
 		"The load balancer's region, e.g.: `nyc1`", requiredOpt())
-	AddStringFlag(cmdRecordCreate, doctl.ArgSizeSlug, "", "lb-small",
-		"The load balancer's size, e.g.: `lb-small`", requiredOpt())
+	AddStringFlag(cmdRecordCreate, doctl.ArgSizeSlug, "", "",
+		fmt.Sprintf("The load balancer's size, e.g.: `lb-small`. Only one of %s and %s should be used", doctl.ArgSizeSlug, doctl.ArgSizeUnit))
+	AddIntFlag(cmdRecordCreate, doctl.ArgSizeUnit, "", 0,
```
andrewsomething, line 79: Is there a good default value we can set?  Is there an equivalent to `lb-small` in the new format?

andrewsomething (approved):
:+1:  Looks good! I did have one question inline.

commands/load_balancers.go
wez470: `1` is equivalent. I've opted to set both defaults to zero values for now. This is b/c size slug is still expected to be used for now and only one of these fields is allowed to be specified. In the case where none of the values are set, `lb-small`/`1` is the assumed size so the behaviour should stay the same as today.

andrewsomething: > In the case where none of the values are set, lb-small/1 is the assumed size so the behaviour should stay the same as today.

:ok_hand:  Perfect. Just wanted to make sure we wouldn't be breaking any scripts that might rely on there being a default.

PR #1035, opened 2021-09-29
k8s: add support for ha flag

sunny-b:
This PR adds support for the `ha` flag for the Kubernetes Create and Update commands. It also makes it so the `HA Control Plane` field is outputted in the JSON response.

cc: @adamwg


adamwg (dismissed):
lgtm

bentranter (commented):
Looks good overall @sunny-b, one request – would be able to add or extend an integration test for this? Adding the flag to [this case](https://github.com/digitalocean/doctl/blob/main/integration/kubernetes_clusters_create_test.go#L130) is probably enough. Thanks!

sunny-b:
@bentranter added `ha` to some of the tests and the tests are passing now

bentranter (approved):
Looks great! Thanks @sunny-b

PR #1036, opened 2021-10-01
Add Go 1.17 to the tests and release

bentranter:
Adds Go 1.17 to our test matrix, and uses it to build the release.


bentranter:
Oh no lol I didn't mean to update the `/vendor` directory with the `gofmt` stuff 😅 I'll fix it

.github/workflows/ci.yml
```diff
@@ -10,7 +10,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [ 1.15.x, 1.16.x]
+        go-version: [ 1.15.x, 1.16.x, 1.17.x ]
```
andrewsomething, line 13: How do we feel about dropping 1.15? Feel like we should try to align to the upstream policy when possible: 

https://golang.org/doc/devel/release#policy

Double checked and the Snap is already on 1.16:

https://github.com/digitalocean/doctl/blob/283ec460997e86b99a09a2bd4ba223636eece782/dockerfiles/Dockerfile.snap#L49

andrewsomething (approved):
:+1: Looks good!

.github/workflows/ci.yml
andrewsomething: On the other hand, it doesn't hurt to know if it passes or not... Maybe just mark as not required?

ChiefMateStarbuck (approved):

Issue #989, opened 2021-06-03
doctl apps create support for update existing apps


lbazetto:
**What is the problem this feature would solve? Please describe.**

I have a few CI/CD pipelines to deploy different applications using the `doctl apps`command.
But at the moment, I have a step just to verify if the app exists and then, choose between `doctl apps create` for new apps or `doctl apps update` for existing ones.

**Describe the solution you'd like**
A new flag in the `doctl apps create --allow-update` or a new command, such as: `doctl apps create-update`

thanks :)

Shanduur:
Hi. I would suggest to create flag (`-u | --update`) that will automatically existing app. If no one is working on it, I will take that.

danaelhe:
Hi there,

Thanks for the write up. I was wondering if you'd be able to accomplish what you'd want with this linux work around:

`doctl apps update b2bb3-b332-b222 --spec-file data.yaml || doctl apps create --spec-file data.yaml`

that way it'd remove the step to verify if the app exists before. This workaround could remove the need to update doctl- unless there's a use case where the linux work around would not suffice?

Shanduur:
@danaelhe I think, @lbazetto was referring to some CI/CD pipelines. Performing multiple commands in single action is often considered as bad practice, especially if those command chains are performing alternative actions.

However I think that approach of branched pipeline (`GET id; if not then CREATE; else UPDATE;`) might be a better idea in some scenarios.

Nevertheless, we write CLI tools to simplify communication with API. He could also use `curl` instead of `doctl`, as those are performing the same purpose here - sending request to API. I think that simplification might be beneficial in the long run to more people.

danaelhe:
I see. I've left a comment on the PR continuing this conversation at #1038

PR #1038, opened 2021-10-04
feat(apps): added create flag to update subcommand

Shanduur:
Added '--create' flag to 'doctl apps update' command that will cause creation of app if the app that will be updated does not exist.
Resolves digitalocean/doctl#989

### DIFF
--- a/commands/apps.go
+++ b/commands/apps.go
@@ -97,6 +97,7 @@ Only basic information is included with the text output format. For complete app
 	AddStringFlag(update, doctl.ArgAppSpec, "", "", `Path to an app spec in JSON or YAML format. Set to "-" to read from stdin.`, requiredOpt())
 	AddBoolFlag(update, doctl.ArgCommandWait, "", false,
 		"Boolean that specifies whether to wait for an app to complete before returning control to the terminal")
+	AddBoolFlag(update, "create", "", false, "Boolean that specifies whether to create the app if not exists")
 
 	deleteApp := CmdBuilder(
 		cmd,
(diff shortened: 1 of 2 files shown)


commands/apps.go
```diff
@@ -307,19 +308,48 @@ func RunAppsUpdate(c *CmdConfig) error {
...
-	app, err := c.Apps().Update(id, &godo.AppUpdateRequest{Spec: appSpec})
+	create, err := c.Doit.GetBool(c.NS, "create")
 	if err != nil {
 		return err
 	}
 
+	exists := true
+	if create {
+		_, err := c.Apps().Get(id)
+		if strings.Contains(err.Error(), "400") {
```
Shanduur, line 319: I would appreciate a review of this part

commands/apps.go
```diff
@@ -307,19 +308,48 @@ func RunAppsUpdate(c *CmdConfig) error {
...
 	wait, err := c.Doit.GetBool(c.NS, doctl.ArgCommandWait)
 	if err != nil {
 		return err
 	}
 
 	if wait {
 		apps := c.Apps()
-		notice("App update is in progress, waiting for app to be running")
+
+		if exists {
```
Shanduur, line 347: Also, I am not sure about how notice should be handled in different scenarios. This is probably the best way.

integration/apps_test.go
```diff
@@ -514,6 +514,32 @@ var _ = suite("apps/update", func(t *testing.T, when spec.G, it spec.S) {
 			expect.Equal(expectedOutput, strings.TrimSpace(string(output)))
 		})
 	})
+	when("the create flag is passed", func() {
```
Shanduur, line 517: This should result in basic coverage of testing the flag behaviour but does not test the app creation path.

danaelhe:
Thank you so much for your contribution! We will review and leave some comments/feedback.

danaelhe:
After discussing with the team, we think adding an `--upsert` flag to the `doctl apps create` command would be a less surprising approach to addressing the original request. "Upsert" would work the same way it does in the context of database terms- create the app if it does not already exist, or update the app if it does. If the `--upsert` flag is passed, `doctl` could find the ID of the existing app with that name (App names must be unique) and issue an update. Currently, attempting to create an app with the same name as an existing one will fail with a 409 Conflict:
```
   doctl apps create --spec sample-golang-test.yaml
   Error: POST https://api.digitalocean.com/v2/apps: 409 (request "5ec4423d-dd47-478e-ba8f-4e2528014588") App name already exists
```

Makes sense? `doctl apps update` requires the ID of an app. Adding `--create` feels like it is overloading it and changing the semantics of the command. If you're up for adding this contribution (just from looking at this PR- you seem to understand the repo well!)- we'd really appreciate it!

Shanduur:
@danaelhe may I request a review?

PR #1039, opened 2021-10-04
doctl/load-balancers: add flag to toggle automatic DNS record creation for certs attached to load balancers

dikshant:
This brings in the change from godo: https://github.com/digitalocean/godo/pull/482 to doctl. We may need a new version release for godo before this can go in. This will probably also require some updates to doctl from DOKS side.


bentranter (approved):
Looks great!

PR #1062, opened 2021-11-10
Update vendored k8s.io libraries.

andrewsomething:
CVE-2021-25741 should not impact us, but it's a good excuse to update our vendored k8s.io libraries. Fixes: https://github.com/digitalocean/doctl/issues/1061


scotchneat (approved):
LGTM

Issue #818, opened 2020-06-02
Have doctl registry logout delete access token generated by doctl registry login


bockelie93:
**What is the problem this feature would solve? Please describe.**
In my pipeline, I get access to my docker registry with the command _doctl registry login_. This generates an API Token, which never expires (oldest is 19 days, so I assume it doesn't expire). For every pipeline run another Access Token is generated, which obviously is a problem over time when it doesn't expire.

I expected the command _doctl registry logout_, in addition to revoke docker's access, to also remove the access token generated. 

**Describe the solution you'd like**
Have _doctl registry logout_ also remove the access token generated by doctl registry login.

**Additional context**
https://github.com/digitalocean/doctl/issues/817

andrewsomething:
As mentioned in #817, doctl v1.44.0 now allows for setting `--expiry-seconds` on the `doctl registry login` command which allows for creating a short lived token that expires, but I'm still going to leave this ticket open.

As the credentials are based on an oAuth token, it might be a good idea to explicitly revoke them on `doctl registry logout` using the normal oAuth call:

https://developers.digitalocean.com/documentation/oauth/#revoke-token-flow

juburr:
I came here to request this as well. I discovered the `--expiry-seconds` workaround through the CLI help menus, but I think it's safter to explicitly remove the token on logout. I don't always know how long each CI job will take to push new images to the registry. Rather than heavily padding the expiration time, I'd rather just remove the token as soon as it's no longer needed by calling logout.

Through some experimentation, it seems like a `doctl registry logout` followed immediately by a `doctl registry login` will always generate a brand new token. With that in mind, I don't think there's a reason that this can't just be the default action on logout (as opposed to adding an additional flag like `--revoke-token` to the logout command).

PR #1068, opened 2021-12-07
registry: Revoke API token on log out.

andrewsomething:
Fixes: https://github.com/digitalocean/doctl/issues/818


bentranter (approved):
👍

PR #1069, opened 2021-12-07
registry: Add support for ListRepositoriesV2 and ListManifests commands.

CollinShoop:
CON-3907

Required bump to godo 1.73

### DIFF
--- a/commands/registry.go
+++ b/commands/registry.go
@@ -144,6 +144,21 @@ func Repository() *Command {
 		Writer, aliasOpt("ls"), displayerType(&displayers.Repository{}),
 	)
 
+	listRepositoriesV2Desc := `This command retrieves information about repositories in a registry, including:
+  - The repository name
+  - The latest manifest of the repository
+  - The latest manifest's latest tag, if any
+  - The number of tags in the repository
+  - The number of manifests in the repository
+`
+
+	CmdBuilder(
+		cmd,
+		RunListRepositoriesV2, "list-v2",
+		"List repositories for a container registry", listRepositoriesV2Desc,
+		Writer, aliasOpt("ls2"), displayerType(&displayers.Repository{}),
+	)
+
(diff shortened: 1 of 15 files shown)


commands/registry.go
```diff
@@ -144,6 +144,21 @@ func Repository() *Command {
 		Writer, aliasOpt("ls"), displayerType(&displayers.Repository{}),
 	)
 
+	listRepositoriesV2Desc := `This command retrieves information about repositories in a registry, including:
+  - The repository name
+  - The latest manifest of the repository
+  - The latest manifest's latest tag, if any
+  - The number of tags in the repository
+  - The number of manifests in the repository
+`
```
andrewsomething, line 153: I wonder if there is some additional info we could provide to users that would help them choose between `list` and `list-v2`?

commands/registry.go
```diff
@@ -144,6 +144,21 @@ func Repository() *Command {
 		Writer, aliasOpt("ls"), displayerType(&displayers.Repository{}),
```
andrewsomething, line 144:
```suggestion
		Writer, aliasOpt("ls"), displayerType(&displayers.Repository{}),
		hiddenCmd(),
```

FYI: If we want to push users towards the v2 command, it is possible to hid the current one from menus while leaving it available in order to not break any scripting.

andrewsomething (approved):
:+1:  This looks great. 

I did have a question inline about helping users understand the difference between `list` and `list-v2`

commands/registry.go
CollinShoop: I'm not sure. I was thinking just the fact of `v2` and seeing the different results would be self explanatory. I'm open to suggestions

andrewsomething: Since we've gone with hiding v1, it probably not very important now.

andrewsomething (approved):

Issue #1071, opened 2021-12-13
apps create and apps create-deployment do not return error codes


0xJem:
**Describe the Issue:**

When running `apps create` with the `--wait` flag, if there is an error in the deployment, the error code remains as 0. This makes it difficult to detect errors.

See the output below:
```
doctl apps create --upsert --spec app-spec.yml --wait --verbose
Notice: App already exists, updating
Notice: App creation is in progress, waiting for app to be running
....
Warning: App deployment couldn't enter `running` state: error deploying app (<ID> (deployment ID: <DID>):
[...]

PR #1072, opened 2021-12-14
ensuring error status 1 occurs when deployment fails

danaelhe:
Per #1071, this PR ensures a status error of 1 occurs when the app is created, but the deployment fails. It changes this:
```
danaelhertani@xxx doctl % go run cmd/doctl/main.go  apps create --upsert --spec app-spec.yml --wait --verbose                            
Notice: App already exists, updating
Notice: App creation is in progress, waiting for app to be running
Warning: App deployment couldn't enter `running` state: GET https://api.digitalocean.com/v2/apps/<APP-ID>/deployments/: 400 (request "<REQ-ID>") invalid uuid
ID                                      Spec Name             Default Ingress                                        Active Deployment ID                    In Progress Deployment ID    Created At                       Updated At
<APP-ID>   sample-golang-test    https://sample-golang-test.ondigitalocean.app    <DEPLOYMENT-ID>                                 2021-12-14 18:19:39 +0000 UTC    2021-12-14 18:36:45.711987968 +0000 UTC
danaelhertani@xxx doctl % echo $?
0
```

to this:
```
danaelhertani@xxx doctl % go run cmd/doctl/main.go  apps create --upsert --spec app-spec.yml --wait --verbose
Notice: App already exists, updating
Notice: App creation is in progress, waiting for app to be running
ID                                      Spec Name             Default Ingress                                        Active Deployment ID                    In Progress Deployment ID    Created At                       Updated At
<APP-ID>    sample-golang-test    https://sample-golang-test.ondigitalocean.app    <DEPLOYMENT-ID>                                 2021-12-14 18:19:39 +0000 UTC    2021-12-14 18:49:59.897000821 +0000 UTC
Error: 1 error occurred:
        * app deployment couldn't enter `running` state: GET https://api.digitalocean.com/v2/apps/<APP-ID>/deployments/: 400 (request "<REQ-ID>") invalid uuid


exit status 1
danaelhertani@xxx doctl % echo $?
1
```

Notice, the warning is turned to an error but we're able to keep the output of the app information. The error status also becomes 1.


commands/apps.go
```diff
@@ -273,13 +274,16 @@ func RunAppsCreate(c *CmdConfig) error {
...
 		err := waitForActiveDeployment(apps, app.ID, "")
 		if err != nil {
-			warn("App deployment couldn't enter `running` state: %v", err)
-			return c.Display(displayers.Apps{app})
+			errs = multierror.Append(errs, fmt.Errorf("app deployment couldn't enter `running` state: %v", err))
+			errs = multierror.Append(errs, c.Display(displayers.Apps{app}))
+			return errs
 		} else {
 			app, _ = c.Apps().Get(app.ID)
 		}
```
andrewsomething, line 289:
```suggestion
			errs = multierror.Append(errs, fmt.Errorf("app deployment couldn't enter `running` state: %v", err))
			if err := c.Display(displayers.Apps{app}); err != nil {
				errs = multierror.Append(errs, err)
			}

			return errs
		}

		app, _ = c.Apps().Get(app.ID)
```

Seems like multierror doesn't mind having a nil error appended, but probably a good practice to check anyways. As we're here, we could also get rid of the unnecessary else statement. (Applies to the other uses below too.)

andrewsomething (approved):
:+1:  Nice stuff! 

I did leave a suggestion inline as well.

app-spec.yml
```diff
@@ -0,0 +1,12 @@
+name: sample-golang-test
```
andrewsomething, line 1: Looks like this slipped back in.

danaelhe: good catch!

commands/apps.go
danaelhe: thanks for the suggestion!

andrewsomething (approved):
:+1:

Issue #856, opened 2020-09-23
Allow picking the context’s name in `doctl kubernetes cluster kubeconfig save`


vandrijevik:
**What is the problem this feature would solve? Please describe.**
Currently, `doctl kubernetes cluster kubeconfig save <NAME>` saves a `kubectl` context named `do-<REGION>-<NAME>`. This makes it difficult to use `doctl` in continuous integration runs, because the name it configures `kubectl` with depends on also knowing the region of the cluster.

**Describe the solution you'd like**
Adding an `--alias=name` option (I picked that name from `aws eks update-kubeconfig`), which saves authentication information under the specified context, would allow control over the context’s name. By making the name predictable, we can then use it elsewhere in CI pipelines, or other scripts locally.

vandrijevik:
Closing, since this hasn’t had any sort of response or reaction in a few months.

andrewsomething:
Hi @vandrijevik, 

Sorry for the lack of communication. I think this is a great suggestion and we have an internal ticket to implement it. Unfortunately, the team has other priorities at the moment. 

I'd be happy to review/accept a PR. If anyone is interested in working on this, here's a bit of context. We return the config here:

https://github.com/digitalocean/doctl/blob/a901a7278a7370ecc3cee919568c4653c1e1973f/commands/kubernetes.go#L1070

The config is a https://godoc.org/k8s.io/client-go/tools/clientcmd/api#Config

After checking for the flag, renaming the context might look something like:

	remoteKubeconfig.Contexts[alias] = remoteKubeconfig.Contexts[remoteKubeconfig.CurrentContext]
	delete(remoteKubeconfig.Contexts, remoteKubeconfig.CurrentContext)
	remoteKubeconfig.CurrentContext = alias

vandrijevik:
Cheers, thanks for letting me know @andrewsomething! I’m glad to read that this is on your radar :)

andrewsomething:
Fixed via https://github.com/digitalocean/doctl/pull/1075

PR #1075, opened 2021-12-20
APICLI-528: adding option to pick context name in 'kubeconfig save'

danaelhe:
per #856- adding the option to add an alias for a kubernetes context, as such:
`go run cmd/doctl/main.go kubernetes cluster kubeconfig save k8ds-test --alias=newalias_k8ds`

### DIFF
--- a/args.go
+++ b/args.go
@@ -98,6 +98,8 @@ const (
 	ArgKubernetesLabel = "label"
 	// ArgKubernetesTaint is a Kubernetes taint argument.
 	ArgKubernetesTaint = "taint"
+	// 	ArgKubernetesAlias is a Kubernetes alias argument that saves authentication information under the specified context.
+	ArgKubernetesAlias = "alias"
 	// ArgKubeConfigExpirySeconds indicates the length of time the token in a kubeconfig will be valid in seconds.
 	ArgKubeConfigExpirySeconds = "expiry-seconds"
 	// ArgImage is an image argument.
--- a/args.go
+++ b/args.go
@@ -98,7 +98,7 @@ const (
 	ArgKubernetesLabel = "label"
 	// ArgKubernetesTaint is a Kubernetes taint argument.
 	ArgKubernetesTaint = "taint"
-	// 	ArgKubernetesAlias is a Kubernetes alias argument that saves authentication information under the specified context.
+	// ArgKubernetesAlias is a Kubernetes alias argument that saves authentication information under the specified context.
 	ArgKubernetesAlias = "alias"
 	// ArgKubeConfigExpirySeconds indicates the length of time the token in a kubeconfig will be valid in seconds.
 	ArgKubeConfigExpirySeconds = "expiry-seconds"
--- a/commands/kubernetes.go
+++ b/commands/kubernetes.go
@@ -399,7 +399,7 @@ This command adds the credentials for the specified cluster to your local kubeco
 	AddBoolFlag(cmdSaveConfig, doctl.ArgSetCurrentContext, "", true, "Boolean indicating whether to set the current kubectl context to that of the new cluster")
 	AddIntFlag(cmdSaveConfig, doctl.ArgKubeConfigExpirySeconds, "", 0,
 		"The length of time the cluster credentials will be valid for in seconds. By default, the credentials are automatically renewed as needed.")
-	AddStringFlag(cmdSaveConfig, doctl.ArgKubernetesAlias, "", "", "The context to save authentication information.")
+	AddStringFlag(cmdSaveConfig, doctl.ArgKubernetesAlias, "", "", "An alias for the cluster context name. Defaults to 'do-<region>-<cluster-name>'.")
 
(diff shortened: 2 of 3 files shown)


danaelhe:
Though the tests runs successfully, I have a mild concern when testing out the command via the cli. I run the command:
` go run cmd/doctl/main.go kubernetes cluster kubeconfig save k8ds-test --alias=newalias_k8ds`
and get the expected response:
```
Notice: Adding cluster credentials to kubeconfig file found in "/Users/myuser/.kube/config"
Notice: Setting current-context to newalias_k8ds
```

I check the "current_context" in the kubeconfig in "/Users/myuser/.kube/config" and get the expected value of the alias, "newalias_k8ds":

```
xxx@xxx doctl % cat /Users/danaelhertani/.kube/config
apiVersion: v1
clusters:
- cluster:
    server: https://xxxx.k8s.ondigitalocean.com
  name: do-nyc1-k8s-1-00-2-do-0-nyc1-1xxx
- context:
    cluster: do-nyc1-k8ds-test
    user: do-nyc1-k8ds-test-admin
  name: newalias_k8ds
current-context: newalias_k8ds
kind: Config
```

but when I run the doctl command:
`go run cmd/doctl/main.go kubernetes cluster kubeconfig show k8ds-test `
it appears the current_context value is the old current_context of "current-context: do-nyc1-k8ds-test", instead of the alias "newalias_k8ds"

I'm wondering if there's just a slight lag or if this will be problematic later?

args.go
```diff
@@ -98,6 +98,8 @@ const (
 	ArgKubernetesLabel = "label"
 	// ArgKubernetesTaint is a Kubernetes taint argument.
 	ArgKubernetesTaint = "taint"
+	// 	ArgKubernetesAlias is a Kubernetes alias argument that saves authentication information under the specified context.
```
ChiefMateStarbuck, line 101:
```suggestion
	// ArgKubernetesAlias is a Kubernetes alias argument that saves authentication information under the specified context.
```

ChiefMateStarbuck (changes requested):
Looking Great!

args.go
danaelhe: good catch! thank you

andrewsomething:
> but when I run the doctl command:
> `go run cmd/doctl/main.go kubernetes cluster kubeconfig show k8ds-test`
> it appears the current_context value is the old current_context of "current-context: do-nyc1-k8ds-test", instead of the alias "newalias_k8ds"

This change only effects the the local file written to disk by the `save` command. The `show` command downloads a new kubeconfig file from the API. So this won't be reflected there.

commands/kubernetes.go
```diff
@@ -399,6 +399,7 @@ This command adds the credentials for the specified cluster to your local kubeco
 	AddBoolFlag(cmdSaveConfig, doctl.ArgSetCurrentContext, "", true, "Boolean indicating whether to set the current kubectl context to that of the new cluster")
 	AddIntFlag(cmdSaveConfig, doctl.ArgKubeConfigExpirySeconds, "", 0,
 		"The length of time the cluster credentials will be valid for in seconds. By default, the credentials are automatically renewed as needed.")
+	AddStringFlag(cmdSaveConfig, doctl.ArgKubernetesAlias, "", "", "The context to save authentication information.")
```
andrewsomething, line 402: How about something like:

```suggestion
	AddStringFlag(cmdSaveConfig, doctl.ArgKubernetesAlias, "", "", "An alias for the cluster context name. Defaults to 'do-<region>-<cluster-name>'.")
```

integration/kubernetes_clusters_kubeconfig_save_test.go
```diff
@@ -119,6 +119,38 @@ var _ = suite("kubernetes/clusters/kubeconfig/save", func(t *testing.T, when spe
...
+			)
+
+			output, err := cmd.CombinedOutput()
+			expect.NoError(err, fmt.Sprintf("received error output: %s", output))
+
+			fileBytes, err := ioutil.ReadAll(f)
+			expect.NoError(err)
+			err = f.Close()
+			expect.NoError(err)
+			expect.Contains(string(fileBytes), fmt.Sprintf("current-context: %s", "newalias_test"))
```
andrewsomething, line 151:
```suggestion
			expect.Contains(string(fileBytes), fmt.Sprintf("current-context: %s", "newalias_test"))
			expect.Contains(string(fileBytes), fmt.Sprintf("name: %s", "newalias_test"))
```

Might be good to also check that the `name` is set.

andrewsomething (approved):
:+1:  Looks great. A couple small suggestions inline.

commands/kubernetes.go
danaelhe: thanks!

integration/kubernetes_clusters_kubeconfig_save_test.go
danaelhe: thanks!

andrewsomething (approved):

ChiefMateStarbuck (approved):
Good to merge on green guardrailscan. great work Dana

PR #1076, opened 2021-12-22
Database Resize Command Clarification

dbrian57:
We received some user feedback via the `doctl` docs where users were trying to downsize database nodes, which is not allowed. I've updated the resize command to reflect this. I also reworded the doc to be a little clearer.


danaelhe (approved):
lgtm on green! Thank you so much!

PR #1089, opened 2022-01-24
Update godo to support new app component type

ZachEddy:
This PR bumps `godo` to incorporate digitalocean/godo#506 changes (released [here](https://github.com/digitalocean/godo/pull/507)), allowing `doctl` users to create and manage apps that have `functions` components.

I verified locally that commands like `apps create` and `apps spec get` work as expected with the new component type.


andrewsomething (approved):
:+1:  LGTM

PR #1098, opened 2022-02-03
Snap Version Fix

ChiefMateStarbuck:
The version was taken by running `version: git` which is a short hand for `version: $(git describe)` which can at times add the commit hash to the end of the version. This change forces the version to be set by the script in the `override-build` section lower in `snapcraft.yml` .


andrewsomething (approved):
:+1:  Nice!

PR #1099, opened 2022-02-03
remove v from snap name:

ChiefMateStarbuck:
The extra `v` is no longer set on the snap name. it was originally set by `version: git`. The new version name will be `doctl_1.70.0_amd64.snap`


andrewsomething (approved):
:shipit:

PR #1100, opened 2022-02-04
Update scripts/version.sh to output desired snap version.

andrewsomething:
This adds a new `--snap` flag to `scripts/version.sh` that is aware if it is being run from a tag and outputs desired snap version. I've added this as a new function to ensure that I don't unintentionally change any other usages of this script.

From a non-tagged version, `+git${SHA}` should be appended to the version. Eg.

```
 ./scripts/version.sh --snap
v1.70.0+git2cdebf6f
```
From a tagged version, it should print tag:

```
 ./scripts/version.sh --snap
v1.70.0
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -66,7 +66,7 @@ jobs:
         id: build
         run: |
           make _build_snap && \
-          find doctl_*.snap -print -exec echo ::set-output name=snap::{} \;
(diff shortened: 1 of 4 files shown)


scotchneat (approved):
Nice. Thank you!

ChiefMateStarbuck:
Oh I see now. Thank you.

PR #1102, opened 2022-02-05
Add sandbox command

joshuaauerbachwatson:
This contribution adds the command `doctl sandbox` with various subcommands:

```
doctl sandbox connect
doctl sandbox deploy
doctl sandbox get-metadata
doctl sandbox init
doctl sandbox install
doctl sandbox status
doctl sandbox uninstall
doctl sandbox watch
doctl sandbox activations get
doctl sandbox activations list
doctl sandbox activations logs
doctl sandbox activations result
doctl sandbox functions get
doctl sandbox functions invoke
doctl sandbox functions list
```

The purpose is to support local development of functions intended for deployment to App Platform when functions components are available there.

The `doctl sandbox` command is hidden unless additional software is installed dynamically.   The development of that software is beyond the scope of this PR.

Further changes (future PRs) will be needed to establish versioning and update, and probably a testing strategy.


joshuaauerbachwatson:
> ⚠️ We detected 3 security issues in this pull request:
> 
> Insecure Use of Dangerous Function (3)

I was unable to open the guardrails dashboard using that link (got a 404) but I looked at the three dangerous function uses that were flagged and read up on the general guardrails doc for those.   I believe the uses (invoking a subprocess where the some arguments appear, at the call site, to be dynamic) is actually safe.   The invoke does not use a shell but rather execs a particular binary (`node`) with a particular body of code (`sandbox.js` ... this is not user-determined).   The arguments passed are dynamic but they are all generated by code internally.  None depends on a user-provided string.

I don't know if I need to do anything to assert this in the code or to try to prove it.   Looking for advice on that.

commands/sandbox.go
```diff
@@ -0,0 +1,370 @@
...
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+const NODE_VERSION = "14.16.0"
```
andrewsomething, line 32: Is there a reason for this specific version? While 14.x is still supported, it looks like 16.x is the current LTS release. 

Looks like there are more recent 14.x releases as well: https://nodejs.org/download/release/latest-v14.x/

commands/sandbox.go
```diff
@@ -0,0 +1,370 @@
...
+	}
+	// Download the nodejs tarball for this os and architecture
+	goos := runtime.GOOS
+	arch := runtime.GOARCH
+	if arch == "amd64" {
+		arch = "x64"
+	}
+	nodeDir := fmt.Sprintf("node-v%s-%s-%s", NODE_VERSION, goos, arch)
+	URL := fmt.Sprintf("https://nodejs.org/dist/v%s/%s.tar.xz", NODE_VERSION, nodeDir)
+	nodeFileName := filepath.Join(tmp, "node-install.tar.xz")
```
andrewsomething, line 105: Looks like Windows might need some special casing here. `runtime.GOOS` for Windows is `windows` while the Node downloads are just `win`.  Is `tar` available there?

commands/activations.go
```diff
@@ -0,0 +1,135 @@
+/*
```
andrewsomething, line 1: These look to have diverged from the code generated by the https://github.com/digitalocean/doctl-sandbox-plugin 

Is that correct?

commands/functions.go
```diff
@@ -0,0 +1,104 @@
...
+	AddStringFlag(get, "save-env-json", "J", "", "save environment variables to FILE as JSON")
+	AddBoolFlag(get, "save", "", false, "save function code to file corresponding to the function name")
+	AddStringFlag(get, "save-as", "", "", "file to save function code to")
+
+	invoke := cmdBuilderWithInit(cmd, RunFunctionsInvoke, "invoke <functionName>", "Invokes a function",
+		`Use `+"`"+`doctl sandbox functions invoke`+"`"+` to invoke a function of your sandbox that has been deployed
+to the cloud.  You can provide inputs and inspect outputs.`,
+		Writer, false)
+	AddBoolFlag(invoke, "web", "", false, "Invoke as a web function, show result as web page")
+	AddStringFlag(invoke, "param", "p", "", "parameter values in KEY VALUE format")
```
andrewsomething, line 49: Trying to do some manual testing and struggling with this. Looking at `nim`, it seems like there is an important difference in how its framework parses flags.

With: 

    nim action:invoke hello --param name world

`name world` are seen as the the input to the `--param` flag.

With:

    ./builds/doctl sandbox functions invoke hello --param=name world

`world` is seen as a positional argument. and returns `Error: (functions.invoke) command contains unsupported arguments`

With:

    ./builds/doctl sandbox functions invoke hello --param="name world"

It is seen as a single value and not a key/value pair and returns `Error: Please provide correct values for flags`

https://github.com/adobe/aio-lib-runtime/blob/becabb33bcd9d95cd3e5da9d8a127a997ccf8fff/src/utils.js#L508

andrewsomething (commented):
A few comments in line with some super early feedback. I need to take a closer pass still.

Re: guardrails: I believe the intention of the warning is just to encourage a closer look at the usage. The dashboard doesn't provide much additional detail. I've flagged them as "Won't Fix" there. That should be enough to silence it.  If it continues being noisy, we can append `// guardrails-disable-line`  to the end of the line. 

commands/sandbox.go
joshuaauerbachwatson: That's the version that is normally baked into an installed version of `nim`.   Yes, it is not current LTS and should be updated.   Since we standardized on it a while back, it is the version I develop on and the version that our GitHub actions run tests on.   I know of no reason why switching to current LTS would break anything.   But, I'd prefer to spin off a task to update `nim` to the later version (should be fine, just needs some testing, including updating the GitHub actions) rather than making the change in this PR.    But, your call.

commands/sandbox.go
joshuaauerbachwatson: There is `tar` for Windows (just about every Unix tool "exists" in that ecosystem).   However, we can't count on it being there.   You are reminding me that there is no shell script for installing `nim` on Windows as there is for other OSs.  We use an installer built using `makensis` to do this ... I think it uses `7zip` to do the unzipping, but I suspect that is bundled into the installer rather than counting on it being on the target machine.   I think we need to get a little Windows-specific expertise on this sub-issue.   Your call on whether we try to fix it in the scope of this PR or make another subtask in SERVERLESS-619 (or wherever you designate).

commands/activations.go
joshuaauerbachwatson: Yes.   I have concluded that we can't keep depending on the codegen, except when importing new bunches of functions.   So, there is a substantial manual editing pass reflected in this PR.

commands/functions.go
joshuaauerbachwatson: Yes.   Embarrassed that I tested `fn invoke` only in the default case (no arguments passed).   The argument passing is broken in this case.

The likely reason is that Adobe I/O uses an atypical parsing pattern for a few of its commands (listed here https://github.com/digitalocean/doctl-sandbox-plugin/blob/a795ffcc72f21e7cbc5ca1fa7df3b8fa912bc2c4/generateCobraWrappers.js#L31).   Oclif actually supports this but I forget the details.  I went to the trouble to add the `multiple` flag as per that table during codegen but then I don't do anything useful with it.  If you look at the table, the only command that we are relying on now in `doctl` is `action:invoke` so hopefully I will not need a completely general fix to fix this.   Will investigate and fix as soon as possible.

joshuaauerbachwatson: This is going to be a limited fix that cannot align exactly with what `nim` does unless you have some insight into cobra options for supporting flags with multiple values and multiple repetitions of the same flag.  AFAIK this is not support currently in cobra but I am hardly an expert on cobra.

The first error
```
(functions.invoke) command contains unsupported arguments
```
comes from cobra and the sandbox isn't even invoked.   So, without some help from cobra we can't support `--param name world` (no quotes, multiple values following the flag).   

On the other hand, we could support `--param "name world"` (quoting the pair) by adding a small fixup (working on it now).

We will hit another cobra limitation because the `--param` flag is also supposed to be capable of being repeated to specify more than one parameter (this is allowed by `oclif` and by `nim`.   But, you can specify multiple parameters like this: `--param "name world address universe"` (specifying multiple parameter pairs with a single `--param` flag.

joshuaauerbachwatson: I have posted a "fix" to this problem (note that quoting is currently needed, so it is not 100% compatible with syntax accepted by `nim`).   But, at least you can invoke functions with parameters 😃 .   You might be a bit concerned (as am I) at the growth of special cases in the `getFlagArgsArray` function but I'm hoping we can discuss the right way to clean that up and hit it on the next pass.

bentranter:
Related to some of the comments here, is it possible to grab `.tar.gz` bundles of Node.js and `doctl-sandbox` instead of `.tar.xz`? If so, that would allow this code to use Go's built in `compress/gzip` and `archive/tar` packages to decompress the archives, removing the dependency on the `tar` command.

commands/functions.go
andrewsomething: I can confirm the change gets this working as documented. I was just looking at `getFlagArgsArray` and thinking it could use some tests :wink: 

Cobra is using `spf13/pflag` under the hood for flags which has `StringSlice` that supports repeated flags.

https://pkg.go.dev/github.com/spf13/pflag#StringSlice

I'd love to drive usage closer to other doctl commands where possible. We have a pattern of using colon separated (e.g. `key:value`) pairs  (see `doctl compute firewall create` for just one example). Paired with `StringSlice`, you can have either `--param foo:bar,name:world` or `--param foo:bar --param name:world` How's that sound?

commands/sandbox.go
andrewsomething: It doesn't necessarily need to happen in this PR as long as we're tracking it elsewhere as we do officially support Windows. Maybe we can just bail out early for now on Windows?

andrewsomething:
@joshuaauerbachwatson  Thinking about how we can approach testing this. I wonder if we can push some of the sandbox pieces into a new "service" that can be mocked. This would make these commands structured much more like the rest of doctl. Shelling out to `nim` could be handles more like any other API. Here's a super quick example of what moving in this direction would look like:

https://github.com/joshuaauerbachwatson/doctl/compare/use-manifest...andrewsomething:asb/nim-mocks

commands/activations.go
joshuaauerbachwatson: I believe we have complete freedom.

joshuaauerbachwatson:
> Related to some of the comments here, is it possible to grab `.tar.gz` bundles of Node.js

Yes.   They offer both.

> and `doctl-sandbox` instead of `.tar.xz`? 

Yes.   Since we control the build of the sandbox we can chose the compression we want.

>If so, that would allow this code to use Go's built in `compress/gzip` and `archive/tar` packages to decompress the archives, removing the dependency on the `tar` command.

I remember looking at that at some point when creating the prototype and I have the feeling I just went with `tar` because it was a bit easier and not for any really good reason.   Once having made that decision, I chose `.xz` since the compression is slightly better.

I think switching to this approach and not using `tar` will basically solve the windows install problem and I'll make a note of that in the follow-on issue.   Thank you @bentranter .

joshuaauerbachwatson:
> @joshuaauerbachwatson Thinking about how we can approach testing this. I wonder if we can push some of the sandbox pieces into a new "service" that can be mocked. This would make these commands structured much more like the rest of doctl. Shelling out to `nim` could be handles more like any other API. Here's a super quick example of what moving in this direction would look like:
> 
> [joshuaauerbachwatson/doctl@use-manifest...andrewsomething:asb/nim-mocks](https://github.com/joshuaauerbachwatson/doctl/compare/use-manifest...andrewsomething:asb/nim-mocks)

Wow @andrewsomething ... I really like that direction, if only because I am not practiced in `go` unit testing techniques and really needed some guidance here.   How do you want to proceed?   I am not sure how "ready to run" your super quick example is.   Can I just turn it into a PR against my branch, merge it, and extend it?   Or should I be more cautious (I realize it was a quick effort)?

andrewsomething:
> Can I just turn it into a PR against my branch, merge it, and extend it? Or should I be more cautious (I realize it was a quick effort)?

I've push some additional changes expanding it some. There is additional refactoring we can do, but it should be safe to merge. We can build off the approach.

commands/sandbox.go
joshuaauerbachwatson: I believe the suggestion by @bentranter to stop using `tar` will allow a straightforward fix for this (though not in this PR, there is a follow-on ticket already).   I'll add a fail-fast in this PR if running on windows.

commands/functions.go
joshuaauerbachwatson: I did end up writing a separate ticket.

joshuaauerbachwatson:
I have merged the PR provided by @andrewsomething .   Running the unit tests on my own mac they are ok.  However, in commit-driven autochecks, I am getting failures with

```
FAIL	github.com/digitalocean/doctl/commands	6.054s
```

Since running in my local environment is normally how I would debug this, and I don't see the failure there, and I'm unfamiliar with how the autocheck workflow is set up in any detail, I could use some help getting to the bottom of this.

andrewsomething:
:facepalm:  The test passes for us because we have the sandbox installed, but the test env does not. We need to expose a way for that to be overridden for testing.

I opened a PR to your branch with a fix: https://github.com/joshuaauerbachwatson/doctl/pull/2

joshuaauerbachwatson:
> The test passes for us because we have the sandbox installed, but the test env does not.

🤦  I should have thought of that.   More generally: is there a way to recover more detail than just the FAIL indicator so next time I might find the problem myself?   I've merged the fix.

andrewsomething (approved):
:+1:  We good to merge into the feature branch. This is a great base to build on.

PR #1104, opened 2022-02-09
Added link for database size slugs

dbrian57:
We've been receiving feedback about not having a list of available size slugs in the docs for `doctl`. As a temporary solution, I've added a bare link in the docs that points to a list of slugs in the API docs. 

Hopefully in the near future we can add Markdown resolution to the doctl docs to make prettier links.


scotchneat (approved):
LGTM. Thanks for adding this!

PR #1107, opened 2022-02-14
sandbox: Add additional tests.

andrewsomething:
This PR adds more tests for the sandbox commands. They are mostly very basic and only test that the correct arguments are passed through to nim. Though I did make some changes to start some deeper testing. Where possible, I've moved to using `CmdConfig.Out` so that we can test the actual command output as well. There should be no functional changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,9 @@ on:
   push:
     branches: [ main ]
   pull_request:
-    branches: [ main ]
+    branches:
+    - main
+    - feature/**
 
 jobs:
   test:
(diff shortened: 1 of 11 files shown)


joshuaauerbachwatson (approved):
This looks good.

To be clear, I did carefully review the changes to non-test code, but for the tests and mocks themselves I am mostly trusting you.

What's the merge protocol?   Given that this is a feature branch that I can write, do I merge or do you?

andrewsomething:
Rebased on main to pick up: https://github.com/digitalocean/doctl/pull/1107/commits/95b228b8662e9e42db74781a7370b01bc3cecdc5

PR #1109, opened 2022-02-16
Versioning and update for 'doctl sandbox'

joshuaauerbachwatson:
This change works in conjunction with a change to the sandbox plugin.  

The sandbox plugin build has been changed to
1.  Include a `version` file in the download.  This file becomes part of the installed sandbox support.
2. Label downloads by encoding the version in the name of the tarball to download.

By convention, version strings encode the Nimbella CLI version followed by the sandbox plugin "shim" version. 

There is an exemplar of this change already out there that can be use to test the present change.   It is at version 2.3.1-1.0.0.

The present change

1.  Adds a constant in `sandbox.go` representing the minimum version of the plugin that this `doctl sandbox` can run with.
2. Generalizes functions that determined if the sandbox is installed, so that they also detect if the sandbox is not up to date.
3. Provides a new error return and a new status in `doctl sandbox status` that expresses that the sandbox is installed but is not up to date.
4. Implements a new subcommand `doctl sandbox upgrade` which installs the "matching" sandbox version.   It shares code with `doctl sandbox install` but takes care to preserve existing credential information and also will preserve `node` if it is at the right version, to avoid using too much download bandwidth.

I have added unit tests for some of the user facing logic in `doctl sandbox [ install | upgrade ]` but these tests don't cover the core operations, which are file-system intensive.   My testing of the actual behavior, thus, was mostly manual.

I would like to merge this and leave. two follow-on items for later.
1.  Adding unit tests for the core install/upgrade operations.   I think some of this will become easier once we re-do some of that logic to stop using `tar` and start using go-native packages instead (definitely not in scope for this PR).   The refactoring to eliminate `tar` can also shift some of the internal interfaces so that more of them can use streams instead of file names and be capable of mocking.
2. Considering whether `doctl sandbox upgrade` should have an option to search for and use the latest version of the sandbox plugin instead of just one that is adequate for the current `doctl`.

### DIFF
--- a/commands/sandbox.go
+++ b/commands/sandbox.go
@@ -60,7 +60,6 @@ connect to the cloud component of the sandbox provided with your account).  Othe
 	}
 
 	// TODO: combine "install" and "connect into a single "enable" command, then "uninstall" should become "disable".
-	// We also need an update strategy.
 	CmdBuilder(cmd, RunSandboxInstall, "install", "Installs the sandbox support",
 		`This command installs additional software under `+"`"+`doctl`+"`"+` needed to make the other sandbox commands work.
 The install operation is long-running, and a network connection is required.`,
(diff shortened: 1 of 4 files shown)


commands/sandbox.go
```diff
@@ -268,12 +250,135 @@ func (c *CmdConfig) PrintSandboxTextOutput(output do.SandboxOutput) error {
...
+		}
+	}
+	fmt.Println("\nDone")
+	return nil
+}
+
+// Check whether the installed node is at the appropriate version
+// Avoids downloading node again during an upgrade
+func canReuseNode(sandboxDir string) bool {
+	return false
```
joshuaauerbachwatson, line 362: Oops.   I forgot to implement this.  It should capture the output of `node --version` and compare it to the node version in the constant.

commands/sandbox.go
```diff
@@ -268,12 +250,135 @@ func (c *CmdConfig) PrintSandboxTextOutput(output do.SandboxOutput) error {
...
+}
+
+func copyExistingNode(from string, to string) error {
+	return nil
```
joshuaauerbachwatson, line 366: Another omission.   This should just copy the node binary from the current (about to be replaced) sandbox directory to the staging one (which is about to be swapped in).

joshuaauerbachwatson:
Also looks like I forgot to update a test.   🤦

joshuaauerbachwatson:
> Also looks like I forgot to update a test. 🤦

Tests are fixed now, and I added a test for the "out of date" case.   I will try to remember to run unit tests pre-push.   I still don't have a test for the upgrade process itself.   I will take advice on how to mock the file system operations.   Meanwhile I'll try to devise a plan to follow for testing all the important cases manually.

On to adding the missing functions.

joshuaauerbachwatson:
> ⚠️ We detected 3 security issues in this pull request:

I remain unclear how to suppress or address these warnings.   I believe guardrails' analysis is not inter-procedural, or at least not powerful enough to detect that all the non-constant arguments in these calls are computed internally without incorporating any user-supplied information.   It's true that the version is read from a file that could have been locally modified by the user (I have been doing this, in fact, to test) but the result is used only to determine whether or not upgrade is required, so I do not see any danger here.

andrewsomething:
> I remain unclear how to suppress or address these warnings.

The guardrails warnings are not blocking. They're mostly just flagging things that might require some additional scrutiny. If it continues being noisy, we can append `// guardrails-disable-line` to the end of the line.

andrewsomething (approved):
:+1: LGTM

PR #1111, opened 2022-02-22
Fix doctl sandbox connect and support additional options

joshuaauerbachwatson:
This change restores functionality to `doctl sandbox connect` that was lost due to changes in the functions cloud.   It uses DigitalOcean APIs to retrieve the necessary information that used to be retrieved by invoking administrative functions.   It also expands the functionality of the command as follows.
```
Usage:
  doctl sandbox connect [<token>|<namespace>] [flags]
```
The argument is now optional.  If it is omitted, the function namespaces of the account are searched to find an eligible sandbox namespace.  If it is provided, it can now be either a namespace name, or a JWT issued by the UI for the namespace (previously only a JWT was accepted).

### DIFF
--- a/commands/sandbox.go
+++ b/commands/sandbox.go
@@ -14,6 +14,7 @@ limitations under the License.
 package commands
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -137,7 +138,11 @@ func RunSandboxConnect(c *CmdConfig) error {
 		return err
 	}
 	token := c.Args[0]
-	result, err := SandboxExec(c, "auth/login", token)
+	creds, err := c.Sandbox().ResolveToken(context.TODO(), token)
+	if err != nil {
+		return err
+	}
+	result, err := SandboxExec(c, "auth/login", "--auth", creds.Auth, "--apihost", creds.ApiHost)
 	if err != nil {
 		return err
 	}
--- a/commands/sandbox.go
+++ b/commands/sandbox.go
@@ -139,10 +139,12 @@ func RunSandboxUninstall(c *CmdConfig) error {
 
 // The connect command
 func RunSandboxConnect(c *CmdConfig) error {
-	var arg string
-	var token bool
-	var creds do.SandboxCredentials
-	var err error
+	var (
+		arg   string
+		token bool
+		creds do.SandboxCredentials
+		err   error
+	)
 	if len(c.Args) > 1 {
 		return doctl.NewTooManyArgsErr(c.NS)
 	}
--- a/commands/sandbox.go
+++ b/commands/sandbox.go
@@ -27,6 +27,7 @@ import (
 	"runtime"
 	"strings"
 
+	"github.com/digitalocean/doctl"
 	"github.com/digitalocean/doctl/do"
 	"github.com/spf13/cobra"
 )
@@ -74,9 +75,14 @@ The install operation is long-running, and a network connection is required.`,
 	CmdBuilder(cmd, RunSandboxUninstall, "uninstall", "Removes the sandbox support", `Removes sandbox support from `+"`"+`doctl`+"`",
 		Writer)
 
-	CmdBuilder(cmd, RunSandboxConnect, "connect <token>", "Connect the cloud portion of your sandbox",
-		`This command connects the cloud portion of your sandbox (needed for testing) by using a provided token.
-You obtain the token from the cloud console (details TBD)`,
+	CmdBuilder(cmd, RunSandboxConnect, "connect [<token>|<namespace>]", "Connect the cloud portion of your sandbox",
+		`This command connects the cloud portion of your sandbox (needed for testing).
+There are several ways to use it.
+1. If your account has been explicitly provisioned with sandbox support, invoke with no arguments.
+2. (DO internal) If you obtained a namespace for your account using the alpha console, you may
+use either the namespace name or the generated token as the argument.  The ability to generate
+and use a token is for compatibility with the previous behavior of the command and is likely
+to be deprecated and removed`,
 		Writer)
 
 	CmdBuilder(cmd, RunSandboxStatus, "status", "Provide information about your sandbox",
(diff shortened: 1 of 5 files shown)


joshuaauerbachwatson:
Temp marking this a draft to fix the tests and possible add more.

joshuaauerbachwatson:
I added another test because there are now two separately mock-able paths through `sandbox connect`, one with at JWT argument and one with a namespace argument.   Assuming they both pass I will mark this reviewable.

commands/sandbox.go
```diff
@@ -132,12 +139,26 @@ func RunSandboxUninstall(c *CmdConfig) error {
 
 // The connect command
 func RunSandboxConnect(c *CmdConfig) error {
-	err := ensureOneArg(c)
+	var arg string
+	var token bool
+	var creds do.SandboxCredentials
+	var err error
```
andrewsomething, line 145:
```suggestion
	var (
		arg   string
		token bool
		creds do.SandboxCredentials
		err   error
	)
```

Just an FYI, you'll more frequently see multiple variables being declared like this in our Go code bases.

do/sandbox.go
```diff
@@ -1,23 +1,73 @@
...
+type inputNamespace struct {
+	Token     string `json:"token,omitempty"`
+	Namespace string `json:"namespace,omitempty"`
+}
+
+// The type of the "namespace" member of the response to API calls.  Only some
+// fields are relevant to each call.
+type outputNamespace struct {
+	ApiHost   string `json:"api_host"`
+	Uuid      string `json:"uuid"`
```
andrewsomething, line 33:
```suggestion
	APIHost   string `json:"api_host"`
	UUID      string `json:"uuid"`
```

Go is opinionated about how "initialisms" should be capitalized. 

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

Unfortunately, this code base does not currently pass `golint` but it would be great to follow its advice for new code. 

```
$ golint do/...
do/sandbox.go:19:2: struct field ApiHost should be APIHost
do/sandbox.go:32:2: struct field ApiHost should be APIHost
do/sandbox.go:33:2: struct field Uuid should be UUID
do/sandbox.go:182:1: context.Context should be the first parameter of a function
do/sandbox.go:199:1: context.Context should be the first parameter of a function
```

do/sandbox.go
```diff
@@ -1,23 +1,73 @@
...
+
+	"github.com/digitalocean/godo"
 )
 
-// SandboxService is an interface for interacting with the sandbox plugin.
+// SandboxCredentials is the type returned by the ResolveToken and ResolveNamespace functions
+// The values in it can be used to connect sandbox support to a specific namespace using the plugin.
+type SandboxCredentials struct {
+	Auth    string
+	ApiHost string
```
andrewsomething, line 19:
```suggestion
	APIHost string
```

Here as well.

do/sandbox.go
```diff
@@ -79,3 +130,91 @@ func (n *sandboxService) Stream(cmd *exec.Cmd) error {
...
+	if err != nil {
+		return SandboxCredentials{}, err
+	}
+	return n.ResolveToken(ctx, token)
+}
+
+// getTokenForNamespace is a subroutine of ResolveNamespace wrapping the call to
+// obtain a JWT.  This step should eventually be eliminated in favor of an API
+// that goes directly from namespace to credentials.
+func getTokenForNamespace(client *godo.Client, ctx context.Context, namespace string) (string, error) {
```
andrewsomething, line 182: Tiny nit. As `golint` points out, it is good practice for `context.Context` to be the first parameter of a function.

do/sandbox.go
```diff
@@ -79,3 +130,91 @@ func (n *sandboxService) Stream(cmd *exec.Cmd) error {
...
+	_, err = client.Do(ctx, req, tokenResponse)
+	if err != nil {
+		return "", err
+	}
+	return tokenResponse.Namespace.Token, nil
+}
+
+// findSandboxNamespace is a subroutine of ResolveNamespace implementing the search for
+// a sandbox namespace.
+func findSandboxNamespace(client *godo.Client, ctx context.Context) (string, error) {
```
andrewsomething, line 199: Same here.

do/sandbox.go
```diff
@@ -79,3 +130,91 @@ func (n *sandboxService) Stream(cmd *exec.Cmd) error {
...
+		return "", err
+	}
+	namespaces := new(namespaceList)
+	_, err = client.Do(ctx, req, namespaces)
+	if err != nil {
+		return "", err
+	}
+	sandboxes := []string{}
+	for _, ns := range namespaces.Namespaces {
+		if strings.Contains(ns.Label, "sandbox") {
```
andrewsomething, line 212: Not blocking for now, but I wonder how future proofed this is? Is a standalone offering, will customers be able to set their own labels with `sandbox` in them?

do/sandbox.go
```diff
@@ -1,23 +1,73 @@
...
+	Uuid      string `json:"uuid"`
+	Key       string `json:"key"`
+	Token     string `json:"token"`
+	Label     string `json:"label"`
+	Namespace string `json:"namespace"`
+}
+
+// postBody is the type of the request body for v2/function/namespaces/... calls that use
+// the POST verb.  Only the relevant parts of the contained inputNamespace need be specified
+type postBody struct {
```
andrewsomething, line 42: A lot of this will eventually want to be in `godo` For now, since these types are shared across the package, it would be great to namespaces them (no pun intended). E.g.

```suggestion
type namespacePostBody struct {
```

do/sandbox.go
```diff
@@ -1,23 +1,73 @@
...
+	Namespace inputNamespace `json:"namespace"`
+}
+
+// responseBody is the type of the response body from API calls other than "list namespaces".
+type responseBody struct {
```
andrewsomething, line 47:
```suggestion
type namespaceResponseBody struct {
```

Same here.

do/sandbox.go
andrewsomething: Ah, just saw the comment about this being subject to change. :+1:

andrewsomething (approved):
:+1:  A couple tiny Go nitpicks in line.

do/sandbox.go
joshuaauerbachwatson: Fixed these, and the `do` directory now passes `golint`.

I notice that it still fails in the `commands` directory.   In the spirit of "follow its advice for new code,"  I plan to do a small follow-on PR to fix lint errors in all the files that are "new" as of `feature/sandbox`.   Not in this PR, though.

do/sandbox.go
joshuaauerbachwatson: Fixed.

do/sandbox.go
joshuaauerbachwatson: Fixed.

do/sandbox.go
joshuaauerbachwatson: Fixed.

joshuaauerbachwatson:
Since all new changes are responsive to the review suggestions, I am merging this now under the original approval.

PR #1112, opened 2022-02-23
Fix lint errors in new sources on feature/sandbox

joshuaauerbachwatson:
In the review of the recent PR #1111, it was pointed out that my additions to `do/sandbox.go` violated `golint` rules.  After fixing those violations, `golint` was running clean on the `do` folder but still failing for `commands`.   @andrewsomething mentioned that existing code was not always lint-conformant but that new code should be if possible.

This PR fixes lint errors for those files introduced by the sandbox effort on the `feature/sandbox` branch.   I did not address failures in older source files.


andrewsomething (approved):
:blue_heart:  Thanks!

PR #1113, opened 2022-02-24
Correct behavior for 'sandbox actv logs --poll'

joshuaauerbachwatson:
This change eliminates a bug in `doctl sandbox activation logs`.   The command was not behaving correctly when one of the flags `--poll`, `--tail`, or `--watch` was specified (these are actually equivalent). 

With these flags set, the command will now keep running, streaming its output to the console, making it appropriate to run in a separate window as designed.

To make this change, I changed the protocol between `doctl` and the sandbox plugin so that the plugin doesn't independently decide when the capture of output should be suppressed.   Rather, `doctl` directs this behavior with a prefixed command.

This in turn required incrementing the minimum required version of the sandbox.  The required version has been staged and will be available to `doctl sandbox upgrade`.

### DIFF
--- a/commands/activations.go
+++ b/commands/activations.go
@@ -64,9 +64,7 @@ for new arrivals.`,
 	AddBoolFlag(logs, "last", "l", false, "Fetch the most recent activation logs (default)")
 	AddIntFlag(logs, "limit", "n", 1, "Fetch the last LIMIT activation logs (up to 200)")
 	AddBoolFlag(logs, "strip", "r", false, "strip timestamp information and output first line only")
-	AddBoolFlag(logs, "tail", "", false, "Fetch logs continuously")
-	AddBoolFlag(logs, "watch", "w", false, "Fetch logs continuously")
-	AddBoolFlag(logs, "poll", "", false, "Fetch logs continuously")
+	AddBoolFlag(logs, "follow", "", false, "Fetch logs continuously")
 
 	result := CmdBuilder(cmd, RunActivationsResult, "result [<activationId>]", "Retrieves the Results for an Activation",
 		`Use `+"`"+`doctl sandbox activations result`+"`"+` to retrieve just the results portion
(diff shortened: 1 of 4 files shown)


andrewsomething (approved):
:+1:  LGTM

Though I wonder from a UX standpoint, if `--poll`, `--tail`, and `--watch` are all equivalent, can we consolidate them? Guess there is some backwards compatibility issues in `nim`, but we can start with a clean slate here. `doctl apps logs` uses `--follow` for streaming continuously.

joshuaauerbachwatson:
> Though I wonder from a UX standpoint, if `--poll`, `--tail`, and `--watch` are all equivalent, can we consolidate them? 

Good idea.  I'm inclined to "just do it" in this PR and merge.  I'll take your "approval with comment" as giving me permission.

> Guess there is some backwards compatibility issues in `nim`  ... `doctl apps logs` uses `--follow` for streaming continuously. 

So, `--follow` it is.   It's a little complicated maybe since there's compatibility with `doctl`, compatibility with `nim`, and compatibility within the sandbox commands.    Re the last of these, note that we have `doctl sandbox watch` not `doctl sandbox follow`, but that's a command, not a flag.   So, I'm thinking that's ok.   On _semantics_ for things that `nim` users have been doing for a long time, I want to be compatible.   But _syntax_ is less important there, as long as we can point out the equivalence.    In the case of `nim activation logs --tail` (etc) that's capability that Adobe introduced pretty recently and we picked it up even more recently.  It's good functionality, but it has relatively little legacy for `nim` users compared to `project watch`.   I believe the `wsk` CLI (which comes along with OpenWhisk but which neither Nimbella nor Adobe uses), has this functionality so users of e.g. IBM Cloud Functions would be aware of it.   Again, as long as we provide the functionality I think we are free to change the spelling.

joshuaauerbachwatson:
As I suggested earlier, I made the change, as part of this PR, to use --follow instead of --poll, --tail, or --watch.   Since @andrewsomething made the suggestion, I figured he approved of the change 🙂  and will merge if all tests pass.

andrewsomething:
Thanks @joshuaauerbachwatson!

PR #1114, opened 2022-02-28
Add an 'undeploy' command

joshuaauerbachwatson:
We deliberately omitted CRUD operations in the `functions` subtree and omitted the `packages` subtree entirely.  This was to avoid confusion between the portion of the sandbox in the local filesystem and the portion in the cloud, and maintain the idea that you push the former to the latter with `deploy`.   As this leaves developers with no way to delete resources from the namespace, we need an `undeploy`.   Given that this is a destructive operation, I made an `--all` flag obligatory when you wish to undeploy everything, and allowed specific resources to be listed on the command line.

```
Usage:
  doctl sandbox undeploy [<package|function>...] [flags]

Flags:
      --all        remove all packages and functions
```
This PR will remain a draft until I have added supporting unit tests.

### DIFF
--- a/commands/sandbox_test.go
+++ b/commands/sandbox_test.go
@@ -367,6 +367,94 @@ func TestSandboxDeploy(t *testing.T) {
 	}
 }
 
+func TestSandboxUndeploy(t *testing.T) {
+	tests := []struct {
+		name            string
+		doctlArgs       []string
+		doctlFlags      map[string]string
+		expectedCmds    []string
+		expectedNimArgs [][]string
+		expectedError   error
+	}{
+		{
+			name:            "no arguments or flags",
+			doctlArgs:       nil,
+			doctlFlags:      nil,
+			expectedCmds:    nil,
+			expectedNimArgs: nil,
+			expectedError:   errUndeployTooFewArgs,
+		},
+		{
+			name:            "with --all flag only",
+			doctlArgs:       nil,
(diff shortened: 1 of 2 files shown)


andrewsomething (approved):
:+1:  LGTM!

commands/sandbox_test.go
```diff
@@ -367,6 +367,94 @@ func TestSandboxDeploy(t *testing.T) {
...
+						if v == "" {
+							config.Doit.Set(config.NS, k, true)
+						} else {
+							config.Doit.Set(config.NS, k, v)
+						}
+					}
+				}
+
+				for i, cmd := range tt.expectedCmds {
+					tm.sandbox.EXPECT().Cmd(cmd, tt.expectedNimArgs[i]).Return(fakeCmd, nil)
```
andrewsomething, line 443: Only suggestion I might make is that the relationship between the `expectedCmds` and `expectedNimArgs` might be unclear to someone not familiar with these tests. 

It might be more clear to create a type for this:

https://github.com/joshuaauerbachwatson/doctl/compare/add-undeploy...digitalocean:asb/nim-undeploy-test

joshuaauerbachwatson:
@andrewsomething many thanks for making your suggestions in the form of PRs.  Seemed right, worked fine, I merged it.  I am merging this now.

PR #1115, opened 2022-03-01
Upgrade node version to 16.13.0

joshuaauerbachwatson:
Also changes the minimum required `nim` release to 2.3.7, since that is the point at which we officially started bundling node 16.x rather than 14.x.

Older versions of `nim` and hence the sandbox would almost certainly work with node 16 but were not explicitly tested with it.


joshuaauerbachwatson:
Sorry.   Accidentally opened this against the wrong base.   Will try again.

PR #1117, opened 2022-03-02
goreleaser: Don't use release label for snapshots.

andrewsomething:
Built a local snapshot with goreleaser and realized `doctl version` shows the `release` label. This makes it so that the `snapshot` label is used when building with `goreleaser build --snapshot`


bentranter (approved):
LGTM pending CI

PR #1119, opened 2022-03-06
Reject 'doctl sbx init --language' value if the backend has no runtime for it.

joshuaauerbachwatson:
Previously, a language would be accepted if it was in a static list of languages for which we know how to generate samples.  But, not all languages in that list necessarily have runtimes in the backend, which was leading to a situation where the developer could initialize a local sandbox area that could not be deployed.

Most of the change is in `nim`, hence the required minimum version of the sandbox is incremented.

But, help texts had to change because we can no longer enumerate the supported languages in the usage help for the `--language` flag.   Contacting the controller is an asynchronous operation, and we don't want to unduly delay or hang usage help (also a bit unclear how to shoe-horn such an operation into `cobra`).    The `nim` solution refers the developer to the separate `nim info --runtimes` command.   We did not have an import of that into `doctl` so I added it but called it `doctl sbx status --languages`.   The `doctl sbx init` help is then augmented appropriately.

The check for. a runtime is omitted if you issue `doctl sbx init` before `doctl sbx connect` because, in that case, we don't have an API host to interrogate.   The language selection is just accepted as before.   Arguably, we should refuse to perform `init` until after `connect`.   But, once the namespace APIs settle down into their final configuration and we are able to do the connect operation by default at install time, the unconnected state should rarely arise and the special rule would no longer be needed.

### DIFF
--- a/commands/sandbox-extra.go
+++ b/commands/sandbox-extra.go
@@ -29,9 +29,9 @@ func SandboxExtras(cmd *Command) {
 		`The `+"`"+`doctl sandbox init`+"`"+` command specifies a directory in your file system which will hold functions and
 supporting artifacts while you're developing them.  When ready, you can upload these to the cloud for
 testing.  Later, after the area is committed to a `+"`"+`git`+"`"+` repository, you can create an app from them.
-`,
+Type `+"`"+`doctl sandbox status --languages`+"`"+` for a list of supported languages.`,
 		Writer)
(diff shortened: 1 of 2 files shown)


ChiefMateStarbuck (approved):

ChiefMateStarbuck:
thank you @joshuaauerbachwatson

PR #1120, opened 2022-03-09
Increment the 'nim' version to 2.3.9

joshuaauerbachwatson:
The Nimbella CLI version 2.3.9 contains two fixes relevant to `doctl sbx` commands.

1. The use of initial './' sequences in .include file entries was causing incorrect behavior.   Although unnecessary, these sequences should be harmless.
2. Incremental deploy (used by 'doctl sbx watch') was not detecting environment changes.   If the change was made to `project.yml` or to a contained `.env` file the watcher would "fire" as per the spec.   However, when calculating whether actions had changed it failed to include the `environment` clause and the affected actions would be unchanged.


joshuaauerbachwatson:
I closed this because I once again based it on the wrong branch.   Sorry.  Will open another.

PR #1122, opened 2022-03-09
sandbox: Use Go native libs to extract archives.

andrewsomething:
This replaces shelling out to `tar` with a new utility package. It is the first step to getting sandbox support working for Windows. The Node download for Windows only comes as a zip, hence the support for extracting zips in the utility package.

### DIFF
--- a/commands/sandbox.go
+++ b/commands/sandbox.go
@@ -29,30 +29,31 @@ import (
 
 	"github.com/digitalocean/doctl"
 	"github.com/digitalocean/doctl/do"
+	"github.com/digitalocean/doctl/pkg/extract"
 	"github.com/spf13/cobra"
 )
 
-const nodeVersion = "v16.13.0"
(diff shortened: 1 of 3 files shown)


andrewsomething:
Guardrails is drunk... No idea why it thinks there are any SQL queries in this code.

> File Traversal When Extracting zip/tar Archive

Both functions already have code specifically meant to address this concern, e.g.:

                path := filepath.Join(target, header.Name)
		if !strings.HasPrefix(path, filepath.Clean(target)+string(os.PathSeparator)) {
			return fmt.Errorf("illegal file path: %s", path)
		}

And our current usage does not actually take user input.

joshuaauerbachwatson (approved):
LGTM.   I've verified both the unit tests and the ability to install correctly on mac with the new support.   I am in the process of verifying that we can install on windows.   I'm fine with going ahead and merging this at any time.  If I find any problems with the windows install we can pursue that post-merge.

PR #1123, opened 2022-03-09
sandbox: Support Windows node downloads.

andrewsomething:
As mentioned in https://github.com/digitalocean/doctl/pull/1122, the Node download for Windows only comes as a zip. Additionally, I found that Windows (or at least the version of PowerShell I was testing in) will not run it without the `exe` file extension. So this required special casing Windows in a few places.

### DIFF
--- a/commands/command_config.go
+++ b/commands/command_config.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"io"
 	"path/filepath"
+	"runtime"
 
 	"github.com/digitalocean/doctl"
 	"github.com/digitalocean/doctl/commands/displayers"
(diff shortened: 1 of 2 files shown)


scotchneat (approved):
I haven't tested this on Windows yet but the code looks good.

joshuaauerbachwatson:
> I haven't tested this on Windows yet but the code looks good.

Looks good to me also.   I am in the process of expanding our internal (intended to evolve to limited-external) download site for feature/sandbox doctl builds to include all supported OS/arch combinations.   I am folding some windows testing into that.   If I don't find problems I'll merge.   Of course, if some else merges first, no biggie.   Since we aren't officially supporting windows yet, there will be no regression.

joshuaauerbachwatson (approved):
I was able to install on windows, then connect to a sandbox, then run some commands.  I am not set up to run a full test suite on windows (and we still need to flesh out the integration tests for sandbox), but this is good enough for me.   @andrewsomething thank you for taking this.

PR #1124, opened 2022-03-12
Improvements to flag handling and include/exclude for 'web'

joshuaauerbachwatson:
This PR addresses two known specific problems with flag handling and changes the strategy for "flag substitution" when calling `nim` to be less confusing and fragile.

1.  Previously, if the developer had a package called `web` and attempted to control its deployment with `--include` or `--exclude`, the results would be incorrect because `web` has special meaning to `nim`.
2. Previously, the `sandbox functions invoke` command `--param` flag required KEY and VALUE to be blank separated, which, in the context of cobra string parsing required quoting.   We now support (require) a colon separator consistent with other syntax decisions in `doctl`.  Also, the value is a string slice now, and generalizes to specifying multiple KEY:VALUE pairs.
3. In conjunction with the specific problems, but generalizing from them, the code in the `getFlagArgsArray` function has been cleaned up so that there is no longer special case processing of particular flags there.   This was fragile because it assumed it knew the meaning of these flags (which, in the long run, might fail if other commands used the same flags with a different meaning).  It was also confusing and likely to grow over time and become more confusing.   The new strategy is, in most cases, for the individual commands to fix up their flags to meet `nim` assumptions by manipulating the `CmdConfig` state prior to calling `getFlatArgsArray`.   For `invoke --param`, this won't quite work because of the many-for-one substitution at the level of tokens.  So, in that case, the command appends to the result of `getFlatArgsArray` after calling it.


commands/activations.go
```diff
@@ -111,10 +112,12 @@ func RunActivationsLogs(c *CmdConfig) error {
 	if argCount > 1 {
 		return doctl.NewTooManyArgsErr(c.NS)
 	}
+	replaceFunctionWithAction(c)
+	augmentPackageWithDeployed(c)
 	if isWatching(c) {
-		return RunSandboxExecStreaming("activation/logs", c, []string{"last", "strip", "follow"}, []string{"function", "package", "limit"})
+		return RunSandboxExecStreaming("activation/logs", c, []string{"last", "strip", "follow", "deployed"}, []string{"action", "package", "limit"})
```
ChiefMateStarbuck, line 118: The values passed to this function can be top file `consts` for better maintainability

commands/activations.go
```diff
@@ -134,9 +137,30 @@ func RunActivationsResult(c *CmdConfig) error {
...
 	return c.PrintSandboxTextOutput(output)
 }
+
+// replaceFunctionWithAction detects that --function was specified and renames it to --action (which is what nim
+// will expect to see).
+func replaceFunctionWithAction(c *CmdConfig) {
+	value, err := c.Doit.GetString(c.NS, "function")
+	if err == nil && value != "" {
+		c.Doit.Set(c.NS, "function", "")
+		c.Doit.Set(c.NS, "action", value)
```
ChiefMateStarbuck, line 154: These strings, `action`, `function` can also be consts

ChiefMateStarbuck (changes requested):
Thank you. Many of the strings passed directly into functions can become top level `vars` or `consts` for easier maintainability.

commands/activations.go
ChiefMateStarbuck: We can focus on the files in this PR alone. `functions.go` and `sandbox-extra.go` can also be updated to use consts. 

The string values don't look like they repeat. For now, we can continue to put the const values at the beginning of each file.

The other files I can look at and update accordingly .

ChiefMateStarbuck: If the string values repeat, then a `commands/const.go` file can be filled with all the consts so that the command files do not have to borrow values from each other. I would suggest moving all the const definitions there.

joshuaauerbachwatson: I pushed a change to implement the suggestion.   I notice, though, that the auto-checks appear stuck.
Hope we can converge soon.

ChiefMateStarbuck (approved):

PR #1126, opened 2022-03-17
Incorporate a newer Nimbella CLI version with a bug fix

joshuaauerbachwatson:
The fix in Nimbella CLI 2.3.10 addresses a problem with remote build when using the default package.


andrewsomething (approved):
:+1:  Good on green

PR #1127, opened 2022-03-18
Add ability to get 'app' project resources

rcj4747:
Apps can be assigned to projects.  This patch adds support for getting an project resource that is an app by running 'doctl resources get do:app:APP_UUID'.


andrewsomething (approved):
:+1:  LGTM, thank @rcj4747!

PR #1133, opened 2022-03-24
Use new simplified API to get sandbox namespace

joshuaauerbachwatson:
This removes the need for the namespace to be added via the cloud console or administratively prior to connecting.  The new API creates the sandbox namespace if it doesn't already exist and doesn't require you to know anything about it a priori.

_Still a draft until I fix the tests_.


andrewsomething (approved):
:+1:  LGTM!

PR #1135, opened 2022-03-27
Maintain sandbox credentials on per-DO-token basis

joshuaauerbachwatson:
When `doctl sandbox connect` is executed, the sandbox credentials are returned and stored for a specific account or team identified by the DO API token used for the call.   But, `doctl` anticipates the use of multiple DO API tokens on a given machine, generally stored in the form of (locally) named "contexts" but also directly specifiable on the command line.  The previous practice of storing the returned sandbox credentials in a single place is thus inappropriate.   This change allows for multiple credential sets to be stored, identified by a short prefix of the token that is in use at the time of connection.   This change then allows for correct semantics when `doctl auth switch` is used or any of the `--context`, `--config` or `--access-token` flags appear on a sandbox command, altering the effective identity of the requestor.

When the sandbox was connected under one identity and the user switches to another identity for which the sandbox was not previously connected, the sandbox becomes disconnected.   This is by design.  It allows the user to do a fresh connect under the new identity.   In general this will connect to a different sandbox.

Note that the underlying Nimbella CLI is capable of storing credentials for multiple namespaces.   I did not want to load the present requirement onto that capability because in the future we are likely to allow multiple namespaces to be used within a single "sandbox."   The current change is simply maintaining the invariant that different tokens imply different identities and hence different resource sets.

### DIFF
--- a/commands/command_config.go
+++ b/commands/command_config.go
@@ -37,7 +37,7 @@ type CmdConfig struct {
 	setContextAccessToken func(string)
 	removeContext         func(string) error
 	checkSandboxStatus    func() error
-	installSandbox        func(string, bool) error
+	installSandbox        func(*CmdConfig, string, bool) error
 
 	// services
 	Keys              func() do.KeysService
@@ -129,7 +129,7 @@ func NewCmdConfig(ns string, dc doctl.Config, out io.Writer, args []string, init
 			}
 			node := filepath.Join(sandboxDir, nodeBin)
 			sandboxJs := filepath.Join(sandboxDir, "sandbox.js")
-			nimbellaDir := filepath.Join(sandboxDir, ".nimbella")
+			nimbellaDir := getCredentialDirectory(c, sandboxDir)
 			c.Sandbox = func() do.SandboxService { return do.NewSandboxService(sandboxJs, nimbellaDir, node, godoClient) }
 
 			return nil
@@ -196,8 +196,8 @@ func NewCmdConfig(ns string, dc doctl.Config, out io.Writer, args []string, init
 			return CheckSandboxStatus()
 		},
 
-		installSandbox: func(dir string, upgrading bool) error {
-			return InstallSandbox(dir, upgrading)
+		installSandbox: func(c *CmdConfig, dir string, upgrading bool) error {
+			return InstallSandbox(c, dir, upgrading)
 		},
 	}
 
--- a/commands/sandbox.go
+++ b/commands/sandbox.go
@@ -34,12 +34,21 @@ import (
 )
 
 const (
-	nodeVersion       = "v16.13.0"
-	minSandboxVersion = "2.3.10-1.1.0"
+	// Minimum required version of the sandbox plugin code.  The first part is
+	// the version of the incorporated Nimbella CLI and the second part is the
+	// version of the bridge code in the sandbox plugin repository.
+	minSandboxVersion = "3.0.0-1.1.0"
+
+	// The version of nodejs to download alongsize the plugin download.
+	nodeVersion = "v16.13.0"
 
 	// noCapture is the string constant recognized by the plugin.  It suppresses output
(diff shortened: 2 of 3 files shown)


commands/command_config.go
```diff
@@ -196,8 +196,8 @@ func NewCmdConfig(ns string, dc doctl.Config, out io.Writer, args []string, init
 			return CheckSandboxStatus()
 		},
 
-		installSandbox: func(dir string, upgrading bool) error {
-			return InstallSandbox(dir, upgrading)
+		installSandbox: func(c *CmdConfig, dir string, upgrading bool) error {
```
joshuaauerbachwatson, line 199: To support upgrading from legacy, the install command needs access to the governing access token.  It seemed the easiest way was to pass along the CmdConfig.

commands/sandbox.go
```diff
@@ -34,12 +34,21 @@ import (
 )
 
 const (
-	nodeVersion       = "v16.13.0"
-	minSandboxVersion = "2.3.10-1.1.0"
+	// Minimum required version of the sandbox plugin code.  The first part is
+	// the version of the incorporated Nimbella CLI and the second part is the
+	// version of the bridge code in the sandbox plugin repository.
+	minSandboxVersion = "3.0.0-1.1.0"
```
joshuaauerbachwatson, line 40: The version needed to be upgraded anyway, but this change is also important for its side-effect of forcing legacy users to run an upgrade once they download "this" version of doctl.   The upgrade step includes the renaming from ".nimbella" to a conforming name for credential storage.

commands/sandbox.go
```diff
@@ -347,18 +356,17 @@ func RunSandboxExecStreaming(command string, c *CmdConfig, booleanFlags []string
...
+		_, err = fmt.Fprintln(c.Out, strings.Join(output.Captured, "\n"))
 	} else if len(output.Table) > 0 {
-		fmt.Fprintf(c.Out, genericJSON(output.Table))
+		_, err = fmt.Fprintln(c.Out, genericJSON(output.Table))
 	} else if output.Entity != nil {
-		fmt.Fprintf(c.Out, genericJSON(output.Entity))
+		_, err = fmt.Fprintln(c.Out, genericJSON(output.Entity))
 	} // else no output (unusual but not impossible)
 
-	_, err := fmt.Fprintln(c.Out)
```
joshuaauerbachwatson, line 360: This change is unrelated to the substance of the PR but one user complained that there is a spurious newline for commands that produce no output.   I suppose it's arguable whether this matters and/or which behavior is better but I thought to remove that one possible irritant.   The original `nim` behavior was no newline.

commands/sandbox.go
```diff
@@ -562,6 +599,16 @@ func getSandboxDirectory() (string, bool) {
 	return sandboxDir, exists
 }
 
+// getCredentialDirectory returns the directory in which credentials should be stored for a given
+// CmdConfig.  The actual leaf directory is a function of the access token being used.  This ties
+// sandbox credentials to DO credentials
+func getCredentialDirectory(c *CmdConfig, sandboxDir string) string {
+	token := c.getContextAccessToken()
+	runes := []rune(token)
+	leafDir := string(runes[:8])
```
andrewsomething, line 608: The DigitalOcean API token format is currently being transitioned. Soon all new tokens will contain an identifiable, versionized prefix and match this regex: `^dop_v\d_([0-9a-f]){64}$`. So 8 characters might not be enough here. Wonder if we can just use a hash?

andrewsomething (changes requested):

commands/sandbox.go
joshuaauerbachwatson: Certainly.   That will guarantee uniqueness regardless of token format.   I will make that change.

joshuaauerbachwatson: Done.

andrewsomething (approved):
:+1:  LGTM!

PR #1143, opened 2022-04-11
Fix potential loop in `apps update --wait`

### LINKED ISSUE #1142: `apps update --wait` may hang forever when the deployment is a no-op
Hi,

I'm trying to use `doctl apps update --wait` in a continuous deployment pipeline. Sometimes this works fine, other times the command hangs until max attempts. Looking at the source, I think I can see the problem.

https://github.com/digitalocean/doctl/blob/62602dba745d478962fa029f2d2a2c61bb4f93e0/commands/apps.go#L352-L355

This is where the `wait` flag is applied. `waitForActiveDeployment` is called with an empty string as the deployment ID.

https://github.com/digitalocean/doctl/blob/62602dba745d478962fa029f2d2a2c61bb4f93e0/commands/apps.go#L452-L460

Inside `waitForActiveDeployment`, if the deployment ID is empty the app is fetched and `deploymentID` is set from the `InProgressDeployment` **if it's non-nil**.

However, what I can see if I hit the `update` command a few times is that sometimes `in_progress_deployment` isn't set right after updating, presumably when there are no changes to make. When this is the case, `update --wait` will loop for `maxAttempts` since `deploymentID` will remain `""` and `app.InProgressDeployment.ID` will remain `nil`.

I'm not too sure of the semantics of the DO API, but the response doesn't seem to include anything that could strictly correlate to the update. If the `update` response could be assumed to contain the `InProgressDeployment` if it's not a no-op, then it might work to only call `waitForActiveDeployment` if `app.InProgressDeployment != nil`, and pass long the `DeploymentID` otherwise.

connec:
- f11ab090 **Fix potential loop in `apps update --wait`**

  When there are no changes to make, it's possible for there to never be
  an `InProgressDeployment` returned from the API – this is more likely
  still when the `InProgressDeployment` from the `update` API is ignored,
  and later fetched via `Get` (as is currently the case).
  
  This fix assumes that `update` will return an `InProgressDeployment` if
  there is one (e.g. if it's not a no-op), so if `InProgressDeployment` is
  `nil` it treats this as a successful deployment.
  
  Fixes #1142.


andrewsomething (changes requested):
Hi @connec, 

Thanks for flagging this problem for us and submitting this PR. Unfortunately, I don't think this this will work as expected. The `InProgressDeployment` is generally not known and returned as part of the initial response to the update request. 

Even in a "no-op" update, a deployment is created. It seems like our current logic is flawed though. Fetching the app to find the `InProgressDeployment` doesn't work as it completes before we find it. 

I think we need to adjust the logic to retrieve a list of deployments for an app rather than the app itself. Then we can use the ID of the most recent deployment even if it has completed.

Perhaps something like this: 

```diff
diff --git a/commands/apps.go b/commands/apps.go
index bbdbbd41..219c0433 100644
--- a/commands/apps.go
+++ b/commands/apps.go
@@ -450,13 +450,13 @@ func waitForActiveDeployment(apps do.AppsService, appID string, deploymentID str
 		}
 
 		if deploymentID == "" {
-			app, err := apps.Get(appID)
+			deployments, err := apps.ListDeployments(appID)
 			if err != nil {
 				return err
 			}
 
-			if app.InProgressDeployment != nil {
-				deploymentID = app.InProgressDeployment.ID
+			if deployments[0] != nil {
+				deploymentID = deployments[0].ID
 			}
 		} else {
 			deployment, err := apps.GetDeployment(appID, deploymentID)
```

connec:
I took the approach you suggested, and this also resolves my issue. I've updated the integration tests this time as well, so hopefully CI would be green.

andrewsomething (approved):
:+1:  

Sorry for the delay in merging, I was out of town last week. Thanks again!

MattIPv4:
Hey, @connec - Thanks a ton for this sweet PR! 😄

PR #1145, opened 2022-04-12
Increment to nim version 3.0.4 to obtain better remote build behavior

joshuaauerbachwatson:
The newer 'nim' invokes newer configurations of the builder system functions, allowing more efficient and secure remote builds.

Also includes a change so that builds of `golang` sandbox projects are remote by default.


andrewsomething (approved):
:+1:

PR #1148, opened 2022-04-16
Support e2e tests with added env vars

joshuaauerbachwatson:
To support the new e2e tests for sandbox functionality (running under `bats` in the companion `doctl-sandbox-plugin` repo), it will often be necessary to manipulate the `minSandboxVersion` variable so that it can be the next (rather than current) version that is tested.    For clean tests that are not contaminated by (and do not contaminate) any existing installation, we also need to relocate the sandbox installation area for the test.   This is especially important when we add e2e tests for install and upgrade.

This PR simply adds support for two new environment variables, `minSandboxVersion`, and `sandboxDirectory` to be used for this purpose.   These will not be documented for production use by customers.

### DIFF
--- a/commands/sandbox.go
+++ b/commands/sandbox.go
@@ -39,7 +39,7 @@ const (
 	// Minimum required version of the sandbox plugin code.  The first part is
 	// the version of the incorporated Nimbella CLI and the second part is the
 	// version of the bridge code in the sandbox plugin repository.
-	minSandboxVersion = "3.0.4-1.2.0"
+	minSandboxVersion = "3.0.5-1.2.0"
 
 	// The version of nodejs to download alongsize the plugin downl
 	nodeVersion = "v16.13.0"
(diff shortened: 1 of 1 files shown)


commands/sandbox.go
```diff
@@ -39,7 +39,7 @@ const (
 	// Minimum required version of the sandbox plugin code.  The first part is
 	// the version of the incorporated Nimbella CLI and the second part is the
 	// version of the bridge code in the sandbox plugin repository.
-	minSandboxVersion = "3.0.4-1.2.0"
+	minSandboxVersion = "3.0.5-1.2.0"
```
joshuaauerbachwatson, line 42: Orthogonal to the purpose of the PR, but, meanwhile, there is a new `nim` that fixes some vulnerabilities.

ChiefMateStarbuck (approved):

PR #1149, opened 2022-04-18
Store API host with different final domain segment

joshuaauerbachwatson:
This change is in support of an upcoming change to serverless clusters whereby they will have two domain names.   The API call to retrieve the sandbox namespace will, by convention, only have one of these names.  We will need to map it, conditionally, to the other.   The prefix of the namespace name ("fn-" or not) is consulted to decide whether the remapping should occur.

All fn* namespaces should be mapped to the cloudflare managed domain - doserverless.co

All other namespaces continue to be mapped to doserverless.io

This is to ensure backward compatibility for existing sandbox namespaces.

### DIFF
--- a/do/sandbox.go
+++ b/do/sandbox.go
@@ -20,6 +20,7 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
+	"strings"
 
 	"github.com/digitalocean/godo"
 )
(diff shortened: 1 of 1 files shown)


piyush0101 (approved):

danaelhe:
Looks good! Thank you.

PR #1150, opened 2022-04-20
Fail early and cleanly when the sandbox was never connected

joshuaauerbachwatson:
Beta users of `doctl sandbox` have reported that a common user error is handled in a confusing way.   If the user has done `doctl sandbox install` but not `doctl sandbox connect`, and then tries other sandbox commands, the errors that come back are arbitrary and confusing because they are attempting to invoke various APIs on the serverless backend.   This change inserts a pre-check for being connected, prior to calling out to the sandbox plugin, meaning that the user always gets the informative message saying that the sandbox needs to be connected.

It is important that the pre-check not double the overhead of most calls, which is what would happen if we invoked the plugin to do the credential check.   So, the check uses a simple file existence check, which is slightly weaker since it doesn't  check for possible corruption of the credentials (etc).  The deeper check is done when `sandbox status` is called.

This PR also includes a small change to sandbox plugin DEBUG support.  It had become incomplete due to an earlier refactoring.

### DIFF
--- a/commands/sandbox.go
+++ b/commands/sandbox.go
@@ -51,6 +51,9 @@ const (
 	// credsDir is the directory under the sandbox where all credentials are stored.
 	// It in turn has a subdirectory for each access token employed (formed as a prefix of the token).
 	credsDir = "creds"
+
+	// credentialsFile is the name of the file where the sandbox plugin stores OpenWhisk credentialsl
+	credentialsFile = "credentials.json"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


commands/sandbox.go
```diff
@@ -292,28 +302,26 @@ func deletePackage(c *CmdConfig, pkg string) error {
...
 	cmd, err := sandbox.Cmd(command, args)
 	if err != nil {
 		return do.SandboxOutput{}, err
 	}
-	// If DEBUG is specified, we need to open up stderr for that stream.  The stdout stream
-	// will continue to work for returning structured results.
-	if os.Getenv("DEBUG") != "" {
-		cmd.Stderr = os.Stderr
-	}
-
```
joshuaauerbachwatson, line 309: This code was relocated to `sandbox.Cmd` where it will catch all paths that end up invoking the plugin.

danaelhe (approved):
This looks like a good approach to the problem, LGTM

PR #1151, opened 2022-04-22
Fixes to (individually) small bugs

joshuaauerbachwatson:
1.  There was an error in `doctl sbx upgrade` on windows due to confusion between `node` and `node.exe`.
2. There were errors during install when the temporary file area is on a different file system from the `doctl` config area.  It was using `os.Rename`.   Rather than introducing a higher overhead file move, this was changed to allocate the temporary staging area under the doctl config area so that `os.Rename` would always work.   A cleanup step was added to delete this area.
3. There were "branding" errors involving `doctl sbx watch`.   When accidentally typed without args you got a `nim` help screen.   When looking at the deployment transcript you would see a nonsensical mixture of `nim` and `doctl` syntaxes.   This was fixed partly in `doctl` and partly in the sandbox plugin (see next).
4. The `nim` and sandbox plugin versions were incremented, partly in service of the previous fix but also to pick up several useful fixes to `nim`.   See release notes: https://github.com/nimbella/nimbella-cli/releases/tag/v3.0.7


joshuaauerbachwatson:
Guardrails is complaining about

> cmd := exec.Command(fullNodeBin, "--version")

I believe it is unhappy because `fullNodeBin` is now calculated more indirectly than it used to be and guardrails is pretty bad at doing inter-procedural analysis.   I believe it is easy to be satisfied that the only values that can be executed here are `path/to/node` and `path/to/node.exe` depending on the os.

andrewsomething (approved):
:+1:  LGTM

PR #1154, opened 2022-04-28
Move the sandbox feature to the main branch

joshuaauerbachwatson:
The `feature/sandbox` has been long-lived and has been the target, to date, of all work on the `doctl sandbox` (alias `sbx`, `serverless`, `sls`) subtree.

The contents of that branch has been the source of "dev" ("beta", "pre-release") versions of `doctl` that I have been building with `goreleaser` but then uploading via manually initiated scripts to locations that have been made known to beta users.

As discussed with @andrewsomething, the purpose of doing the merge now is to work toward eliminating these special builds and releasing this feature as part of official `doctl` with some lead time prior to full support.

Note that after this PR is merged, the next release of `doctl ` will
- still have the `sandbox` tree hidden until an install step is performed
- still have the `sandbox` tree labeled as "beta" once it does become visible

Unhiding the feature and/or removing the "beta" label is not part of this PR but would presumably be done when we are ready to give full public support to the feature.


Blanketjack (commented):
Changes Commited

andrewsomething (commented):
:rocket:  Let's do it!

andrewsomething (approved):

PR #1157, opened 2022-05-03
Increment nim version (improves error reporting)

joshuaauerbachwatson:
This change increments the included `nim` version from 3.0.8 to 3.0.10.   Benefits:

- a more comprehensible error message when a function build for `nodejs` includes an invalid `package.json` file
- the ability to see more information about remote build errors when debugging is enabled


andrewsomething (approved):
:+1:

PR #1159, opened 2022-05-10
Switch from GET to POST when connecting to sandbox

joshuaauerbachwatson:
It has been requested by security to switch from GET to POST when connecting to the sandbox.

The portal API has already been amended to accept both for the time being so we can cut over.

This PR accomplishes the switch.   Once this is approved and merged, a new `doctl` release will be needed before support for GET can be removed from the portal API.


andrewsomething (approved):
:+1:  lgtm

PR #1160, opened 2022-05-10
Improve friendliness of the languages table

joshuaauerbachwatson:
The current output of `doctl sbx status --languages` is not helpful in telling the user what keywords will actually be accepted on `doctl sbx init <path> --language <?>`.   It conflates "runtime categories" like `nodejs` with real languages (like `JavaScript`) and might also imply to some users that they need to (or even can) specify the exact release of a language runtime in the `doctl sbx init` command.

This change attempts to be clearer about the terminology and give a more helpful display.  It is designed so that the static table of languages compiled into `doctl` is inclusive of all languages for which `nim` had samples, but the actual table that is displayed is responsive to the backend reporting which runtimes are currently available.   This means that changes to the `doctl` code will not be required when (1) a new runtime is deployed for a language in the list, (2) a runtime is deprecated, (3) the default version of a runtime changes.


andrewsomething (approved):
:+1:  lgtm

PR #1161, opened 2022-05-12
Unhide the sandbox and remove beta indicator

joshuaauerbachwatson:
This PR removes the logic that hides the sandbox command when sandbox support is not installed.  It also removes the "beta" indicator from the command's help.


andrewsomething (approved):
:shipit:  LGTM!

PR #1163, opened 2022-05-13
Add AppPlatformApp to Projects documentation

rcj4747:
This PR adds App Platform apps to the documented list of resources
supported by projects and adds an example URN for an app.


ChiefMateStarbuck (changes requested):
Thank you

a test would be beneficial. possibly adding a line here https://github.com/digitalocean/doctl/blob/main/commands/projects_test.go#L219

rcj4747:
> Thank you
> 
> a test would be beneficial. possibly adding a line here https://github.com/digitalocean/doctl/blob/main/commands/projects_test.go#L219

Sounds good, I can do that.  Marking the PR as a draft until I get it added.

ChiefMateStarbuck (approved):

ChiefMateStarbuck (approved):

PR #1170, opened 2022-05-25
Fix bug in handling --param values with colons

joshuaauerbachwatson:
In `doctl sls fn invoke` the syntax `--param  key:value` is mandated for passing parameters.   Because the code is currently handing off to the sandbox plugin, where the syntax is `--param key value`, there is a translation between the two in the code.   This change fixes a flaw in the translation:  it was not taking into account that the value might contain a colon.   So, for example, `--param url:https://example.com` was failing with a spurious and confusing error message.


andrewsomething (approved):
LGTM

PR #1173, opened 2022-05-26
sandbox: Support running when installed as a Snap.

andrewsomething:
In order to support running when installed as a Snap, we need to ship the sandbox and node in the snap package. To do so, we:

- Override the sandbox directory both during install and at runtime so that the sandbox is installed to `/snap/doctl/current/opt/sandbox`
- Run `doctl sandbox install` during the snap build.
- When run as a snap, the sandbox credentials directory ignores the `OVERRIDE_SANDBOX_DIR` environment variable as `/snap/doctl/current/opt/sandbox` is not writable. The default directory in the user's home will always be used (`$XDG_CONFIG_HOME/doctl/sandbox/creds`).
- Make sure the credential directory is actually created as this normally happens when `doctl sandbox install` is run, but that happens on the build host for snaps.
- Handle `ErrSandboxNotConnected` in both `RunSandboxInstall` and `RunSandboxUpgrade` to prevent them from attempting to install/upgrade the sandbox when it is not necessary. Without this, attempting to run these commands as a snap will fail.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -29,6 +29,11 @@ parts:
     after: [launcher]
     source: .
     plugin: go
+    build-environment:
+    # Results in installation to /snap/doctl/current/opt/sandbox
+    - OVERRIDE_SANDBOX_DIR: "$SNAPCRAFT_PART_INSTALL/opt/sandbox"
+    # Override to prevent 'invalid cross-device link' calling doctl
+    - XDG_CONFIG_HOME: "$SNAPCRAFT_PART_INSTALL/tmp/.config"
     override-pull: |
       git clone https://github.com/digitalocean/doctl.git .
     override-build: |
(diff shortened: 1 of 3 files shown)


snap/snapcraft.yaml
```diff
@@ -29,6 +29,11 @@ parts:
     after: [launcher]
     source: .
     plugin: go
+    build-environment:
+    # Results in installation to /snap/doctl/current/opt/sandbox
+    - OVERRIDE_SANDBOX_DIR: "$SNAPCRAFT_PART_INSTALL/opt/sandbox"
+    # Override to prevent 'invalid cross-device link' calling doctl
+    - XDG_CONFIG_HOME: "$SNAPCRAFT_PART_INSTALL/tmp/.config"
     override-pull: |
       git clone https://github.com/digitalocean/doctl.git .
```
andrewsomething, line 38: Not that this builds using main. If you attempt to build this locally, you need to check out this branch to get the correct code.

ChiefMateStarbuck (approved):

PR #1177, opened 2022-05-31
Change 'actions' to 'functions' in externals where feasible

joshuaauerbachwatson:
This change substitutes the word `functions` for the word `actions` in several places commonly visited while using `doctl serverless`.

1.  The output of `doctl serverless functions list` has been redone using a `doctl` displayer so it no longer depends on the formatting done by the `aio` component of `nim`.
2.  Version 4.0 of `nim` is incorporated.   This brings several changes.
      a.  In deploy operations, a `functions` object is accepted in lieu of an `actions` object (but `actions` is still accepted for backward compatibility).
      b.  When `project.yml` is generated in `doctl serverless init`, the `functions` syntax is used rather than `actions`.   While this is not intended to be a breaking change (the deploy operation will accept the new syntax), it makes the public exposure of this change sensitive since App Platform must also be using the latest `nim` by the time a `doctl` release is made with the change.
      c.  The `doctl serverless get-metadata` command will return a JSON structure in which `functions` is now used in lieu of `actions`.   This _could_ be a breaking change for any tools the depend on parsing this JSON structure.   I don't know of any, but it is impossible to know if any have been coded by customers since `doctl serverless` was released.

This PR targets the feature branch, not main, since product documentation will probably need some attention when this change is released.

### DIFF
--- a/commands/displayers/functions.go
+++ b/commands/displayers/functions.go
@@ -1,0 +1,87 @@
+/*
+Copyright 2018 The Doctl Authors All rights reserved.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package displayers
+
+import (
+	"io"
+	"strings"
+	"time"
+
+	"github.com/digitalocean/doctl/do"
+)
+
+// Functions is the type of the displayer for functions list
+type Functions struct {
+	Info []do.FunctionInfo
+}
+
+var _ Displayable = &Functions{}
+
+// JSON is the displayer JSON method specialized for functions list
+func (i *Functions) JSON(out io.Writer) error {
+	return writeJSON(i.Info, out)
+}
+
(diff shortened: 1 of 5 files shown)


andrewsomething (commented):
Moving to doctl displayers where possible is nice as it will feel more native. Command that use displayers can also take advantage of some shared bits for a `--format` flag. This allows you to specify only certain fields to display as output. 

This:

```diff
diff --git a/commands/functions.go b/commands/functions.go
index a281c2f6..a684e08d 100644
--- a/commands/functions.go
+++ b/commands/functions.go
@@ -61,7 +61,7 @@ to the cloud.  You can provide inputs and inspect outputs.`,
        list := CmdBuilder(cmd, RunFunctionsList, "list [<packageName>]", "Lists all the functions",
                `Use `+"`"+`doctl sandbox functions list`+"`"+` to list the functions of your sandbox that are deployed
 to the cloud.`,
-               Writer)
+               Writer, displayerType(&displayers.Functions{}))
        AddStringFlag(list, "limit", "l", "", "only return LIMIT number of functions (default 30, max 200)")
        AddStringFlag(list, "skip", "s", "", "exclude the first SKIP number of functions from the result")
        AddBoolFlag(list, "count", "", false, "show only the total number of functions")
```

allows for this:

```
$ ./builds/doctl sandbox functions list --format Function,Version
Function Name    Latest Version
sample/hello     0.0.1
```

it also adds a `no-header` flag that is very useful in scripting:

```
$ FUNC_VERSION=$(./builds/doctl sandbox functions list --format Version --no-header)
$ echo $FUNC_VERSION
0.0.1
```

Might want to make `count` and `format` mutually exclusive. 

:+1:  Everything else looks good

PR #1178, opened 2022-06-01
Rename sandbox to serverless

joshuaauerbachwatson:
The `doctl` command for developing and testing functions has historically been called `sandbox` but has the aliases `serverless` and `sls` as well as `sbx`.

This change inverts the primary command and full-word alias so that the command is now called `serverless` and `sandbox` becomes an alias.   The shorter aliases `sls` and `sbx` are retained.

The change also adjusts all of the help texts and error texts (all texts made visible at runtime) to avoid the term "sandbox" altogether.    This was not accompanied by a wholesale editing of purely internal uses (e.g. function names, variable names, comments): the term sandbox is still used there, and also for the directory name of the "hidden" area where the software is installed.

By avoiding the word "sandbox" certain concepts had to be given alternative names.   The visible texts now use the "functions namespace" for what used to be called the cloud portion of the sandbox, and "functions project" for what used to be called the local disk areas of the sandbox.
 
This change targets the feature branch since it will require updates of product docs at release time.

### DIFF
--- a/commands/activations.go
+++ b/commands/activations.go
@@ -24,16 +24,16 @@ func Activations() *Command {
 		Command: &cobra.Command{
 			Use:   "activations",
 			Short: "Work with activation records",
-			Long: `The subcommands of ` + "`" + `doctl sandbox activations` + "`" + ` will list or retrieve results, logs, or complete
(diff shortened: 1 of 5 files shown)


andrewsomething (approved):
:rocket:

Issue #1176, opened 2022-05-31
doctl auth init flag --access-token do not have effect


aleixisp:
**flag -t or --access-token not overwriting token in config.yaml**

See images of the issue.
CLI does not use the token passed and always grab the one in the config.yaml
![api-portal](https://user-images.githubusercontent.com/54447877/171148293-807d26e2-b5e2-4565-9e2d-188acb3f4150.png)
![terminal](https://user-images.githubusercontent.com/54447877/171148300-d8b172c7-a99c-4155-aaba-7f3db4c57fd4.png)

**Environment:**

 - doctl version: doctl version 1.76.2-release
 - OS: macOS Monterey (although I think my ubuntu machines are affected too // 20.04.1 LTS (GNU/Linux 5.4.0-74-generic x86_64))
 - Installation method: Homebrew (macOs, docker Ubuntu machines)

andrewsomething:
Hi @aleixisp,

Thanks for raising this issue for us. We are unable to reproduce it using the defaults, but are noticing some unexpected behavior when there are multiple auth contexts. Do you have `doctl` configure with multiple context? Is the `DIGITALOCEAN_CONTEXT` variable set in your environment by any chance?

tristan-zander:
I'm also seeing this issue in both my CircleCI runner and my local Linux install, both running 1.82.0. When I removed the context, it respected the argument passed by `-t`.

`DIGITALOCEAN_CONTEXT` was not set in my environment, but `--context` was set.

tristan-zander:
What's even worse is that it doesn't respect the fact that it's in a non-interactive environment at all with the context set. https://app.circleci.com/pipelines/github/tristan-zander/runback-discord-bot/25/workflows/53994616-9fc3-4c7c-a12f-9002b14a8533/jobs/26/parallel-runs/0/steps/0-106

aleixisp:
Hey, @andrewsomething I did had two contexts, yes. Is it mandatory to pass the `--context` flag alongside with the auth token every-time?.

Rheisen:
I also have seen this issue when running the following:
```
doctl auth init --context projectname -c ~/Library/Application\ Support/doctl/config.yaml --access-token=$(op item get digitalocean_token --fields label=password)
```

It does not use the given access token, instead it goes to the interactive prompt. This is quite frustrating from a scripting perspective.

Reintjuu:
Is there a workaround available? I'd like to use contexts when running this non-interactive.

[...]

PR #1179, opened 2022-06-02
auth: Add integration tests for --access-token flag usage during init.

andrewsomething:
Test cases for https://github.com/digitalocean/doctl/issues/1176


scotchneat (approved):
Looks good. 
It would be good to add a case that covers using `auth init -t <token>` when a context is set in the env var to address the issue in #1176.

andrewsomething:
I'm going to go ahead and merge this. It's useful as is, and it's looking like we need to make some changes to how context work under the hood before adding tests for those cases.

PR #1183, opened 2022-06-22
Set a distinct user agent header for doctl serverless

joshuaauerbachwatson:
Previously, when `doctl serverless` drove `nim` via the sandbox plugin, requests arriving at the serverless cluster endpoint could not be distinguished from ones generated by direct uses of `nim`.   This change ensures that requests originating via `doctl sls` will present a distinct and informative user agent header.

Version 4.1.0 of `nim` and version 1.3 of the sandbox plugin are needed to cooperate in this change, so changing to those versions is part of this change.

Note that `doctl` also sets a user header for use via `godo`.    Requests with that user agent header go to the main DigitalOcean API endpoint.   I don't believe there is a conflict here because the endpoints are different and the overall traffic to them is different and their needs to discriminate between different sources of requests are different.


andrewsomething (approved):
:+1:  LGTM!

PR #1185, opened 2022-06-28
account: Display team name.

andrewsomething:
Now that the API returns team info, we can can display it in the output here. Additionally I've updated the existing headers to clarify that the email and UUID displayed are for the user.  I've retained the existing arguments to the `--format` flag for them to prevent breaking any existing scripting. e.g. `doctl account get --format Email,UUID`


mikeberezowski (approved):
This looks good to me. 👍

scotchneat (approved):
👍🏽

PR #1187, opened 2022-06-30
snap: Update build env to use go 1.18

andrewsomething:
When cutting the last release, I happened to notice that the Snap builds were still on Go 1.17. Of course things did not end up as straight forward as simply bumping the version string. Rebuilding the image using the Dockerfile in currently in main now creates an unusable image:

```
$ docker run sammytheshark/doctl-snap-base:latest
Traceback (most recent call last):
  File "/snap/snapcraft/current/bin/snapcraft", line 5, in <module>
    from snapcraft.cli import run
ModuleNotFoundError: No module named 'snapcraft'
```

So I've brought in a fix from the upstream Dockerfile this is based on. I also brought in a few small improvements that make the image smaller and the layers more cachable.

### DIFF
--- a/dockerfiles/Dockerfile.snap
+++ b/dockerfiles/Dockerfile.snap
@@ -4,7 +4,7 @@
 #
 # For help with the technical aspects of this Dockerfile, see
 #   https://snapcraft.io/docs/t/creating-docker-images-for-snapcraft/11739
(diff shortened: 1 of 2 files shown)


scotchneat (approved):
LTGM

PR #1188, opened 2022-07-05
Refactor sandbox service for better encapsulation

joshuaauerbachwatson:
The `SandboxService` is our abstraction of the serverless capabilities for mocking and testing purposes.  

The original version of this abstraction left out status checking and installation, leaving some serverless code paths to be abstracted directly in `CmdConfig`, outside the abstraction boundary.   We worked around this deviation for testing purposes, but it creates a conflict going forward because the service itself only partly separates serverless concerns from other concerns. 

Specifically, we are about to temporarily complicate the service, giving it a direct path to the functions backend cluster, as part of an effort to reduce dependence on the plugin for serverless capabilities (ultimately, it should get simplified again as serverless APIs are regularized, but this will take time).

I did not want to continue the practice of complicating `CmdConfig` as the serverless path goes through a series of non-trivial changes.   So, this change attempts to move installation and status checking from the `commands` package to the `do` package where they become part of `SandboxService`.   

There is a fair amount of code relocation in this PR but it is conceptually simple.   I did make an orthogonal split of `commands/sandbox.go` to put its utility functions (those that were not relocated to `do`) in a separate source file.  I did not make the corresponding split in `do/sandbox.go`.

Small warts are left: `do` cannot depend on `commands` because that creates a circularity, but `commands` is where the current access token is kept as well as the conventions for finding the config directory.   This is dealt with in the initialization of the service and in the signatures of some functions.

In this change I dropped support for the `OVERRIDE_SANDBOX_DIR` environment variable because it unduly complicated the coding and has never actually been used in testing.  The installation and upgrade tests in `doctl-sandbox-plugin` run under a github action in a clean container.   Even for manual testing during development, since the sandbox area contains no mutable state other than the latest connected namespace, it is no problem in practice to delete and recreate it.

### DIFF
--- a/commands/sandbox_util.go
+++ b/commands/sandbox_util.go
@@ -26,21 +26,9 @@ import (
 )
 
 const (
-	// Minimum required version of the sandbox plugin code.  The first part is
-	// the version of the incorporated Nimbella CLI and the second part is the
-	// version of the bridge code in the sandbox plugin repository.
-	minSandboxVersion = "4.1.0-1.3.0"
-
-	// The version of nodejs to download alongsize the plugin download.
-	nodeVersion = "v16.13.0"
-
 	// noCapture is the string constant recognized by the plugin.  It suppresses output
 	// capture when in the initial (command) position.
 	noCapture = "nocapture"
-
-	// credsDir is the directory under the sandbox where all credentials are stored.
-	// It in turn has a subdirectory for each access token employed (formed as a prefix of the token).
-	credsDir = "creds"
 )
 
 // SandboxExec executes a sandbox command
--- a/commands/sandbox_util.go
+++ b/commands/sandbox_util.go
@@ -1,0 +1,183 @@
+/*
+Copyright 2018 The Doctl Authors All rights reserved.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package commands
+
+import (
+	"crypto/sha1"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/digitalocean/doctl/do"
+)
+
+const (
+	// Minimum required version of the sandbox plugin code.  The first part is
+	// the version of the incorporated Nimbella CLI and the second part is the
+	// version of the bridge code in the sandbox plugin repository.
+	minSandboxVersion = "4.1.0-1.3.0"
+
+	// The version of nodejs to download alongsize the plugin download.
+	nodeVersion = "v16.13.0"
+
+	// noCapture is the string constant recognized by the plugin.  It suppresses output
+	// capture when in the initial (command) position.
+	noCapture = "nocapture"
+
+	// credsDir is the directory under the sandbox where all credentials are stored.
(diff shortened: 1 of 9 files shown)


joshuaauerbachwatson:
> @joshuaauerbachwatson One thing to note here is that we are using this in order to support the Snap packages.

Thanks.   And oops.   I'll restore the capability.

joshuaauerbachwatson:
> > @joshuaauerbachwatson One thing to note here is that we are using this in order to support the Snap packages.
> 
> Thanks. And oops. I'll restore the capability.

Done.   I will re-run e2e tests now.

joshuaauerbachwatson:
> > Thanks. And oops. I'll restore the capability.
> 
> Done. I will re-run e2e tests now.

Tests passed: https://github.com/digitalocean/doctl-sandbox-plugin/actions/runs/2617558110

But those tests use the usual logic for installing `doctl` in a container, which is not the snap install.

commands/sandbox_util.go
```diff
@@ -0,0 +1,183 @@
...
+	"strings"
+
+	"github.com/digitalocean/doctl/do"
+)
+
+const (
+	// Minimum required version of the sandbox plugin code.  The first part is
+	// the version of the incorporated Nimbella CLI and the second part is the
+	// version of the bridge code in the sandbox plugin repository.
+	minSandboxVersion = "4.1.0-1.3.0"
```
andrewsomething, line 32: I don't love having to maintain this in two different places. Looks like there is only one instance of this const being used in the commands package. Could it be replaced with `do.GetMinSandboxVersion()`?

commands/sandbox_util.go
```diff
@@ -0,0 +1,183 @@
...
+
+	// The version of nodejs to download alongsize the plugin download.
+	nodeVersion = "v16.13.0"
```
andrewsomething, line 35: Doesn't look like the `nodeVersion` const from the commands package is being used.

```
$ grep -r nodeVersion .
./commands/sandbox_util.go:     nodeVersion = "v16.13.0"
./do/sandbox.go:        nodeVersion = "v16.13.0"
./do/sandbox.go:                nodeDir = fmt.Sprintf("node-%s-%s-%s", nodeVersion, goos, arch)
./do/sandbox.go:                nodeURL = fmt.Sprintf("https://nodejs.org/dist/%s/%s.tar.gz", nodeVersion, nodeDir)
./do/sandbox.go:                        nodeURL = fmt.Sprintf("https://nodejs.org/dist/%s/%s.zip", nodeVersion, nodeDir)
./do/sandbox.go:                return installed == nodeVersion
```

commands/sandbox_util.go
```diff
@@ -0,0 +1,183 @@
...
+
+	// noCapture is the string constant recognized by the plugin.  It suppresses output
+	// capture when in the initial (command) position.
+	noCapture = "nocapture"
+
+	// credsDir is the directory under the sandbox where all credentials are stored.
+	// It in turn has a subdirectory for each access token employed (formed as a prefix of the token).
+	credsDir = "creds"
```
andrewsomething, line 43: Doesn't look like the `credsDir` const from the commands package is being used either.

andrewsomething (commented):
I left a couple small comments, but overall this looks great. I also validated that the snap build continues to work as expected.

joshuaauerbachwatson:
Review comments are right on.   I believe I cloned the `const` declarations as I started the refactoring and meant to trim the result later but forgot.  I'll make sure every constant is defined only once and will investigate any that are seemingly unused.

commands/sandbox_util.go
joshuaauerbachwatson: Agreed.   Fixed.

commands/sandbox_util.go
joshuaauerbachwatson: Right.  Fixed.

commands/sandbox_util.go
joshuaauerbachwatson: Yes.   Fixed.

PR #1189, opened 2022-07-08
auth remove example

dbrian57:
Received a few pieces of feedback about how the `auth remove` command could be clearer. I've added an example of the command's use here.

### DIFF
--- a/commands/auth.go
+++ b/commands/auth.go
@@ -99,7 +99,7 @@ To see a list of available authentication contexts, call `+"`"+`doctl auth list`
 For details on creating an authentication context, see the help for `+"`"+`doctl auth init`+"`"+`.`, Writer, false)
 	cmdAuthSwitch.AddValidArgsFunc(authContextListValidArgsFunc)
 
-	cmdAuthRemove := cmdBuilderWithInit(cmd, RunAuthRemove, "remove", "Remove authentication contexts ", `This command allows you to remove authentication contexts you've already created.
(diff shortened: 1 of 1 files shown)


commands/auth.go
```diff
@@ -99,7 +99,7 @@ To see a list of available authentication contexts, call `+"`"+`doctl auth list`
 For details on creating an authentication context, see the help for `+"`"+`doctl auth init`+"`"+`.`, Writer, false)
 	cmdAuthSwitch.AddValidArgsFunc(authContextListValidArgsFunc)
 
-	cmdAuthRemove := cmdBuilderWithInit(cmd, RunAuthRemove, "remove", "Remove authentication contexts ", `This command allows you to remove authentication contexts you've already created.
+	cmdAuthRemove := cmdBuilderWithInit(cmd, RunAuthRemove, "remove", "Remove authentication contexts ", `This command allows you to remove authentication contexts you've already created. For example, to remove a context called `example-context` run ` + "`" + `doctl auth switch --context <name>` + "`" + `.
```
andrewsomething, line 102:
```suggestion
	cmdAuthRemove := cmdBuilderWithInit(cmd, RunAuthRemove, "remove --context <name>", "Remove authentication contexts ", `This command allows you to remove authentication contexts you've already created.
```

If you add it here, it will show up in the usage section of the help.

andrewsomething (commented):
The error message comes from line 167. We could probably improve that by mentioning the flag while we're here.

commands/auth.go
dbrian57: @andrewsomething Added! Thank you.

hazel-nut (approved):
:shipit:

andrewsomething (approved):

PR #1194, opened 2022-07-27
Remove duplicate suggestions

### LINKED ISSUE #1192: Getting Duplicate Suggestions for invalid commands
**Describe the Issue:**

Currently if a user use a invalid command ```doctl apa``` , we are getting duplicate suggestions for that commands . 
image

**Environment:**

 - doctl version: 1.78.0-release
 - OS: windows/amd64
 - Installation method: Source 

**Additional Details:**

ChiefMateStarbuck:
fixes #1192 

Word suggestions are being printed twice. The line to that removes the duplication (so far) does not affect other error print statements.

### DIFF
--- a/commands/doit.go
+++ b/commands/doit.go
@@ -127,7 +127,6 @@ func configHome() string {
 // Execute executes the current command using DoitCmd.
 func Execute() {
 	if err := DoitCmd.Execute(); err != nil {
-		fmt.Println(err)
 		os.Exit(-1)
 	}
 }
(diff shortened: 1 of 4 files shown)


commands/doit.go
```diff
@@ -127,7 +127,6 @@ func configHome() string {
 // Execute executes the current command using DoitCmd.
 func Execute() {
 	if err := DoitCmd.Execute(); err != nil {
-		fmt.Println(err)
 		os.Exit(-1)
```
scotchneat, line 130: I'm concerned this would hide errors we would want to return to the user.
What if we did something like this instead:

```suggestion
		if !strings.Contains(err.Error(), "unknown command") {
			fmt.Println(err)
		}
		os.Exit(-1)
```

andrewsomething (commented):
Would be nice to get an integration test for this as well in case something changes in Cobra.

scotchneat (approved):

PR #1198, opened 2022-08-10
Add full type declarations for projects and reader support

programmersstudio:
The aim of the PR is to reduce the dependency on the deployer by moving the project reader/writer service here. The deployment part of the project is still being handled by the deployer.

This PR only contains the project reader service.


joshuaauerbachwatson (changes requested):
Nice, overall, @programmersstudio.   Some required changes and a few comments that don't require action (hopefully clear which are which).   Also, please fill in the description saying briefly what this PR does.

Note to the `doctl` team: this code will be dormant for a while, but it is useful to us to have it merged and in the codebase going forward because many other planned changes will build on it.

commands/serverless-extra.go
```diff
@@ -151,7 +155,17 @@ func RunServerlessExtraGetMetadata(c *CmdConfig) error {
 	if err != nil {
 		return err
 	}
-	output, err := RunServerlessExec(projectGetMetadata, c, []string{flagJSON}, []string{flagEnv, flagInclude, flagExclude})
+	r, _ := c.Doit.GetBool(c.NS, flagProjectReader)
+
+	var output do.ServerlessOutput
+	project := do.ServerlessProject{
+		ProjectPath: args[0],
```
joshuaauerbachwatson, line 162: Sorry.   My previous suggestion was slightly broken.   That's the problem with the github suggestion thing.  It encourages the  reviewer to propose (and the PR author to accept) untested code.

```suggestion
		ProjectPath: c.Args[0],
```

commands/serverless-extra.go
```diff
@@ -151,7 +155,17 @@ func RunServerlessExtraGetMetadata(c *CmdConfig) error {
...
+	}
+	if r {
+		output, err = c.Serverless().ReadProject(&project, args)
```
joshuaauerbachwatson, line 165:
```suggestion
		output, err = c.Serverless().ReadProject(&project, c.Args)
```

joshuaauerbachwatson (changes requested):
Sorry.   A previous suggestion of mine was broken.   I suggested a fix.   This time I tested it :-).   So, we need one more tweak.

commands/serverless-extra.go
programmersstudio: Not a problem, I noticed that and fixed and tested it in the local system.

joshuaauerbachwatson (approved):
Looks good to me now.   Thanks, @programmersstudio .

 I'm going to wait for an approval from a member of the doctl team before merging.

We are going to do a feature merge and public release toward the end of the month and this code will be there (dormant).

commands/serverless-extra.go
```diff
@@ -59,6 +61,8 @@ project (a directory you have designated for functions development).  This can b
 	AddStringFlag(getMetadata, "env", "", "", "Path to environment file")
 	AddStringFlag(getMetadata, "include", "", "", "Functions or packages to include")
 	AddStringFlag(getMetadata, "exclude", "", "", "Functions or packages to exclude")
+	AddBoolFlag(getMetadata, "project-reader", "", false, "Test new project reader service")
+	getMetadata.Flags().MarkHidden("project-reader")
```
andrewsomething, line 65: I'm curious about the purpose of this flag. Is only intended for internal testing, or will we expose it to customers? Think I'm just missing some context.

joshuaauerbachwatson: It will never be exposed to customers.    The plan is that `get-metadata` will _always_ using the new project-reader service once we've gained enough confidence to switch over.   The flag is "for testing only" while we're not yet ready to switch, and will help us gain the confidence by allowing us to compare behavior with and without the flag.   So the flag will eventually be removed without ever being exposed.

do/serverless.go
```diff
@@ -589,6 +680,93 @@ func (s *serverlessService) GetConnectedAPIHost() (string, error) {
...
+func (s *serverlessService) ReadProject(project *ServerlessProject, args []string) (ServerlessOutput, error) {
+	err := readTopLevel(project)
+	if err != nil {
+		return ServerlessOutput{}, err
+	}
+	_, err = readProjectConfig(project.ConfigPath)
+	if err != nil {
+		return ServerlessOutput{}, err
+	}
+	return ServerlessOutput{}, nil
```
scotchneat, line 695: Even though this code path is hidden, since we're not doing anything with the `readProjectConfig` spec output here, it might make sense to return an error message like this until it's fully implemented (unless this is intended behavior):

```suggestion
	_, err = readProjectConfig(project.ConfigPath)
	if err != nil {
		return ServerlessOutput{}, err
	}
	return ServerlessOutput{}, fmt.Errorf("not implemented")
```

scotchneat (commented):
Looks good overall. I added a couple of comments inline.

PR #1201, opened 2022-08-11
Filter AP namespaces from all ListNamespaces responses

joshuaauerbachwatson:
Properly speaking, the ListNamespaces API should return only self-managed namespaces, not AP-managed namespaces.   latter should not be modified except via AP.   Currently, the API does appear to return AP-managed namespaces.   As a short-term fix, this change does this filtering in doctl code.

### DIFF
--- a/do/serverless.go
+++ b/do/serverless.go
@@ -493,7 +493,22 @@ func (s *serverlessService) ListNamespaces(ctx context.Context) (NamespaceListRe
 	if err != nil {
 		return NamespaceListResponse{}, err
 	}
-	return *decoded, nil
+	return removeAppNamespaces(*decoded), nil
+}
+
+// removeAppNamespaces modifies a NamespaceListResponse to exclude namespaces prefixed by ap-.
+// Those are supposed to be managed by App Platform and should not be available to doctl serverless
+// for connection or modification of any kind.  This is intended to be temporary because the filtering
+// really should be done within the API.
+func removeAppNamespaces(input NamespaceListResponse) NamespaceListResponse {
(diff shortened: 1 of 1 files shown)


do/serverless.go
```diff
@@ -493,7 +493,22 @@ func (s *serverlessService) ListNamespaces(ctx context.Context) (NamespaceListRe
...
-	return *decoded, nil
+	return removeAppNamespaces(*decoded), nil
+}
+
+// removeAppNamespaces modifies a NamespaceListResponse to exclude namespaces prefixed by ap-.
+// Those are supposed to be managed by App Platform and should not be available to doctl serverless
+// for connection or modification of any kind.  This is intended to be temporary because the filtering
+// really should be done within the API.
+func removeAppNamespaces(input NamespaceListResponse) NamespaceListResponse {
+	newList := []OutputNamespace{}
```
ddebarros, line 504: I would recommend using the `ns.Owner === 'APP'` instead of the prefix on the namespace.

joshuaauerbachwatson: Unfortunately, the ListNamespaces API does not return the Owner field.   I am neutral on whether it "should" but, since this fix is designed to be short term and we will have some requirements for changes to the API, I'd rather request that it filter out AP namespaces altogether instead of returning the Owner field so that doctl can do it.

ddebarros: Sounds good.

andrewsomething (approved):
:+1:  LGTM

ddebarros (approved):

PR #1205, opened 2022-08-17
Set user as optional field in calls to databases pool

DMW2151:
The DO API allows DB pools to be created without a user specified. This PR modifies the  `databases pool` options and removes user as a required field. Without user specified, pool uses the connection credentials of the inbound user.

example usage:

```bash
doctl databases pool create $DB_UUID p-04 --db defaultdb --size 2 -t $DIGITALOCEAN_TOKEN
```

```
User    Name    Size    Database     Mode           URI                                                                                                                           
        p-04    2       defaultdb    transaction    postgresql://:@${HOST}/p-04?sslmode=require
```

### DIFF
--- a/commands/databases.go
+++ b/commands/databases.go
@@ -809,7 +809,7 @@ Connection pools can be created and deleted with these commands, or you can simp
 
 	connectionPoolDetails := `
 
-- The username of the database user account that the connection pool uses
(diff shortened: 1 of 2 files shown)


andrewsomething (approved):
:+1:  LGTM

PR #1207, opened 2022-08-18
appdev: extract pager component and improve build log output

kamaln7:


commands/charm/pager.go
```diff
@@ -0,0 +1,217 @@
...
+		Foreground(lipgloss.Color("#dddddd")).
+		BorderStyle(b).
+		BorderForeground(lipgloss.Color("#9B9B9B")).
+		Padding(0, 1)
+}
+
+func (m *pagerModel) infoStyle() lipgloss.Style {
+	b := lipgloss.RoundedBorder()
+	b.Left = "┤"
+	return m.titleStyle().Copy().BorderStyle(b).Foreground(lipgloss.Color("#dddddd"))
```
nicktate, line 203: Is it worth adding this and 9b9b9b as color constants?

commands/charm/pager.go
```diff
@@ -0,0 +1,217 @@
...
+			//
+			// Render the viewport one line below the header.
+			m.viewport.YPosition = headerHeight + 1
+		} else {
+			m.viewport.Width = msg.Width
+			m.viewport.Height = msg.Height - verticalMarginHeight
+		}
+	case msgUpdate:
+		m.viewport.SetContent(m.buffer.String())
+		m.viewport.GotoBottom()
```
nicktate, line 157: Does this override mouse events? e.g. if I scroll up and new content comes in will it snap the view back to the bottom?

commands/apps_dev.go
```diff
@@ -269,215 +265,81 @@ func RunAppsDevBuild(c *CmdConfig) error {
...
+	if Interactive {
+		pager, err := charm.NewPager(
+			charm.PagerWithTitle("Building " + component),
+		)
+		if err != nil {
+			return fmt.Errorf("starting log pager: %w", err)
 		}
+		wg.Add(1)
+		go func() {
+			defer cancel()
```
nicktate, line 287: Why do we defer this cancel in multiple locations?

nicktate (commented):
This looks good overall, just a few questions inline

commands/charm/pager.go
kamaln7: I meant to do that 🤦🏼‍♂️  will address in the next PR

commands/charm/pager.go
kamaln7: Yes, that was intentional because right now there is no indication that new output was added if you're scrolled up. Ideally we'd either add a scrollbar or some badge at the bottom indicating the presence of unread logs

commands/apps_dev.go
kamaln7: Both the pager and the builder inherit the parent context:

```
	cli, err := c.Doit.GetContainerEngineClient()
	if err != nil {
		return err
	}

	ctx, cancel := context.WithCancel(ctx)
	defer cancel()
```

Once the builder exits we need to cancel the pager's context so that it can shut down cleanly. But we also want the builder to exit cleanly (remove containers etc) once the pager exits — this enables ctrl-c to work correctly while the pager is active. So we need to ensure that the context is canceled once either the pager or the builder exit, whichever comes first.

nicktate (approved):

PR #1208, opened 2022-08-18
Add support for an app dev config and management

nicktate:

* Adds `doctl app dev config set|unset` for local build configuration
* Adds `appDevConfig` struct which simplifies working across the `dev config` and the default `doctl falgs`

### DIFF
--- a/commands/apps_dev_config.go
+++ b/commands/apps_dev_config.go
@@ -1,0 +1,298 @@
+package commands
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strings"
+
+	"github.com/digitalocean/doctl"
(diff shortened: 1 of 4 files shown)


commands/apps_dev_config.go
```diff
@@ -0,0 +1,272 @@
...
+	} else if err != nil {
+		return err
+	}
+
+	b, err := ioutil.ReadFile(file)
+	if err != nil {
+		return err
+	}
+
+	if exists, err := regexp.Match(val, b); err != nil {
```
kamaln7, line 235:
```suggestion
	if exists, err := regexp.Match(regexp.QuoteMeta(val), b); err != nil {
```

kamaln7 (approved):

kamaln7 (approved):
LGTM; one note - let's use `filepath.Join` for windows compat

kamaln7 (approved):

PR #1214, opened 2022-08-22
Add snap support for docker interface for app dev builds

nicktate:

Adds `app-dev-build` plug w/ the `docker` interface

> This will require the user to have docker installed via snap and execute `snap connect doctl:app-dev-build docker:docker-daemon` to grant permissions

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -64,6 +64,8 @@ parts:
       bin/doctl: bin/doctl.real
 
 plugs:
+  app-dev-build:
+    interface: docker
   dot-docker:
     interface: personal-files
     write:
@@ -88,3 +90,4 @@ apps:
       - dot-docker
       - doctl-config
       - kube-config
+      - app-dev-build
--- a/commands/apps_dev.go
+++ b/commands/apps_dev.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"sync"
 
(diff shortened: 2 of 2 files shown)


snap/snapcraft.yaml
```diff
@@ -35,7 +35,8 @@ parts:
     # Override to prevent 'invalid cross-device link' calling doctl
     - XDG_CONFIG_HOME: "$SNAPCRAFT_PART_INSTALL/tmp/.config"
     override-pull: |
-      git clone https://github.com/digitalocean/doctl.git .
+      git clone --branch ntate/appdev/snapcraft https://github.com/nicktate/doctl.git .
```
andrewsomething, line 38: Just flagging that this will need to be removed before merging to main.

andrewsomething (approved):
:+1:  LGTM!

kamaln7 (approved):

andrewsomething:
Also wanted to point out that we do have a few places where we do some custom error messages to point snap users in the right direction. E.g.

https://github.com/digitalocean/doctl/blob/51c4f6b63a5f6dd7c29529c1df4e06350d592779/commands/registry.go#L425-L429

Might want to consider doing that for app dev builds as well.

nicktate:
> Also wanted to point out that we do have a few places where we do some custom error messages to point snap users in the right direction. E.g.
> 
> https://github.com/digitalocean/doctl/blob/51c4f6b63a5f6dd7c29529c1df4e06350d592779/commands/registry.go#L425-L429
> 
> Might want to consider doing that for app dev builds as well.

💯 @kamaln7 and I were just talking about this as well. I will add it in, thanks for the reference

PR #1219, opened 2022-08-24
appdev: config updates: add hierarchy and per-component overrides, updated env-file handling

kamaln7:
@andrewsomething - please review the `doit.go` changes.

### DIFF
--- a/commands/apps_dev.go
+++ b/commands/apps_dev.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"io/fs"
 	"os"
+	"path/filepath"
 	"sync"
 
 	"github.com/MakeNowJust/heredoc"
@@ -22,13 +23,10 @@ import (
 	"github.com/digitalocean/doctl/internal/apps/workspace"
 
 	"github.com/digitalocean/godo"
-	"github.com/joho/godotenv"
 	"github.com/spf13/cobra"
 )
 
 const (
-	// AppsDevDefaultSpecPath is the default spec path for an app.
-	AppsDevDefaultSpecPath = ".do/app.yaml"
 	// AppsDevDefaultEnvFile is the default env file path.
 	AppsDevDefaultEnvFile = ".env"
 )
@@ -86,7 +84,7 @@ func AppsDev() *Command {
 	)
 
 	AddStringFlag(
-		build, doctl.ArgAppDevBuildCommand,
+		build, doctl.ArgBuildCommand,
 		"", "",
 		"Optional build command override for local development.",
 	)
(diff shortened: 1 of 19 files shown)


commands/apps_dev.go
```diff
@@ -106,33 +109,36 @@ func AppsDev() *Command {
...
+
+	conf, err := newAppDevConfig(c)
 	if err != nil {
-		return err
+		return fmt.Errorf("initializing config: %w", err)
 	}
+
+	var (
+		// doctlConfig is the CLI config source.
+		doctlConfig = config.DoctlConfigSource(c.Doit, c.NS)
```
nicktate, line 120: Do you think it is worth wrapping this logic inside `newAppDevConfig` and creating convenience methods to access the corresponding config types? I just worry this is a lot of non-obvious boilerplate each run command has to manage.

internal/apps/config/config.go
```diff
@@ -0,0 +1,116 @@
...
+	IsSet(key string) bool
+	GetString(key string) string
+	GetBool(key string) bool
+	GetDuration(key string) time.Duration
+}
+
+type mutatingConfigSource struct {
+	cs          ConfigSource
+	mutateKey   func(key string) string
+	mutateIsSet bool // some config sources except IsSet to _not_ include the namespace
```
nicktate, line 19: I think this is a bit confusing to have differing access patterns based on this flag; I'd vote we always use the `mutateKey` function if it exists for a mutating source

internal/apps/config/appdev.go
```diff
@@ -0,0 +1,223 @@
...
+
+// type appDevUnknownKeyErr struct {
+// 	key string
+// }
+
+// func (e *appDevUnknownKeyErr) Error() string {
+// 	return fmt.Sprintf("unknown key: %s\nvalid keys: %s", e.key, ValidAppDevKeys())
+// }
+
+// var validAppDevKeys = map[string]bool{
```
nicktate, line 32: Are we planning on bringing this concept back? If not we should remove the commented code.

internal/apps/config/appdev.go
```diff
@@ -0,0 +1,223 @@
...
+func (c *AppDev) ClearCacheDir(ctx context.Context, component string) error {
+	return os.RemoveAll(c.ContextPath(".do", "cache", component))
+}
+
+func (c *AppDev) WriteConfig() error {
+	return c.viper.WriteConfig()
+}
+
+func (c *AppDev) Set(key string, value any) error {
+	// if !validAppDevKeys[key] {
```
nicktate, line 77: nit: lets remove if no longer needed

commands/apps_dev.go
```diff
@@ -241,29 +235,49 @@ func RunAppsDevBuild(c *CmdConfig) error {
...
+		// componentConfig is the per-component config source.
+		componentConfig = appsDevFlagConfigCompat(conf.Components(component))
+		// componentGlobalConfig is per-component config that can be overridden at the global/app level as well as via cli flags.
+		componentGlobalConfig = config.Multi(
+			doctlConfig,
+			componentConfig,
+			appsDevFlagConfigCompat(conf),
+		)
+		// componentArgConfig is per-component config that can be overridden via cli flags.
+		componentArgConfig = config.Multi(
```
nicktate, line 248: why do we need this? Shouldn't this be covered by the `componentGlobalConfig` since first match would win?

commands/apps_dev.go
```diff
@@ -274,19 +288,19 @@ func RunAppsDevBuild(c *CmdConfig) error {
...
-			confirm.WithPersistPrompt(confirm.PersistPromptIfNo),
-		).Prompt()
-		if err != nil {
-			return err
-		}
-		if choice != confirm.Yes {
-			return fmt.Errorf("cancelled")
-		}
-	}
+	// if Interactive {
```
nicktate, line 291: uncomment?

nicktate (commented):
I like this a lot 🙇🏻 . Left a few comments inline but my main questions are:

1. Can we keep mutatingConfig semantics the same between `IsSet` and other `Getters`?
2. Can we reduce command runner boilerplate by providing the standard multi-configs in the `appDevConfig` struct?

internal/apps/config/appdev.go
kamaln7: I'm not sure, I left it in so we can decide without unblocking these changes

commands/apps_dev.go
kamaln7: commented out temporarily until we add more useful output prior or "do we need to ask for confirmation" logic

commands/apps_dev.go
kamaln7: `componentGlobalConfig` would allow for an app-wide `build_command` in `dev-config.yaml` which wouldn't make much sense

commands/apps_dev.go
kamaln7: Yep that would be nicer 👍🏼

commands/apps_dev.go
nicktate: Would it really allow for it? We enforce setting through the `app dev config set` so it would just simplify the accessor pattern in my mind.

commands/apps_dev.go
```diff
@@ -106,33 +109,25 @@ func AppsDev() *Command {
...
-	}
-	timeout, err := c.Doit.GetDuration(c.NS, doctl.ArgTimeout)
+
+	ws, err := appDevWorkspace(c)
 	if err != nil {
-		return err
+		return fmt.Errorf("preparing workspace: %w", err)
 	}
+
+	globalConfig := ws.Config.Global(true)
```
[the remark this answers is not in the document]
kamaln7: Not sure why this showed up, it's from the old diff, but yes I've renamed all "global" refs to "workspace" 👍🏼

internal/apps/config/doctl_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+		assert.Equal(t, true, cs.GetBool("interactive"))
+		assert.Equal(t, time.Minute, cs.GetDuration("last-seen-ago.bufo"))
+		assert.Equal(t, time.Duration(0), cs.GetDuration("bufo")) // does not exist
+	})
+
+	t.Run("yes namespace", func(t *testing.T) {
+		cs := DoctlConfigSource(doctlConfig, "last-seen-ago")
+		assert.True(t, cs.IsSet("interactive"))
+		assert.False(t, cs.IsSet("last-seen-ago.bufo")) // namespace not considered by IsSet
```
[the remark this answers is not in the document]
kamaln7: `IsSet` is still excluded from key mutation:

```go
// DoctlConfigSource converts a doctl.Config into a ConfigSource with an optional default namespace.
func DoctlConfigSource(config doctl.Config, ns string) ConfigSource {
	...

	// doctl expects the namespace to be present on all calls except IsSet.
	return MutatingConfigSource(&doctlConfigSource{config}, mutateKey, []string{"IsSet"}) // <----
}
```

doctl's IsSet implementation doesn't accept namespaces. [The LiveConfig implementation](https://github.com/digitalocean/doctl/blob/5408a2bb2fe7f8a97134074411754b248d967171/doit.go#L268-L288) basically loops through os.Args to find any references to the given key. For example, `doctl apps dev build --build-command` on the CLI maps to `cmdConfig.Doit.GetString("apps.dev.build", "build-command")`. Because IsSet searches the cli args, it wouldn't find a match for `apps.dev.build.build-command` but rather only `build-command`. 

[TestConfig behaves similarly](https://github.com/digitalocean/doctl/blob/5408a2bb2fe7f8a97134074411754b248d967171/doit.go#L456-L466).

PR #1223, opened 2022-09-06
Unhide the multiple namespaces feature

joshuaauerbachwatson:
This change unhides the support for multiple namespaces.

It should probably not be merged until feature flippers are opened to make the capability available to all (it is still in beta as of now).

The change also fixes a bug in error returns from serverless functions get.


andrewsomething (approved):
:+1:  Looks good to merge pending feature rollout

PR #1224, opened 2022-09-06
serverless: work-a-round for installing in snap build.

andrewsomething:
The snap builds have failed after the serverless refactor as the command now requires doctl to be configured with an API token. 

https://github.com/digitalocean/doctl/runs/8207250662?check_suite_focus=true#step:4:610

This change allows us to by pass that requirement if `SNAP_SANDBOX_INSTALL` is  set.

### DIFF
--- a/commands/serverless.go
+++ b/commands/serverless.go
@@ -68,10 +68,10 @@ Other ` + "`" + `doctl serverless` + "`" + ` commands are used to develop and te
 		},
 	}
 
-	CmdBuilder(cmd, RunServerlessInstall, "install", "Installs the serverless support",
+	cmdBuilderWithInit(cmd, RunServerlessInstall, "install", "Installs the serverless support",
 		`This command installs additional software under `+"`"+`doctl`+"`"+` needed to make the other serverless commands work.
 The install operation is long-running, and a network connection is required.`,
-		Writer)
+		Writer, false)
 
 	CmdBuilder(cmd, RunServerlessUpgrade, "upgrade", "Upgrades serverless support to match this version of doctl",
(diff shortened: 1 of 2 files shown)


commands/serverless.go
```diff
@@ -114,9 +114,28 @@ the entire packages are removed.`, Writer)
...
+		status       error
+	)
+
+	// When building the snap package, we need to install the serverless plugin
+	// without a fully configured and authenticated doctl. So we only fully init
+	// the service if SNAP_SANDBOX_INSTALL is not set.
+	_, isSnapInstall := os.LookupEnv("SNAP_SANDBOX_INSTALL")
+	if isSnapInstall {
+		serverlessDir := os.Getenv("OVERRIDE_SANDBOX_DIR")
+		serverless = do.NewServerlessService(nil, serverlessDir, "")
```
joshuaauerbachwatson, line 129: I believe the NewServerlessService function will interrogate OVERRIDE_SANDBOX_DIR itself and ignore the argument, so perhaps it is unnecessary to do it twice.   Just a nit.  it should be harmless.

joshuaauerbachwatson (approved):
LGTM.   One comment, which is a nit.  I need to leave it to you to test whether the snap build is really fixed (not sure how to do that) but I will verify that ordinary installs are not regressed.

ChiefMateStarbuck (dismissed):

joshuaauerbachwatson:
Verified that ordinary installs are working fine after the change (no surprise ... just thought I'd check 😄 ).

joshuaauerbachwatson:
@andrewsomething thank you for being on top of this so quickly and sorry for the disruption.

ChiefMateStarbuck (approved):

PR #1225, opened 2022-09-07
Add support for typescript in the nodejs runtime

joshuaauerbachwatson:
The set of supported languages in DigitalOcean Functions actually includes TypeScript.   TypeScript does not require its own runtime but is supported by setting up the function project correctly to run in one of the `nodejs` runtimes.

This change (1) corrects the "languages" display (which is actually organized by runtime) to stop assuming there will eventually be a TypeScript runtime but rather correctly indicating that TypeScript is supported in the `nodejs` runtime.

(2) It also incorporates a later version of the serverless plugin which knows how to properly set up a functions project for Typescript and correctly associates the `.ts` file suffix with the `nodejs` runtime.

### DIFF
--- a/commands/serverless.go
+++ b/commands/serverless.go
@@ -39,17 +39,16 @@ var (
 	// Note: this table has all languages for which we possess samples.  Only those with currently
 	// active runtimes will display.
 	languageKeywords map[string][]string = map[string][]string{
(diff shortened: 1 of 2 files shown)


andrewsomething (approved):
:+1:  LGTM

PR #1226, opened 2022-09-07
Eliminate plugin usage in 'doctl sls fn invoke'

joshuaauerbachwatson:
This change is one in an ongoing process of reducing the dependence of `doctl sls` commands on the `nodejs`-based "sandbox" plugin.   It redoes the `doctl sls fn invoke` command to go directly to backend APIs without using the plugin.


andrewsomething (approved):
:+1:  LGTM

PR #1228, opened 2022-09-11
Connect ReadProject service to the get-metadata command to support testing and further progress

joshuaauerbachwatson:
After the merge of PR #1198 there was a (hidden) partly implemented `ReadProject` service and a hidden flag in the `get-metadata` command that would cause it to be exercised.   However, the two were not meaningfully connected to each other, so, the intent (comparing `get-metadata` results with and without the flag to guide the next phase of implementation) was not accomplished.

This change rectifies the omission in the earlier PR, then, based on some actual testing, fills in some more of the `ReadProject` functionality to the point where the outputs, with and without the flag, are easier to compare.   The result is not ready to be unhidden and used in production, but we didn't expect that, this just brings us a step closer.

### DIFF
--- a/commands/serverless-extra.go
+++ b/commands/serverless-extra.go
@@ -155,19 +155,23 @@ func RunServerlessExtraGetMetadata(c *CmdConfig) error {
 	if err != nil {
 		return err
 	}
-	r, _ := c.Doit.GetBool(c.NS, flagProjectReader)
+	useProjectReader, _ := c.Doit.GetBool(c.NS, flagProjectReader)
 
 	var output do.ServerlessOutput
 	project := do.ServerlessProject{
 		ProjectPath: c.Args[0],
 	}
-	if r {
-		output, err = c.Serverless().ReadProject(&project, c.Args)
+	if useProjectReader {
+		spec, err := c.Serverless().ReadProject(&project)
+		if err != nil {
+			return err
+		}
+		output.Entity = spec
 	} else {
 		output, err = RunServerlessExec(projectGetMetadata, c, []string{flagJSON, flagProjectReader}, []string{flagEnv, flagInclude, flagExclude})
-	}
-	if err != nil {
-		return err
+		if err != nil {
+			return err
+		}
 	}
 	return c.PrintServerlessTextOutput(output)
 }
--- a/do/serverless.go
+++ b/do/serverless.go
@@ -194,7 +194,7 @@ type ServerlessService interface {
 	InvokeFunction(string, interface{}, bool, bool) (map[string]interface{}, error)
 	InvokeFunctionViaWeb(string, map[string]interface{}) error
 	GetConnectedAPIHost() (string, error)
-	ReadProject(*ServerlessProject, []string) (ServerlessOutput, error)
+	ReadProject(*ServerlessProject) (ProjectSpec, error)
 	WriteProject(ServerlessProject) (string, error)
 }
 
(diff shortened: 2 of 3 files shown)


commands/serverless-extra.go
```diff
@@ -155,19 +155,23 @@ func RunServerlessExtraGetMetadata(c *CmdConfig) error {
 	if err != nil {
 		return err
 	}
```
joshuaauerbachwatson, line 157: Changes in this file are to adjust to the signature change in `ReadProject`.   It returns a `ProjectSpec` rather than ready-for-output information.

do/serverless.go
```diff
@@ -131,21 +131,44 @@ type Annotation struct {
...
+// This structure is never serialized.
 type ServerlessProject struct {
-	ProjectPath string   `json:"project_path"`
-	ConfigPath  string   `json:"config"`
-	Packages    string   `json:"packages"`
-	Env         string   `json:"env"`
-	Strays      []string `json:"strays"`
+	ProjectPath string                 `json:"project_path"`
+	ConfigPath  string                 `json:"config"`
+	Packages    []ServerlessPackageDir `json:"packages"`
```
joshuaauerbachwatson, line 141: This needs to be an array, and the elements of the array have substructure.

do/serverless.go
```diff
@@ -131,21 +131,44 @@ type Annotation struct {
...
+// TODO the FileStructure member is temporary, since the work to merge file/directory structure in
+// ServerlessProject with the spec information here is still pending.
 type ProjectSpec struct {
-	Parameters  map[string]interface{} `json:"parameters,omitempty"`
-	Environment map[string]interface{} `json:"environment,omitempty"`
-	Packages    []*ProjectSpecPackage  `json:"packages,omitempty"`
+	Parameters    map[string]interface{} `json:"parameters,omitempty"`
+	Environment   map[string]interface{} `json:"environment,omitempty"`
+	Packages      []*ProjectSpecPackage  `json:"packages,omitempty"`
+	FileStructure *ServerlessProject     `json:"file_sructure,omitempty"`
```
joshuaauerbachwatson, line 171: As the comment above states, the `FileStructure` member is temporary pending the implementation of a proper merge of information from the file structure with information from the `project.yml` file.    What's in the `FileStructure` information needs to be absorbed into the `Packages` member of the `ProjectSpec`

do/serverless.go
```diff
@@ -680,52 +704,148 @@ func (s *serverlessService) GetConnectedAPIHost() (string, error) {
...
+	var (
+		empty  ProjectSpec
+		answer ProjectSpec
+	)
+	err := readProjectStructure(project)
 	if err != nil {
-		return ServerlessOutput{}, err
+		return empty, err
+	}
+	if project.ConfigPath != "" {
```
joshuaauerbachwatson, line 719: Note that not all projects have a `project.yml`.   Some are expressed purely through file structure.   The (as yet unimplemented) merge of the two sources of information needs to deal correctly with that.

joshuaauerbachwatson:
Although I _had_ been wondering why this wasn't getting reviewed and was just about to nudge, I have since concluded that this PR is low priority right now and some higher priority work is coming up.   So, I am reverting this to draft status for the moment.

joshuaauerbachwatson:
I've decided to close this PR without merging because it isn't clear any more how it fits in our long-range goals.  I'll keep the branch with the code around to be picked up later if appropriate.

PR #1230, opened 2022-09-14
appdev: static site builds

kamaln7:


internal/apps/builder/docker.go
```diff
@@ -14,139 +14,165 @@ import (
...
+ARG output_dir
+
+COPY nginx.conf /etc/nginx/conf.d/default.conf
+COPY --from=content ${output_dir}/ /www
+`)
+
+	buildArgs = map[string]*string{
+		"app_image":   strPtr(b.AppImageOutputName()),
+		"nginx_image": strPtr(staticSiteNginxImage),
+		"output_dir":  strPtr(c.GetOutputDir()),
```
nicktate, line 173: Do we need to join this with the workspace path similar to the cnb asset path calculation?

nicktate (approved):
Overall this looks great, nice cleanup on the initial POC 🙇🏻 ; one small question in-line about the dockerfile-based asset path

internal/apps/builder/docker.go
kamaln7: nope, `output_dir` must be an absolute path for dockerfile builds. we are missing that validation rule though—I'll add it

kamaln7: https://github.com/digitalocean/doctl/pull/1230/commits/7fa0c2a8eff9d634af4a10e30048ae004a78fbcb

PR #1232, opened 2022-09-15
Add doctl serverless trigger support (for scheduled functions)

joshuaauerbachwatson:
This PR adds support (still hidden) for a future release of `doctl serverless` that includes support for scheduled functions.   Concretely, it supports triggers, associated with functions, where the trigger has a `cron` expression governing when the function is invoked.

Support consists of the following:

1.  In `doctl sls deploy`, if the `project.yml` contains a `triggers` object nested in a function object, the triggers are deployed, associated with the function.   This is accomplished by adopting the latest serverless plugin version, which incorporates the latest deployer version.
2. In `doctl sls undeploy`, if there are triggers associated with any functions that are being undeployed, they are undeployed as well.   There is also a new flag, `--triggers` (currently hidden) which allows triggers to be undeployed independent of the functions they are attached to.   This is accomplished by a combination of changes in the deployer, and direct calls to the triggers API (when deleting only triggers).
3. There is a new command `doctl sls triggers` (currently hidden) with subcommands `list` and `get` that allows discovery of what triggers exist and permits further details to be inspected.   These commands do not use the plugin but go directly to the triggers API.

### DIFF
--- a/commands/command_config.go
+++ b/commands/command_config.go
@@ -118,7 +118,7 @@ func NewCmdConfig(ns string, dc doctl.Config, out io.Writer, args []string, init
 			c.Apps = func() do.AppsService { return do.NewAppsService(godoClient) }
 			c.Monitoring = func() do.MonitoringService { return do.NewMonitoringService(godoClient) }
 			c.Serverless = func() do.ServerlessService {
-				return do.NewServerlessService(godoClient, getServerlessDirectory(), hashAccessToken(c))
+				return do.NewServerlessService(godoClient, getServerlessDirectory(), accessToken)
 			}
 
 			return nil
--- a/commands/serverless.go
+++ b/commands/serverless.go
@@ -35,6 +35,9 @@ var (
 	// errUndeployTooFewArgs is the error returned when neither --all nor args are specified on undeploy
 	errUndeployTooFewArgs = errors.New("either command line arguments or `--all` must be specified")
 
+	// errUndeployTrigPkg is the error returned when both --packages and --triggers are specified on undeploy
+	errUndeployTrigPkg = errors.New("the `--packages` and `--triggers` flags are mutually exclusive")
+
 	// languageKeywords maps the backend's runtime category names to keywords accepted as languages
 	// Note: this table has all languages for which we possess samples.  Only those with currently
 	// active runtimes will display.
@@ -100,11 +103,14 @@ Functions should be listed in `+"`"+`pkgName/fnName`+"`"+` form, or `+"`"+`fnNam
 The `+"`"+`--packages`+"`"+` flag causes arguments without slash separators to be intepreted as packages, in which case
 the entire packages are removed.`, Writer)
 	AddBoolFlag(undeploy, "packages", "p", false, "interpret simple name arguments as packages")
+	AddBoolFlag(undeploy, "triggers", "", false, "interpret all arguments as triggers")
 	AddBoolFlag(undeploy, "all", "", false, "remove all packages and functions")
+	undeploy.Flags().MarkHidden("triggers") // support is experimental at this point
 
 	cmd.AddCommand(Activations())
 	cmd.AddCommand(Functions())
 	cmd.AddCommand(Namespaces())
+	cmd.AddCommand(Triggers())
 	ServerlessExtras(cmd)
 	return cmd
 }
@@ -359,6 +365,7 @@ func showLanguageInfo(c *CmdConfig, APIHost string) error {
 func RunServerlessUndeploy(c *CmdConfig) error {
 	haveArgs := len(c.Args) > 0
 	pkgFlag, _ := c.Doit.GetBool(c.NS, "packages")
+	trigFlag, _ := c.Doit.GetBool(c.NS, "triggers")
 	all, _ := c.Doit.GetBool(c.NS, "all")
 	if haveArgs && all {
 		return errUndeployAllAndArgs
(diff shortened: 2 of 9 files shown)


do/serverless.go
```diff
@@ -204,6 +230,7 @@ type serverlessService struct {
 	credsDir      string
 	node          string
 	userAgent     string
+	accessToken   string
```
joshuaauerbachwatson, line 233: We are still using the plugin for 'doctl sls deploy|undeploy' and it cannot contact the triggers API without an actual access token.

do/serverless.go
```diff
@@ -276,7 +303,7 @@ type ServerlessOutput struct {
 }
 
 // NewServerlessService returns a configured ServerlessService.
-func NewServerlessService(client *godo.Client, usualServerlessDir string, credsToken string) ServerlessService {
+func NewServerlessService(client *godo.Client, usualServerlessDir string, accessToken string) ServerlessService {
```
joshuaauerbachwatson, line 306: We need the actual access token, not just a hash of it, to support invoking the deployer and enabling it to manage triggers.

do/serverless.go
```diff
@@ -296,13 +324,25 @@ func NewServerlessService(client *godo.Client, usualServerlessDir string, credsT
...
 		client:        client,
 		owClient:      nil,
+		accessToken:   accessToken,
 	}
 }
 
+// HashAccessToken converts a DO access token string into a shorter but suitably random string
+// via hashing.  This is used to form part of the path for storing OpenWhisk credentials
+func HashAccessToken(token string) string {
+	hasher := sha1.New()
```
joshuaauerbachwatson, line 334: Relocated from the `commands` package to the `do` package to avoid having to pass both the token and its hash at init time.

do/serverless.go
```diff
@@ -212,7 +239,7 @@ const (
 	// Minimum required version of the sandbox plugin code.  The first part is
 	// the version of the incorporated Nimbella CLI and the second part is the
 	// version of the bridge code in the sandbox plugin repository.
-	minServerlessVersion = "4.2.3-1.3.1"
+	minServerlessVersion = "4.2.6-1.3.1"
```
joshuaauerbachwatson, line 242: 4.2.6 contains necessary supporting capabilities in the deployer.

commands/command_config.go
```diff
@@ -118,7 +118,7 @@ func NewCmdConfig(ns string, dc doctl.Config, out io.Writer, args []string, init
 			c.Apps = func() do.AppsService { return do.NewAppsService(godoClient) }
 			c.Monitoring = func() do.MonitoringService { return do.NewMonitoringService(godoClient) }
 			c.Serverless = func() do.ServerlessService {
-				return do.NewServerlessService(godoClient, getServerlessDirectory(), hashAccessToken(c))
+				return do.NewServerlessService(godoClient, getServerlessDirectory(), accessToken)
```
joshuaauerbachwatson, line 121: See my comment about this change in do/serverless.go

commands/serverless_util.go
```diff
@@ -112,11 +110,7 @@ func (c *CmdConfig) PrintServerlessTextOutput(output do.ServerlessOutput) error
 }
 
 func hashAccessToken(c *CmdConfig) string {
-	token := c.getContextAccessToken()
-	hasher := sha1.New()
-	hasher.Write([]byte(token))
-	sha := hasher.Sum(nil)
-	return hex.EncodeToString(sha[:4])
+	return do.HashAccessToken(c.getContextAccessToken())
```
joshuaauerbachwatson, line 113: Most of the logic is relocated to the do package but call sites remain in the commands package, hence the original function is retained.

commands/serverless.go
```diff
@@ -359,21 +365,36 @@ func showLanguageInfo(c *CmdConfig, APIHost string) error {
...
+	var sls do.ServerlessService
+	if trigFlag {
+		ctx = context.TODO()
+		sls = c.Serverless()
+	}
 	for _, arg := range c.Args {
 		var err error
-		if strings.Contains(arg, "/") || !pkgFlag {
+		if trigFlag {
+			err = sls.DeleteTrigger(ctx, arg)
```
joshuaauerbachwatson, line 396: Undeploy is currently a bit irregular in that it continues to rely on the plugin for _most_ cases but the deployer has no direct analog for deleting just a trigger without its attached function.  So, here, we call the triggers API directly to effect deletion.  This will all get simplified / consolidated as part of the ongoing effort to reduce dependence on the plugin.

do/serverless.go
```diff
@@ -745,6 +783,100 @@ func (s *serverlessService) WriteProject(project ServerlessProject) (string, err
...
+		return err
+	}
+	path := "v2/functions/trigger/" + creds.Namespace + "/" + name
+	_, err = s.client.NewRequest(ctx, http.MethodDelete, path, nil)
+	return err
+}
+
+// fixupCron detects the optional seconds field and removes it
+func fixupCron(cron string) string {
+	parts := strings.Split(cron, " ")
```
joshuaauerbachwatson, line 873: Seconds are not supported externally but the scheduler adds a random seconds field when storing the task in order to avoid bursts.   In the development of the various API layers sometimes we. were getting back 6-element cron expressions (in which the first field, seconds, was misleading) and sometimes these were already trimmed.   The code here will work either way and either way the seconds are not displayed.

commands/triggers.go
```diff
@@ -0,0 +1,98 @@
...
+	return cmd
+}
+
+// RunTriggersList provides the logic for 'doctl sls trig list'
+func RunTriggersList(c *CmdConfig) error {
+	if len(c.Args) > 0 {
+		return doctl.NewTooManyArgsErr(c.NS)
+	}
+	fcn, _ := c.Doit.GetString(c.NS, "function")
+	list, err := c.Serverless().ListTriggers(context.TODO(), fcn)
```
andrewsomething, line 58: These commands use the DO API rather than the OpenWisk one, but they still depend on the `credentials.json` file to determine the connected namespace. It might be nice to guard them with a status check so that they provide a hint to connect if run without having done so instead of just `no such file or directory`.

    $ ./builds/doctl serverless triggers list 
    Error: open /home/asb/.config/doctl/sandbox/creds/ea841a9e/credentials.json: no such file or directory

andrewsomething (commented):
Overall, looks great. I did have a suggestion in-line regarding ensuring that a namespace is actually connected.

commands/triggers.go
joshuaauerbachwatson: I did as suggested. 

In thinking about this issue, I suspect there are other cases where we should be guarding with a status check but are not.   The original idea was that that status check could be placed in utility functions that dispatch to the plugin rather that duplicating it in individual commands.   But, we have a gradually increasing number of commands that don't go through the plugin now and the whole area probably deserves a bit of refactoring.   I would like to defer that more general investigation to a follow-on, though, since I don't want to complicate this PR unduly.

PR #1235, opened 2022-09-19
Ntate/appdev/macossupport

nicktate:

* Mounts the vm socket in docker-for-desktop mac

### DIFF
--- a/internal/apps/builder/cnb.go
+++ b/internal/apps/builder/cnb.go
@@ -25,7 +25,7 @@ import (
 
 const (
 	// CNBBuilderImage represents the local cnb builder.
-	CNBBuilderImage = "digitaloceanapps/cnb-local-builder:v0.49.0"
+	CNBBuilderImage = "digitaloceanapps/cnb-local-builder:v0.50.2"
 
 	appVarAllowListKey = "APP_VARS"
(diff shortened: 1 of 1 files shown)


internal/apps/builder/cnb.go
```diff
@@ -63,9 +64,16 @@ func (b *CNBComponentBuilder) Build(ctx context.Context) (res ComponentBuilderRe
...
-	sourceDockerSock, err := filepath.EvalSymlinks(dockerSocketPath)
-	if err != nil {
-		return res, fmt.Errorf("finding docker engine socket: %w", err)
+	var sourceDockerSock string
+	switch runtime.GOOS {
+	case "darwin":
+		// mac docker-for-desktop includes the raw socket in the VM
+		sourceDockerSock = "/var/run/docker.sock.raw"
+	default:
+		sourceDockerSock, err = filepath.EvalSymlinks(dockerSocketPath)
```
kamaln7, line 73: This will break tests on macOS as `/var/run/docker.sock.raw` won't exist even if docker is installed. Can we replace the global `dockerSocketPath` with a `func() string` that implements this `switch` so we can override it still in tests

kamaln7 (approved):

PR #1247, opened 2022-09-23
disable stdin from `app dev build` command

nicktate:

Accepting an app spec via `stdin` does not play well with the `Interactivity` config. Rather than implicitly disable interactivity in that scenario we are removing support for `stdin` to `app dev build --spec=-`; users can still load arbitrary specs with a path based syntax.


andrewsomething (approved):
:+1:

PR #1252, opened 2022-09-26
Finish eliminating plugin usage in 'doctl sls functions' + testing improvements

joshuaauerbachwatson:
This change covers internal improvements not intended to be visible to typical `doctl` users.

1. Toward the goal of reducing the separate plugin for serverless:
     1.  Remove dependence on the plugin for `doctl sls fn list`
     2. Remove dependence on the plugin for `doctl sls status`
2.  Adds a (permanently) hidden pair of flags to `doctl sls connect` permitting connection to serverless clusters that are not in production (and hence unknown to the serverless portal).   This is needed to support internal testing and debugging use cases for which we currently use the `nim` CLI (which should be phased out).

### DIFF
--- a/commands/functions.go
+++ b/commands/functions.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"sort"
 	"strings"
 
 	"github.com/apache/openwhisk-client-go/whisk"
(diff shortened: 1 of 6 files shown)


joshuaauerbachwatson:
This will stay as a draft until PR #1232 is merged since it has higher priority.  Then it will be rebased and marked for review.

do/serverless.go
```diff
@@ -697,6 +720,23 @@ func (s *serverlessService) GetFunction(name string, fetchCode bool) (whisk.Acti
...
 }
 
+// ListFunctions lists the functions of the connected namespace
+func (s *serverlessService) ListFunctions(pkg string, skip int, limit int) ([]whisk.Action, error) {
+	err := initWhisk(s)
+	if err != nil {
+		return []whisk.Action{}, err
+	}
+	if limit == 0 {
+		limit = 30
```
joshuaauerbachwatson, line 730: This was the default before the changeover.   Maintaining it to avoid surprising new behaviors.

commands/functions.go
```diff
@@ -250,34 +251,48 @@ func RunFunctionsList(c *CmdConfig) error {
...
-	// Add JSON flag so we can control output format
-	if !count {
-		c.Doit.Set(c.NS, flagJSON, true)
-	}
-	output, err := RunServerlessExec(actionList, c, []string{flagCount, flagNameSort, flagNameName, flagJSON}, []string{flagLimit, flagSkip})
+	// Retrieve other flags
+	skip, _ := c.Doit.GetInt(c.NS, flagSkip)
+	limit, _ := c.Doit.GetInt(c.NS, flagLimit)
+	nameSort, _ := c.Doit.GetBool(c.NS, flagNameSort)
+	nameName, _ := c.Doit.GetBool(c.NS, flagNameName)
```
andrewsomething, line 268: This is not something new to this PR, but I wonder if we should hide one of these flags as they have the same purpose?

andrewsomething (approved):
:+1:  LGTM. I did have one question in line, but it is not specific to this PR.

commands/functions.go
joshuaauerbachwatson: The Apache OpenWhisk CLI (`wsk`) just uses `--name-sort`, short form `-n`.   When Adobe issued their CLI (`aio`), it appears that they introduced the "alias" flag `--name` and they moved the `-n` shortcut to it.   By itself, this flag is indeed kind of confusing (seems like it should take a value and allow you to specify a name, somehow).  

When I incorporated parts of `aio` into `nim` I retained the alias behavior.   When I created the `doctl` delegation path I again retained the behavior.   In this PR, in recoding the logic in `golang` I _again_ retained the behavior (copying is easier than thinking).   I think it's appropriate to go back to what `wsk` does, which retains `--name-sort` and `-n` but not `--name`.   This is technically a breaking change but I think it's extremely unlikely to matter to anyone.

@andrewsomething do you agree with that plan?  If so, I'm inclined to just do it (by tacking the change onto this PR).   What do you think?

joshuaauerbachwatson: Ach.  Since it is a behavior change, and technically breaking, I decided not to do it as part of this PR so I'm proceeding to merge.   We can think about making the proposed simplification in the future.

andrewsomething: :+1:  Sounds good.

PR #1257, opened 2022-09-28
appdev: correct source_dir and dockerfile_path interaction

kamaln7:
dockerfile_path in the app spec is relative to the workspace root and so it can be outside of source_dir.
However the docker tooling expects it to be present in the build context that is uploaded to the docker engine and therefore relative to source_dir.

This fixes the interaction between the two options and adds support for building with dockerfiles that are outside of source_dir.

Most of the logic is copied from the `docker` cli: https://github.com/docker/cli/blob/9400e3dbe8ebd0bede3ab7023f744a8d7f4397d2/cli/command/image/build.go#L180

Test are WIP but putting up a PR to unblock reviews.


nicktate (commented):
updates lgtm; we will need to update some of the tests to account for the new docker build context processing

andrewsomething:
@kamaln7  Can you include that link in a code comment? It will be extremely helpful for any future debugging.

ChiefMateStarbuck (approved):

nicktate (approved):

PR #1258, opened 2022-09-29
Reveal trigger support

joshuaauerbachwatson:
This change removes the "Hidden" option from the cobra sequences defining trigger support (that is, from `doctl sls triggers` and from `doctl sls undeploy --triggers`). 

The latest `nim` version is also adopted, which provides more robustness in the deletion of triggers when trigger deletion fails for some reason.


joshuaauerbachwatson:
We won't be revealing trigger support nearly as soon as I thought so closing this.

PR #1261, opened 2022-09-30
pkg/extract error handling refactoring

heartwilltell:
1. It is always a great idea to use the following approach 👇

```go
if err := myFunc(); err != nil {
    return err
}
```
instead of this one 👇
```go
err := myFunc() 
if err != nil {
    return err
}
```

In the first example the `err` is in the scope of `if` block, which reduces the possibility of overwriting an existing variable with the same name.

2. It is always a good practice to name errors differently. While `i err != nil {}` is a very nice and convinient, it is at the same time very error prone in cases when we need to handle many error in the same fuction. Different naming for errors helps to avoid the _shadowing_ of errors.

### DIFF
--- a/pkg/extract/extract.go
+++ b/pkg/extract/extract.go
@@ -18,20 +18,19 @@ func Extract(source, target string) error {
 	if _, err := os.Stat(target); os.IsNotExist(err) {
 		return errors.New("target directory does not exist")
 	}
+
 	if _, err := os.Stat(source); os.IsNotExist(err) {
 		return errors.New("source archive does not exist")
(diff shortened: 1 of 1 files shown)


ChiefMateStarbuck (commented):
hello @heartwilltell,

Yes, you are correct. the refactored error handling is a better and safer option. 

Unfortunately, the older method of error handling stretches through the entire codebase and would take a few thousand lines of changes. For safety of the codebase, the old method will be used and the new method will be used if a V2 of doctl is ever written. Thank you for the PR.

PR #1265, opened 2022-09-30
Migrate existing confirmation dialogs to use confirm charm.

andrewsomething:
This migrates all of our existing confirmation dialogs to use the new `confirm` package added with the `app dev` command. This seems like a good place to start incorporating that work into the rest of `doctl` as we get a lot of bang for our buck. Changing the existing helper method brings it into many of our existing commands. Though for that same reason, I'm opening this as a draft for discussion first. We need to ensure there is no unexpected breakage.

The biggest functional difference is that you can now use the arrow keys to move between the options rather than just typing, but typing `y` or `n` still works. It's not a huge change, but I think it brings a more modern feeling.

Old prompt:

image

Old no selected:

image

Old yes selected:

image

New propmt: 

image

New no selected: 

image

New yes selected:

image


bentranter (approved):
Looks like a nice simple change to the code, I say we go for it

PR #1266, opened 2022-10-01
Add --wait flag to load_balancer create

### LINKED ISSUE #1264: Add --wait flag to load-balancer create command

`doctl` aims to be useful in scripting. One way it does this is by providing a `--wait` flag for long running, async operations to block until until they are complete rather than making the user implement polling in their own script. Some examples include:

- `doctl compute droplet create --wait`
- `doctl kubernetes cluster create --wait`
- `doctl apps create --wait`

Currently the `doctl compute load-balancer create` command does not have a `--wait` flag.

Add a `--wait` flag to the `doctl compute load-balancer create` command. It should poll the [`/v2/load_balancers/$LOAD_BALANCER_ID` endpoint](https://docs.digitalocean.com/reference/api/api-reference/#operation/loadBalancers_get), blocking until the load balancer's `status` attribute moves from `new` to `active`.

**Additional context**

Following the code path starting here demonstrates how `doctl apps create`'s  `--wait` flag is implemented:

https://github.com/digitalocean/doctl/blob/de1918ad1ee4b1e523c335fa9254e85d415d18aa/commands/apps.go#L282

VictorAvelar:
## Description

Adds `--wait` flag to the doctl compute load-balancer create command. 

It polls the [/v2/load_balancers/$LOAD_BALANCER_ID endpoint](https://docs.digitalocean.com/reference/api/api-reference/#operation/loadBalancers_get), blocking until the load balancer's status attribute moves from new to active.

## Motivation and context

Closes #1264 

## Types of changes

Feature enhancement.

## Checklist:

- [x] I have read the **[CONTRIBUTING](CONTRIBUTING.md)** document.
- [x] My pull request addresses exactly one patch/feature.
- [x] I have created a branch for this patch/feature.
- [x] Each individual commit in the pull request is meaningful.
- [x] I have added tests to cover my changes.
- [ ] If my change requires a change to the documentation, I have updated it accordingly.

### DIFF
--- a/commands/load_balancers.go
+++ b/commands/load_balancers.go
@@ -15,9 +15,11 @@ package commands
 
 import (
 	"fmt"
+	"os"
 	"reflect"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/digitalocean/doctl"
 	"github.com/digitalocean/doctl/commands/displayers"
@@ -98,6 +100,7 @@ With the load-balancer command, you can list, create, or delete load balancers, 
 		"A comma-separated list of key-value pairs representing recent health check results, e.g.: `protocol:http,port:80,path:/index.html,check_interval_seconds:10,response_timeout_seconds:5,healthy_threshold:5,unhealthy_threshold:3`")
 	AddStringFlag(cmdRecordCreate, doctl.ArgForwardingRules, "", "",
 		forwardingRulesTxt)
+	AddBoolFlag(cmdRecordCreate, doctl.ArgCommandWait, "", false, "Boolean that specifies whether to wait for a load balancer to complete before returning control to the terminal")
 
 	cmdRecordUpdate := CmdBuilder(cmd, RunLoadBalancerUpdate, "update <id>",
 		"Update a load balancer's configuration", `Use this command to update the configuration of a specified load balancer.`, Writer, aliasOpt("u"))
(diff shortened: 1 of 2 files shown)


commands/load_balancers.go
```diff
@@ -513,3 +538,37 @@ func buildRequestFromArgs(c *CmdConfig, r *godo.LoadBalancerRequest) error {
...
+
+		lb, err := lbs.Get(lbID)
+		if err != nil {
+			return err
+		}
+
+		if lb.Status == wantStatus {
+			return nil
+		}
+
```
andrewsomething, line 565: We should also check if the status is `errored`. If so, we can exit early with an error rather than waiting for the timeout.

commands/load_balancers.go
```diff
@@ -201,6 +204,28 @@ func RunLoadBalancerCreate(c *CmdConfig) error {
...
+			return fmt.Errorf(
+				"load balancer couldn't enter `active` state: %v",
+				err,
+			)
+		}
+
+		lb, _ = lbs.Get(lb.ID)
+	}
+
+	notice("Load balancer created")
```
andrewsomething, line 227: This line is causing the test failures.  We could remove it or adjust the expected output:

https://github.com/digitalocean/doctl/blob/dc8385a75cf9062b7de5f3a68c22aa780dbb31fb/integration/lb_create_test.go#L103

andrewsomething (changes requested):
This is great work! I left a few comments in line.

It would also be great to add a basic test here: 

https://github.com/digitalocean/doctl/blob/main/integration/lb_create_test.go

That would require adding a new case to the mock server that handles GET `/v2/load_balancers/$LOAD_BALANCER_ID` and returns a response with the status set to `active`. Let me know if you have any questions.

commands/load_balancers.go
VictorAvelar: You are right, I will change it.

commands/load_balancers.go
VictorAvelar: Which one will you prefer? IMO we should add the notice to make it consistent with the other implementation but I can also get rid of it, just need to know which one makes more sense from the maintainer POV.

andrewsomething: :+1:  Good point. Let's leave it in.

andrewsomething (approved):
:+1:  Looks good! Thank you so much for this great contribution!

VictorAvelar:
Thanks for your guidance and comments. 🚀

Issue #1262, opened 2022-09-30
Use consistent aliases for `list` and `delete` sub-commands


andrewsomething:
**What is the problem this feature would solve? Please describe.**

Many `doctl` commands provide aliases allowing you to use a shorter version of the command. In particular, `list` is often aliased as `ls` and `delete` is often aliased as `rm`. These are short memorable aliases that recall common Unix commands. Unfortunately, we have not always made sure every new sub-command adds these aliases. We should provide a consistent experience. User should know that they can use `ls` and `list` interchangeably across `doctl`.

**Describe the solution you'd like**

- `ls` should be added as an alias for all  `list` sub-commands
- `rm` should be added as an alias for all  `delete` sub-commands

You can find an example here:

https://github.com/digitalocean/doctl/blob/c7bf8d427e73e395f59f743cee05f6747392b677/commands/sizes.go#L35-L36

**Additional context**

Adding a test to `commands/commands_test.go` to iterate through the commands and inspect their aliases can help to identify where changes are needed. For example: 

```go
func assertCommandAliases(t *testing.T, cmd *cobra.Command) {
	for _, c := range cmd.Commands() {
		if c.Name() == "list" {
			assert.Contains(t, c.Aliases, "ls", "Missing 'ls' alias for 'list' command.")
		}
		if c.Name() == "delete" {
			assert.Contains(t, c.Aliases, "rm", "Missing 'rm' alias for 'delete' command.")
		}
	}
}

func TestCommandAliases(t *testing.T) {
	for _, cmd := range DoitCmd.Commands() {
		t.Run(cmd.Name(), func(t *testing.T) {
			assertCommandAliases(t, cmd)
			for _, c := range cmd.Commands() {
				t.Run(c.Name(), func(t *testing.T) {
					assertCommandAliases(t, c)
				})
			}
		})
	}
}
```

selborsolrac:
Hello, I'd like to work on this issue for Hacktoberfest 2022!

Thank you for the example test logic. I used this slightly modified version to find missing aliases:
```
func assertCommandAliases(t *testing.T, cmd *cobra.Command) {
	for _, c := range cmd.Commands() {
		if c.Name() == "list" {
			assert.Contains(t, c.Aliases, "ls", "Missing 'ls' alias for 'list' command.")
		}
		if c.Name() == "delete" {
			assert.Contains(t, c.Aliases, "rm", "Missing 'rm' alias for 'delete' command.")
		}
	}
}

[...]

PR #1269, opened 2022-10-04
commands: Consistent aliases for "list" and "delete"

selborsolrac:
This pull request
- Adds missing `ls` and `rm` aliases to their respective `list` and `delete` subcommands
- Changes the alias order of `doctl compute volume delete` to `d, rw` in order to be consistent with other `delete` alias uses
- Adds the test suggested in the linked issue in order to ensure any future `list` and `remove`  sub-commands have the expected aliases.

Fixes https://github.com/digitalocean/doctl/issues/1262

Please let me know if I missed any aliases!

@andrewsomething


andrewsomething (approved):
:+1:  Nice work. Thanks for another great contribution!

PR #1270, opened 2022-10-04
Avoid plugin in serverless activations [ get | result ]

joshuaauerbachwatson:
This change alters the logic for `doctl sls actv get` and `doctl sls actv result` to use the OpenWhisk API directly rather than using the plugin.   The `list` and `logs` commands are still using the plugin (to be addressed in a future PR).

### DIFF
--- a/commands/activations.go
+++ b/commands/activations.go
@@ -14,10 +14,15 @@ limitations under the License.
 package commands
 
 import (
+	"encoding/json"
 	"fmt"
+	"io"
+	"regexp"
+	"time"
 
 	"github.com/apache/openwhisk-client-go/whisk"
 	"github.com/digitalocean/doctl"
+	"github.com/digitalocean/doctl/commands/charm/text"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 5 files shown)


commands/activations.go
```diff
@@ -85,12 +100,146 @@ func RunActivationsGet(c *CmdConfig) error {
...
+		if err != nil {
+			return err
+		}
+		if len(list) == 0 {
+			return fmt.Errorf("no activations were returned")
+		}
+		activation := list[0]
+		id = activation.ActivationID
+		if !quietFlag && (logsFlag || resultFlag) {
+			makeBanner(c.Out, activation)
```
andrewsomething, line 138: Wonder if these banners should be printed to stderr?

andrewsomething (approved):
:+1: I did have one questions inline, but LGTM!

commands/activations.go
joshuaauerbachwatson: I believe the purpose of the banners is to avoid letting the logs or results from different activations merge together.  I believe they were added recently (in `aio`) for that purpose and I've simply retained them here.   In order for them to do that job I think they need to go to the same output stream as the logs and results themselves (so that they are inline).   That suggests keeping them on stdout.

andrewsomething: :+1:

PR #1272, opened 2022-10-05
functions: action.Exec.Code is nil when saving env.

andrewsomething:
`doctl sbx fn get` panics when passing the `--save-env` or `--save-env-json` flags. This is because  `action.Exec.Code` is nil when calling `doSavingForFunctionGet` since we only fetch the code when `codeFlag || saveFlag || saveAsFlag` is true. I've broken the environment saving out of the code saving function to make their individual purposes more clear.

As an additional follow up, we may want to error if some of these flags are used together as they seem to be mutually exclusive.

### DIFF
--- a/commands/functions.go
+++ b/commands/functions.go
@@ -34,7 +34,7 @@ func Functions() *Command {
 		Command: &cobra.Command{
 			Use:   "functions",
 			Short: "Work with the functions in your namespace",
-			Long: `The subcommands of ` + "`" + `doctl serverless functions` + "`" + ` operate on your functions namespace. 
+			Long: `The subcommands of ` + "`" + `doctl serverless functions` + "`" + ` operate on your functions namespace.
 You are able to inspect and list these functions to know what is deployed.  You can also invoke functions to test them.`,
 			Aliases: []string{"fn"},
 		},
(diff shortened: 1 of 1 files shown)


commands/functions.go
```diff
@@ -87,34 +87,44 @@ func RunFunctionsGet(c *CmdConfig) error {
...
 	output := do.ServerlessOutput{Entity: action}
 	return c.PrintServerlessTextOutput(output)
 }
 
-// doSavingForFunctionGet performs the save operations for code and for environment variables.
-func doSavingForFunctionGet(action whisk.Action, save bool, saveAs string, saveEnv string, saveEnvJSON string,
-	parms []do.FunctionParameter) error {
+// doSaveFunctionCode performs the save operations for code and for environment variables.
+func doSaveFunctionCode(action whisk.Action, save bool, saveAs string) error {
 	// First process save and saveAs
```
joshuaauerbachwatson, line 128:
```suggestion
```

joshuaauerbachwatson (approved):
I had a couple of minor suggestions (comment cleanup).

LGTM 👍

andrewsomething:
Closing and reopening to hopefully get actions to run...

Issue #1274, opened 2022-10-09
Add a Dependabot to Automate dependency checks


bishal7679:
Add a Dependabot to Automate dependency checks in `.github`

ChiefMateStarbuck:
Thank you @bishal7679 !

PR #1275, opened 2022-10-09
Automate dependency checks with dependabot

bishal7679:
Fixes :- #1274


.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
+updates:
+- package-ecosystem: maven
```
andrewsomething, line 3:
```suggestion
- package-ecosystem: gomod
```

`doctl` is a Go project that manages its dependencies using `gomod`.

.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
...
+  directory: "/"
```
andrewsomething, line 4:
```suggestion
  directory: "/"
  vendor: true
```

We vendor our dependencies in the `vendor/` directory.

.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
...
+  schedule:
+    interval: weekly
+  open-pull-requests-limit: 10
+  labels:
+  - skip-changelog
```
andrewsomething, line 9: I think we can remove this. We want to track all changes in the changelog.

andrewsomething (changes requested):
Hi @bishal7679,

Thanks for the PR! Generally we've been pretty conservative about update dependencies only as needed in this project, but enabling dependabot might be a good experiment.

`doctl`  is a Go project. I've suggested some updates to reflect that.

bishal7679:
> Hi @bishal7679,
> 
> Thanks for the PR! Generally we've been pretty conservative about update dependencies only as needed in the is project, but enabling dependabot might be a good experiment.
> 
> `doctl` is a Go project. I've suggested some updates to reflect that.

Oh! my bad. I have mistakenly used maven its for java project.
Nice suggestion!

bishal7679:
@andrewsomething pls Take a look again!

andrewsomething (approved):
:+1:  Looks good! Thank you for the contribution.

PR #1277, opened 2022-10-11
Re-writes activations list command subtree in Go

ddebarros:
This PR re-write the `doctl sls activations` subtree in native go-lang. All existing behavior is maintained.


joshuaauerbachwatson:
@ddebarros I am seeing commits in this PR that were already merged in the earlier PR.   Did this get properly rebased?

ddebarros:
> @ddebarros I am seeing commits in this PR that were already merged in the earlier PR. Did this get properly rebased?

I rebased it again. Should be all good now.

commands/activations.go
```diff
@@ -245,14 +232,53 @@ func printActivationRecord(writer io.Writer, activation whisk.Activation) {
...
+	sinceFlag, _ := c.Doit.GetInt(c.NS, flagSince)
+	upToFlag, _ := c.Doit.GetInt(c.NS, flagUpto)
+	limitFlag, _ := c.Doit.GetInt(c.NS, flagLimit)
+
+	limit := limitFlag
+	if limitFlag > 200 {
+		limit = 200
+	} else if limitFlag > 0 {
+		limit = limitFlag
+	}
```
joshuaauerbachwatson, line 258: (1) Isn't this now unnecessary since you've changed the cobra definition to enforce the default and max?  (2) In any case, it seems like the else clause cannot result in any actual change since limit is already set to the value of limitFlag.

commands/displayers/activations.go
```diff
@@ -0,0 +1,116 @@
...
+	case 1:
+		return "application error"
+	case 2:
+		return "developer error"
+	case 3:
+		return "system error"
+	default:
+		return "unknown"
+	}
+}
```
joshuaauerbachwatson, line 116: We should avoid having two copies of this function.   This can be accomplished by

1. capitalize this function, which exports it.
2. eliminate the second copy in serverless_util.go.
3. in `commands/activations.go` call it with `displayers.GetActivationStatus`

commands/serverless_util.go
```diff
@@ -163,3 +163,19 @@ func getFlatArgsArray(c *CmdConfig, booleanFlags []string, stringFlags []string)
 func getServerlessDirectory() string {
 	return filepath.Join(defaultConfigHome(), "sandbox")
 }
+
```
joshuaauerbachwatson, line 166: See comment in the displayer.

joshuaauerbachwatson (commented):
Excellent job so far @ddebarros .   I left some comments which should probably be addressed but I think the main issue at this point is that the tests don't pass yet.

commands/activations.go
ddebarros: capping it at 200 is still required, the cobra definition has a default but I don't see a way to define a max value.

commands/activations.go
```diff
@@ -245,14 +232,57 @@ func printActivationRecord(writer io.Writer, activation whisk.Activation) {
...
+	countFlags, _ := c.Doit.GetBool(c.NS, flagCount)
+	fullFlag, _ := c.Doit.GetBool(c.NS, flagFull)
+	skipFlag, _ := c.Doit.GetInt(c.NS, flagSkip)
+	sinceFlag, _ := c.Doit.GetInt(c.NS, flagSince)
+	upToFlag, _ := c.Doit.GetInt(c.NS, flagUpto)
+	limitFlag, _ := c.Doit.GetInt(c.NS, flagLimit)
+
+	limit := limitFlag
+	if limitFlag > 200 {
+		limit = 200
```
joshuaauerbachwatson, line 255: Agreed.  I had an earlier comment based on having misread the cobra (I saw the max in the help text and thought that was code 😄 ).

joshuaauerbachwatson (commented):
I just noticed that this PR includes a change to a vendor-supplied package.  I would think we should not be doing that (certainly not lightly).  I misunderstood in earlier conversations.   I think we need to understand why this change is necessary and can the goal be accomplished differently.

ddebarros:
> I just noticed that this PR includes a change to a vendor-supplied package. I would think we should not be doing that (certainly not lightly). I misunderstood in earlier conversations. I think we need to understand why this change is necessary and can the goal be accomplished differently.

@joshuaauerbachwatson I created a [PR on the openwhisk-client](https://github.com/apache/openwhisk-client-go/pull/155) and updated our go.mod file to reflect this change.

joshuaauerbachwatson (approved):
Excellent.   I approve.
Over to @andrewsomething  for permission to merge.

commands/activations.go
```diff
@@ -245,14 +232,57 @@ func printActivationRecord(writer io.Writer, activation whisk.Activation) {
...
+	var name string
+	if argCount > 0 {
+		name = c.Args[0]
+	}
+
+	countFlags, _ := c.Doit.GetBool(c.NS, flagCount)
+	fullFlag, _ := c.Doit.GetBool(c.NS, flagFull)
+	skipFlag, _ := c.Doit.GetInt(c.NS, flagSkip)
+	sinceFlag, _ := c.Doit.GetInt(c.NS, flagSince)
+	upToFlag, _ := c.Doit.GetInt(c.NS, flagUpto)
```
andrewsomething, line 250: These flags are both constructed with `AddStringFlag` and then `GetInt` is used to retrieve their value. Viper seems to handle the type casting, but might be more clear if we changed to using `AddIntFlag`.  E.g. the help text will say `int` instead of `string` which make it more obvious this should be a Unix timestamp and not some other value.

commands/displayers/activations.go
```diff
@@ -0,0 +1,120 @@
...
+var _ Displayable = &Activation{}
+
+// ColMap implements Displayable
+func (a *Activation) ColMap() map[string]string {
+	return map[string]string{
+		"Datetime":     "Datetime",
+		"Status":       "Status",
+		"Kind":         "Kind",
+		"Version":      "Version",
+		"ActivationId": "Activation ID",
```
andrewsomething, line 25:
```suggestion
		"ActivationID": "Activation ID",
```

For consistency, we generally prefer `ID` over `Id`.

commands/displayers/activations.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// Cols implements Displayable
+func (a *Activation) Cols() []string {
+	return []string{
+		"Datetime",
+		"Status",
+		"Kind",
+		"Version",
+		"ActivationId",
```
andrewsomething, line 40:
```suggestion
		"ActivationID",
```

commands/displayers/activations.go
```diff
@@ -0,0 +1,120 @@
...
+func (a *Activation) KV() []map[string]interface{} {
+	out := make([]map[string]interface{}, 0, len(a.Activations))
+
+	for _, actv := range a.Activations {
+		o := map[string]interface{}{
+			"Datetime":     time.UnixMilli(actv.Start).Format("01/02 03:04:05"),
+			"Status":       GetActivationStatus(actv.StatusCode),
+			"Kind":         getActivationAnnotationValue(actv, "kind"),
+			"Version":      actv.Version,
+			"ActivationId": actv.ActivationID,
```
andrewsomething, line 63:
```suggestion
			"ActivationID": actv.ActivationID,
```

andrewsomething (commented):
Looks great! A did have a few tiny nitpicks inline.

andrewsomething (approved):
:+1:  LGTM!

PR #1285, opened 2022-10-18
Interactive namespace selection

ddebarros:
- Users a more interactive list for selecting a namespace.
- Adds a new SpinningLoader component.

### DIFF
--- a/commands/charm/spinner/spinner.go
+++ b/commands/charm/spinner/spinner.go
@@ -1,0 +1,90 @@
+package spinner
+
+import (
+	"fmt"
+	"os"
+
+	s "github.com/charmbracelet/bubbles/spinner"
+	tea "github.com/charmbracelet/bubbletea"
+)
+
+type SpinningLoader struct {
+	model   s.Model
+	prog    *tea.Program
+	cancel  bool
+	Message string
+}
+
+type Option func(*SpinningLoader)
+
+// New creates a new spinning loader.
+func New(opts ...Option) SpinningLoader {
+	sm := s.New()
+	sm.Spinner = s.Dot
+
+	l := SpinningLoader{
+		model:   sm,
+		Message: "",
+	}
+
+	for _, opt := range opts {
+		opt(&l)
+	}
+	return l
+}
+
+// New creates a new list.
+func (sl *SpinningLoader) Start() error {
+	p := tea.NewProgram((*SpinningLoader)(sl))
+	sl.prog = p
+	p.Start()
+
+	if sl.cancel {
+		os.Exit(1)
+	}
+	return nil
+}
+
+func (sl *SpinningLoader) Stop() {
+	sl.prog.Kill()
+}
+
+// Init implements bubbletea.Model.
+func (sl *SpinningLoader) Init() tea.Cmd {
+	return sl.model.Tick
+}
+
+// Update implements bubbletea.Model.
+func (sl *SpinningLoader) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
+	switch msg := msg.(type) {
+	case tea.KeyMsg:
+		switch keypress := msg.String(); keypress {
+		case tea.KeyCtrlC.String():
+			sl.cancel = true
+			return sl, tea.Quit
+		}
+
+	case s.TickMsg:
+		var cmd tea.Cmd
+		sl.model, cmd = sl.model.Update(msg)
+		return sl, cmd
+	}
+
+	return sl, nil
+}
+
+// View implements bubbletea.Model.
+func (sl *SpinningLoader) View() string {
+	return fmt.Sprintf("%s %s", sl.model.View(), sl.Message)
+}
+
+// Model returns the underlying SpinningLoader.model
+func (sl SpinningLoader) Model() *s.Model {
+	return &sl.model
+}
+
+func WithSpinner(s s.Spinner) Option {
+	return func(sl *SpinningLoader) {
+		sl.model.Spinner = s
+	}
+}
--- a/commands/namespaces.go
+++ b/commands/namespaces.go
@@ -19,7 +19,10 @@ import (
 	"sort"
 	"strings"
 
+	s "github.com/charmbracelet/bubbles/spinner"
 	"github.com/digitalocean/doctl"
+	"github.com/digitalocean/doctl/commands/charm/spinner"
+
 	"github.com/digitalocean/doctl/commands/displayers"
 	"github.com/digitalocean/doctl/do"
 	"github.com/spf13/cobra"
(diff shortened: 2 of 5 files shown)


commands/charm/spinner/spinner.go
```diff
@@ -0,0 +1,95 @@
...
+	l := SpinningLoader{
+		model:   sm,
+		Message: "",
+	}
+
+	for _, opt := range opts {
+		opt(&l)
+	}
+	return l
+}
```
kamaln7, line 34: let's unexport the message field and make it a required argument on New:

```suggestion
type SpinningLoader struct {
	model   s.Model
	prog    *tea.Program
	cancel  bool
	message string
}

type Option func(*SpinningLoader)

// New creates a new spinning loader.
func New(message string, opts ...Option) SpinningLoader {
	sm := s.New()
	sm.Spinner = s.Dot

	l := SpinningLoader{
		model:   sm,
		message: message,
	}

	for _, opt := range opts {
		opt(&l)
	}
	return l
}
```

commands/namespaces.go
```diff
@@ -226,8 +229,15 @@ func getValidRegion(value string) string {
 // get the Namespaces that match a pattern, where the "pattern" has no wildcards but can be a
 // prefix, infix, or suffix match to a namespace ID or label.
 func getMatchingNamespaces(ctx context.Context, ss do.ServerlessService, pattern string) ([]do.OutputNamespace, error) {
+	loader := spinner.New(spinner.WithSpinner(s.Dot))
```
kamaln7, line 232: nit: `s.Dot` is already the default

```suggestion
	loader := spinner.New()
```

kamaln7 (approved):
nice one Davi!

joshuaauerbachwatson (commented):
I haven't reviewed the code in detail but I want to raise a fundamental problem with the change.   I tried it out, and it behaves nicely when I run it in a true terminal window.   However, when I run it in an emacs shell (which is not a full-fledged terminal), or even in emacs `term` (which is more like a real terminal but with some emacs intermediation) the results are confusing.

In emacs shell, when selection from a prompt is required due to multiple namespace matches:  things spew on the window and an informed selection is impossible.   I'm not even going to try to show it here.

In emacs shall, when prompting is not required (an unambiguous namespace hint is provided), the experience is better but the "spinner" messages are annoying and keep the final message from being read easily:

```
ds connect lon
⣽  Loading namespaces ...⣻  Loading namespaces ...⢿  Loading namespaces ...⡿  Loading namespaces ...⣟  Loading namespaces ...⣯  Loading namespaces ...⣷  Loading namespaces ...⣾  Loading namespaces ...⣽  Loading namespaces ...⣻  Loading namespaces ...⢿  Loading namespaces ...Connected to functions namespace '<redacted>' on API host '<redacted>' (label=london)
```

In emacs `term` when a sufficient hint is provided, the result is satisfactory.

In emacs `term` when there is actual prompting, the prompt is impossible to use effectively because the cursor movements don't work as they would on a regular terminal (this might depend on emacs key bindings, not sure).

General points:

- the spinner is IMO unnecessary and less is more there
- the new prompt logic is nice to look at and might be fine in a full function terminal window but what was there before was good enough and by virtue of being minimal was more likely to work in more environments.

I'd be comfortable approving this iff the spinner was dropped and the improved selection experience were somehow guarded by a solid test for being in a real terminal.  When the terminal is not "real", there should be a fallback to the simpler logic.

ddebarros:
> I haven't reviewed the code in detail but I want to raise a fundamental problem with the change. I tried it out, and it behaves nicely when I run it in a true terminal window. However, when I run it in an emacs shell (which is not a full-fledged terminal), or even in emacs `term` (which is more like a real terminal but with some emacs intermediation) the results are confusing.
> 
> In emacs shell, when selection from a prompt is required due to multiple namespace matches: things spew on the window and an informed selection is impossible. I'm not even going to try to show it here.
> 
> In emacs shall, when prompting is not required (an unambiguous namespace hint is provided), the experience is better but the "spinner" messages are annoying and keep the final message from being read easily:
> 
> ```
> ds connect lon
> ⣽  Loading namespaces ...⣻  Loading namespaces ...⢿  Loading namespaces ...⡿  Loading namespaces ...⣟  Loading namespaces ...⣯  Loading namespaces ...⣷  Loading namespaces ...⣾  Loading namespaces ...⣽  Loading namespaces ...⣻  Loading namespaces ...⢿  Loading namespaces ...Connected to functions namespace '<redacted>' on API host '<redacted>' (label=london)
> ```
> 
> In emacs `term` when a sufficient hint is provided, the result is satisfactory.
> 
> In emacs `term` when there is actual prompting, the prompt is impossible to use effectively because the cursor movements don't work as they would on a regular terminal (this might depend on emacs key bindings, not sure).
> 
> General points:
> 
> * the spinner is IMO unnecessary and less is more there
> * the new prompt logic is nice to look at and might be fine in a full function terminal window but what was there before was good enough and by virtue of being minimal was more likely to work in more environments.
> 
> I'd be comfortable approving this iff the spinner was dropped and the improved selection experience were somehow guarded by a solid test for being in a real terminal. When the terminal is not "real", there should be a fallback to the simpler logic.

I updated the `Interactive` to check if in a dumb terminal.

joshuaauerbachwatson (commented):
I tried out the latest version of this in emacs shell.   Result:

```
Error: Namespace is required when running non-interactively
```

This is a regression in terms of UX.   Before the change, I would have gotten the simple plain-text prompt.   I _am_ actually running interactively, albeit in a less than fully functional terminal.    I was assuming that, when a deficient terminal is detected, the behavior would revert to limited plain text interactivity.    This is what commands like `man` do.

I also tried this in `emacs term` (I don't generally use that but others undoubtedly do and it is often given as the solution when people find that things they want to use don't work in `emacs shell`).    The behavior there is unchanged from before the latest change: the behavior is "interactive" but it is difficult and counterintuitive to read the prompts and make a selection due to the way the text is formatted.   IMO the plain text solution would be better there as well.

I don't want to stand in the way of more interactivity in `doctl` (a good thing) and more use of state-of-the-art interaction paradigms, text coloring, etc. _on real terminals_ (a good thing) but I feel our interactivity framework and practices must take into account that there are interactive environments that are not terminals and allow users who use those environments to have an acceptable and non-frustrating experience.

ChiefMateStarbuck:
Thank you for the PR @ddebarros and thank you for the feedback @joshuaauerbachwatson. 

This PR opens up a larger question of how and if Doctl stays up to date with new UI CLI methodologies. At the same time, we cannot allow any breaking changes to Doctl that might increase churn or loss of revenue. Doctl runs on an unknown amount of systems, shells, and environments. With the current priorities of the company and team, we will have to revisit Doctl UI updates.

This is a great starting point @ddebarros and I look to merging this and similar PRs in the future. For today and for the foreseeable future, we will leave this PR open and unmerged.

Issue #1281, opened 2022-10-14
Error: open : The system cannot find the file specified.


SzabKel:
**Describe the Issue:**
[Following the guide here](https://docs.digitalocean.com/reference/doctl/how-to/install/), I downloaded doctl.exe, put it on the path. Running `doctl auth init` does not work however (tried it with powershell, CMD, with and without administrator rights).

It asks for the access token, then throws an error:
![image](https://user-images.githubusercontent.com/3736434/195769195-51fbc5ad-353e-4aea-9776-0dfbba739634.png)

`Error: open : The system cannot find the file specified.`

**Environment:**

 - doctl version: doctl-1.83.0-windows-amd64
 - OS: Windows 10 21H2
 - Installation method: manually downloaded the exe

**Additional Details:**

The `%APPDATA%\doctl\config.yaml` file exists and if I try to auth init with a `--name` it will also save the access token to the config file. However running `auth switch --context <name>` throws the save error. Running `doctl account get` does not work.

Manually specifying the config path does not work as well (the file exists however):
![image](https://user-images.githubusercontent.com/3736434/195771714-bfb08317-631b-43a6-9243-f3df7d04cb91.png)

andrewsomething:
Hi @SzabKel,

Thanks for flagging this for us. I haven't been able to reproduce the problem yet myself.

`%APPDATA%` is a variable that should be expanded. What does `$env:APPDATA` output for you? Are you able to write to that directory?  You should be able to check permission on the file itself with something like: 

    ​(Get-ACL -Path "\path\to\config.yaml").Access | Format-Table IdentityReference,FileSystemRights,AccessControlType,IsInherited,InheritanceFlags -AutoSize

emersonp:
Running into the same error as OP.

My `$env:APPDATA` outputs `C:\Users\parke\AppData\Roaming\`, and within I can find a `doctl\config.yaml`.

The `(Get-ACL`... command throws an error (maybe I'm bad at PowerShell?)
```(Get-ACL -Path "C:\Users\parke\AppData\Roaming\doctl\config.yaml").Access | Format-Table IdentityReference,FileSystemRights,AccessControlType,IsInherited,InheritanceFlags -AutoSize```

```​ 
: The term '​' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the
spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:1 char:1
[...]

Issue #1280, opened 2022-10-13
config.yml not found when executing doctl commands - default path corrupted


lu-roth:
When executing doctl commands I get the following error:
Error: open : no such file or directory

Thats because the default path of the config.yaml is corrupted and the blank space betweet "Application Support" is not escaped like it should be "Application\ Support". 

current: ${HOME}/Library/Application Support/doctl/config.yaml
should be: ${HOME}/Library/Application\ Support/doctl/config.yaml

After adding the "--config" option and setting the config path by myself the doctl commands work but I think its tedious
to do this for every command.

**Environment:**

 - doctl version: 1.83.0
 - OS: macOS Monterey
 - Installation method: Installed from source (doctl-1.83.0-darwin-arm64.tar.gz)

andrewsomething:
Hi @lu-roth,

Thanks for flagging this for us. We have not been able to reproduce the issue ourselves. Can you help by providing some additional information?

- Is the variable `DIGITALOCEAN_CONFIG` set in the environment?
- What shell are you using?
- What terminal emulator are you using?
- Is there anything unique about your environment that we should know?

lu-roth:
Hi @andrewsomething 

- No, DIGITALOCEAN_CONFIG is not set when doing `printenv`
- The `doctl --help` outputs as default for the `--config` flag: `${HOME}/Library/Application Support/doctl/config.yaml`
- I am using `zsh` as shell
- I'm using iTerm2 as terminal emulator (https://iterm2.com/)
- No nothing special about my environment

Other people reported the same issue in a different issue, maybe the comment section will provide you more clarity: 
https://github.com/digitalocean/doctl/issues/604

andrewsomething:
I believe we have tracked down the root of this problem. It should be fixed in the [v1.84.0 release](https://github.com/digitalocean/doctl/releases/tag/v1.84.0) via https://github.com/digitalocean/doctl/pull/1286. Unfortunately, the issue was with the config file itself. So if you have been experiencing this problem, you will need to either delete your existing config file or edit it to remove the stanza that looks like:

```
config:
  set:
    dev-config: ""
```

Please let us know if this has resolved the issue for you.

grantkee:
Still having this issue with version 1.92.1-release installed on MacOS using homebrew

PR #1286, opened 2022-10-18
apps: Nest dev-config settings under 'dev'

andrewsomething:
The `flagName` method that is used when [binding flags to viper](https://github.com/digitalocean/doctl/blob/de1918ad1ee4b1e523c335fa9254e85d415d18aa/commands/doit.go#L221-L230) only supports  nesting a setting three levels deep.

https://github.com/digitalocean/doctl/blob/de1918ad1ee4b1e523c335fa9254e85d415d18aa/commands/doit.go#L287-L292

This causes issues for some deeply nested commands. This is particularly problematic for the `doctl apps dev config set/unset` commands which contain `--dev-config` flag. The resulting setting for it is `config.set.dev-config` which conflicts with the top level `config` setting that should be a string. 

When Viper is init-ed, there is a race to register the setting. Sometimes, it results in the config file containing:

```
config:
  set:
    dev-config: ""
```

Others, it has the appropriate:

    config: /home/<user>/.config/doctl/config.yaml

This breaks things in bad ways causing `doctl auth init` to fail unpredictably. 

This PR forces the `dev-config` setting to be nested more deeply, i.e. `dev.config.set.dev-config`. I don't entirely love this solution, but a more general one would be a breaking change to the configuration file format For example, the settings for DOKS are like `cluster.create.region` rather than `kubernetes.cluster.create.region`. I think we should prioritize getting this specific problem fixed over a broader solution.

Fixes: https://github.com/digitalocean/doctl/issues/1281, https://github.com/digitalocean/doctl/issues/1280

### DIFF
--- a/commands/apps_dev.go
+++ b/commands/apps_dev.go
@@ -36,7 +36,8 @@ import (
 
 const (
 	// AppsDevDefaultEnvFile is the default env file path.
-	AppsDevDefaultEnvFile = ".env"
+	AppsDevDefaultEnvFile     = ".env"
+	appDevConfigFileNamespace = "dev"
 )
 
 // AppsDev creates the apps dev command subtree.
@@ -64,7 +65,7 @@ func AppsDev() *Command {
 		"Build an app component",
 		heredoc.Docf(`
 			[BETA] Build an app component locally.
-			
+
 			  The component name is optional unless running non-interactively.
 
 			  All command line flags as optional. You may specify flags to be applied to the current build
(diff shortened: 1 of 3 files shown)


bentranter (dismissed):
LGTM!

nicktate (dismissed):
good catch and fix on this -- I'm on board with the current solution as a fast fix

kamaln7 (dismissed):

andrewsomething:
Pushed an additional commit to not nest deeper if it is already under `dev.`, e.g. if the `app dev build` command gets a `dev-config` flag.

danaelhe (approved):

bentranter (approved):

scotchneat (approved):
LGTM

Marotenten:
Am waiting for what to do

Issue #1284, opened 2022-10-17
serverless connect + status (and others) fails with Snap


FuegoFro:
**Describe the Issue:**

When installed via snap, seemingly all functions that interact with serverless fail, even after a successful connect. For instance:
```
❯ doctl serverless connect
Connected to functions namespace 'fn-<redacted>' on API host 'https://faas-sfo3-<redacted>.doserverless.co' (label=<redacted>)

❯ doctl serverless status
Error: serverless support is installed but not connected to a functions namespace (use `doctl serverless connect`)
```

Note that I get the same failure for `doctl serverless functions list`

**Environment:**

 - doctl version: 
 ```
doctl version 1.83.0-release
Git commit hash: c9b8dec9
```
 - OS: WSL2 on Windows 11:
```
WSL version: 0.70.0.0
Kernel version: 5.15.68.1
WSLg version: 1.0.45
MSRDC version: 1.2.3575
Direct3D version: 1.606.4
DXCore version: 10.0.25131.1002-220531-1700.rs-onecore-base2-hyp
Windows version: 10.0.22000.1098
```
Running in an Ubuntu 20.04 container:
```
❯ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.1 LTS
Release:        20.04
Codename:       focal
```
 - Installation method: Snap

**Additional Details:**

The `connect` command *does* write the credentials out to `~/.config/doctl/sandbox/creds/<hash>/credentials.json`, but the `status` command is only looking to read them from `/snap/doctl/<number>/opt/sandbox/creds/<hash>/credentials.json`.

The mismatch seems to come from unconditionally using `usualServerlessDir` at
https://github.com/digitalocean/doctl/blob/a1f1c350b96080999048a1b32368ec50a7eb631c/do/serverless.go#L322
and the fact that some things use `serverlessDir` to lookup the creds whereas others use `credsDir`. But even more, it *seems* that `status` should actually be looking in the home dir since the snap dir is only writeable by root (at least for me).

Let me know if there's any other useful information I can provide!

Recent Snap PRs:
- https://github.com/digitalocean/doctl/pull/1173
- https://github.com/digitalocean/doctl/pull/1224

Refactor that introduced the above line:
- https://github.com/digitalocean/doctl/pull/1222

[...]

PR #1287, opened 2022-10-20
Fix bugs in credentials directory management and connection checking

joshuaauerbachwatson:
As reported in issue #1284, it is possible for `serverless connect` to work but subsequent commands to report "not connected".   This only happens when `doctl` is installed via `snap` and only in some system configurations (most seem to work, which is why this has not been noticed before).

There is clearly a snap-specific bug in the handling of the credentials directory.   The type `serverlessService` contains two fields, `serverlessDir` which represents where the plugin is installed, and `credsDir` which represents where credentials should be stored.   Except under `snap`, these are the same, but for `snap` installs they are different.    The fields are initialized correctly.   They are also used correctly in _some_ places but, sadly, not _all_ places.   In particular, the status check for being connected ends up re-forming the credentials directory from the install directory, which is incorrect under snap.

Among the two fixes in this PR is the completion of the program for using `credsDir` consistently everywhere.   As a corollary, the function to check status no longer needs a hashed token argument because this is handled in the initialization of `credsDir`.   That, in turn, results in hitting a bunch of call sites: those changes are trivial (simply dropping the argument).

So, why didn't this break all `snap` users always?    Because, the check for whether the credential store exists was narrowly looking for a "not found" error.   The `snap` assembly had a `creds` directory but it is empty (no leaf directories, no credentials files) and also protected to only be searchable by the user.  Depending on details like the the OS filesystem setup and what privileges the process executing the command has, there might be a "not found" error (which surfaces the bug) or, in many cases, rather, an "access denied" error, which ends up being ignored.   In the latter case, if the code that was guarded by the check ends up using the `credsDir` field correctly, then no harm no foul.    In my case, the ignore path was taken which is why I could not reproduce the error reported in the issue.

This PR also changes the check to be broader: any file system error in accessing the credential store will result in a "not connected" state.   If that had been in place, then the other bug would have broken all `snap` users always and any reasonable test would have found it. 

Testing the adequacy of this fix may involve using a tool like `strace`.   For many/most users, things worked fine before and all you can determine now is that they still work fine.   But, before the change, you will find cases in an `strace` looking at file activity where `credentials.json` is being looked for under `/snap` (improper) and you will see the access denied error (which is subsequently ignored).

### DIFF
--- a/commands/activations_test.go
+++ b/commands/activations_test.go
@@ -95,7 +95,7 @@ func TestActivationsGet(t *testing.T) {
 					}
 				}
 
-				tm.serverless.EXPECT().CheckServerlessStatus(hashAccessToken(config)).MinTimes(1).Return(nil)
+				tm.serverless.EXPECT().CheckServerlessStatus().MinTimes(1).Return(nil)
 				tm.serverless.EXPECT().Cmd("activation/get", tt.expectedNimArgs).Return(fakeCmd, nil)
 				tm.serverless.EXPECT().Exec(fakeCmd).Return(do.ServerlessOutput{}, nil)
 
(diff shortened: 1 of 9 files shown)


andrewsomething (approved):
:+1:  Looks good!

Running `doctl serverless status` under strace installed via snap, I can see: 

    582151 newfstatat(AT_FDCWD, "/snap/doctl/1090/opt/sandbox/creds/f5b778d2/credentials.json", 0xc0002b9ca8, 0) = -1 EACCES (Permission denied)

That demonstrates both the wrong directory being used and the error handling issue. After building a new snap from your fork, I can no longer see that behavior.

PR #1290, opened 2022-10-20
Various improvements to 'doctl serverless'

joshuaauerbachwatson:
This PR merges `feature/serverless` into `main` as part of a plan to stop using the feature branch to stage everything.   We may resurrect it to stage large multi-PR multi-developer changes in the future, but, for now, we will be opening pull requests directly against `main`.

The changes being merged here are all internal-facing.  There is a substantial reduction in the number of commands that use the serverless plugin and a corresponding increase in the number of commands coded natively as part of doctl and going more directly to the appropriate remote endpoints via https.   There are a couple of hidden flags on the `connect` command that allow it to be used with test backends (not intended for normal customer use).   Otherwise, behavior should be preserved.


andrewsomething (approved):
:+1:  LGTM

PR #1292, opened 2022-10-22
Note that --region is required for doctl compute droplet create

kraai:


kraai:
Ah, sorry about that. `dropletCreateLongDesc` says it’s required and I didn’t check. Should I remove it from the list of requirements there or remove that list entirely?

PR #1296, opened 2022-10-24
Use the smaller deployer-only plugin instead of the original one

joshuaauerbachwatson:
The purpose of this change is to make the incorporated serverless plugin smaller and simpler to maintain.  Customers will notice no change, since a plugin still has to be installed to support the `deploy`, `get-metadata`, and `watch` commands.  The download time for the plugin, already not too bad, is not that much improved since it is dominated by the size of the `node` binary.   However, from a long-term code maintenance standpoint the change should be a win.

This change goes along with a corresponding change in the plugin repo:
digitalocean/doctl-sandbox-plugin#6.   Most of the important changes are there.  The present change increments the plugin minimum version and makes a few other minor adjustments.

### DIFF
--- a/commands/serverless-extra.go
+++ b/commands/serverless-extra.go
@@ -168,7 +168,7 @@ func RunServerlessExtraDeploy(c *CmdConfig) error {
 	if err != nil {
 		return err
 	}
-	output, err := RunServerlessExec(projectDeploy, c, []string{flagInsecure, flagVerboseBuild, flagVerboseZip, flagYarn, flagRemoteBuild, flagIncremental},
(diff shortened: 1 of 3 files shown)


joshuaauerbachwatson:
Apologies for withdrawing this back to a draft after requesting reviews.  It's still fine to review it but I belatedly ran the e2e tests (as I should have earlier) and found several problems that will need to be addressed, either in the deployer or in the plugin bridge, before I can be fully confident that this will work.

joshuaauerbachwatson:
This is now passing all the e2e tests that we have, so I am marking it reviewable again.

bentranter (approved):

PR #1297, opened 2022-10-25
Implement sls undeploy in native golang

ddebarros:
Implements `doctl sls undeploy` subtree in goland instead of using the nim plugin.

### DIFF
--- a/commands/serverless.go
+++ b/commands/serverless.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 
 	"github.com/digitalocean/doctl"
+	"github.com/digitalocean/doctl/commands/charm/template"
 	"github.com/digitalocean/doctl/do"
 	"github.com/spf13/cobra"
 )
@@ -116,6 +117,7 @@ the entire packages are removed.`, Writer)
 	AddBoolFlag(undeploy, "packages", "p", false, "interpret simple name arguments as packages")
 	AddBoolFlag(undeploy, "triggers", "", false, "interpret all arguments as triggers")
 	AddBoolFlag(undeploy, "all", "", false, "remove all packages and functions")
+	AddBoolFlag(undeploy, doctl.ArgForce, doctl.ArgShortForce, false, "Delete namespace resources without confirmation prompt")
 	undeploy.Flags().MarkHidden("triggers") // support is experimental at this point
 
 	cmd.AddCommand(Activations())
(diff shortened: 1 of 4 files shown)


do/serverless.go
```diff
@@ -678,6 +681,42 @@ func (s *serverlessService) DeleteNamespace(ctx context.Context, name string) er
 	return err
 }
 
+func (s *serverlessService) CleanNamespace() error {
+	// Deletes all triggers
+	ctx := context.TODO()
+	triggers, err := s.ListTriggers(ctx, "")
+
```
joshuaauerbachwatson, line 688: I understand why errors aren't returned here and the trigger deletion is just skipped.  But, perhaps the reason should be stated in a comment for clarity.

do/serverless.go
```diff
@@ -735,6 +774,62 @@ func (s *serverlessService) ListFunctions(pkg string, skip int, limit int) ([]wh
...
+	err := initWhisk(s)
+	if err != nil {
+		return err
+	}
+
+	if deleteTriggers {
+		ctx := context.TODO()
+		triggers, err := s.ListTriggers(ctx, name)
+
+		if err == nil {
```
joshuaauerbachwatson, line 788: See previous comment about clarifying the reason for interpreting error as empty list.

do/serverless.go
```diff
@@ -678,6 +681,42 @@ func (s *serverlessService) DeleteNamespace(ctx context.Context, name string) er
...
+	}
+
+	for _, fn := range fns {
+		name := fn.Name
+		pkg := strings.Split(fn.Namespace, "/")
+		if len(pkg) == 2 {
+			name = pkg[1] + "/" + fn.Name
+		}
+
+		err := s.DeleteFunction(name, false)
```
joshuaauerbachwatson, line 711: I understand why you're using `false` here; a clarifying comment might be in order though.

joshuaauerbachwatson (commented):
Very nice job @ddebarros .   With the exception of the substantive comment questioning the new `--force` flag and prompt logic, my other comments are nits.   The logic looks good.

commands/serverless.go
```diff
@@ -116,6 +117,7 @@ the entire packages are removed.`, Writer)
 	AddBoolFlag(undeploy, "packages", "p", false, "interpret simple name arguments as packages")
 	AddBoolFlag(undeploy, "triggers", "", false, "interpret all arguments as triggers")
 	AddBoolFlag(undeploy, "all", "", false, "remove all packages and functions")
+	AddBoolFlag(undeploy, doctl.ArgForce, doctl.ArgShortForce, false, "Delete namespace resources without confirmation prompt")
```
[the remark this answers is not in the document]
ddebarros: Fair point, to not introduce a breaking change, I removed the confirmation but left the `--force` flag hidden so we can start introducing it to our e2e tests without errors. 

I also opened an internal ticket for future work.

joshuaauerbachwatson (approved):
👍

andrewsomething (approved):
:+1:  LGTM!

PR #1302, opened 2022-11-03
Add more support for project_id in doctl

ghostlandr:
Specifically, I'm adding support for project ids in two places:

* `doctl apps create` can now take in a `--project-id` parameter. This parameter will be used to assign the created app and resources to the project you specified!
* `doctl apps list` can now take a `--with-projects` flag. Adding this flag will make the `project_id` field populate on each app.

This also bumps the godo version to 1.89

### DIFF
--- a/args.go
+++ b/args.go
@@ -36,6 +36,8 @@ const (
 	ArgActionType = "action-type"
 	// ArgApp is the app ID.
 	ArgApp = "app"
+	// ArgWithProjects will determine whether project ids should be fetched along with listed apps.
+	ArgWithProjects = "with-projects"
 	// ArgAppSpec is a path to an app spec.
 	ArgAppSpec = "spec"
 	// ArgAppLogType the type of log.
--- a/args.go
+++ b/args.go
@@ -36,8 +36,8 @@ const (
 	ArgActionType = "action-type"
 	// ArgApp is the app ID.
 	ArgApp = "app"
-	// ArgWithProjects will determine whether project ids should be fetched along with listed apps.
-	ArgWithProjects = "with-projects"
+	// ArgAppWithProjects will determine whether project ids should be fetched along with listed apps.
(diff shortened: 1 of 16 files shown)


commands/apps.go
```diff
@@ -62,6 +62,7 @@ func Apps() *Command {
 	AddBoolFlag(create, doctl.ArgCommandWait, "", false,
 		"Boolean that specifies whether to wait for an app to complete before returning control to the terminal")
 	AddBoolFlag(create, doctl.ArgCommandUpsert, "", false, "Boolean that specifies whether the app should be updated if it already exists")
+	AddStringFlag(create, doctl.ArgProjectID, "", "", "The id of the project to assign the created app and resources to. If not provided, the default project will be used.")
```
kamaln7, line 65:
```suggestion
	AddStringFlag(create, doctl.ArgProjectID, "", "", "The ID of the project to assign the created app and resources to. If not provided, the default project will be used.")
```

kamaln7 (approved):

args.go
```diff
@@ -36,6 +36,8 @@ const (
 	ArgActionType = "action-type"
 	// ArgApp is the app ID.
 	ArgApp = "app"
+	// ArgWithProjects will determine whether project ids should be fetched along with listed apps.
+	ArgWithProjects = "with-projects"
```
bentranter, line 40: The rest of these app-specific args pseudo-namespaced by containing the word "App" - could we do that here as well for consistency?
```suggestion
	ArgAppWithProjects = "with-projects"
```

bentranter (changes requested):
One tiny nitpick, after that it'll be good to go

args.go
ghostlandr: Oh thanks, I missed that bit of namespacing! Will udpate.

bentranter (approved):

PR #1303, opened 2022-11-07
changes to support usage in serverless backend development

joshuaauerbachwatson:
A corollary to recent efforts to reduce dependence on the serverless plugin:  we are also trying to eliminate internal usages of `nim` in our own development processes, replacing it with `doctl`.   In order for that to work, `doctl` needs some enhancements to work with serverless backend clusters that are not in production (and hence not (yet) known to the portal).

The changes are all in the area of credential management.    Basically, they allow both setting and interrogating credentials via their low-level OpenWhisk characteristics without using the namespace APIs.

Only those commands whose `nim` equivalents are actually used in this way have been changed.   The additional flags are all hidden, since this low-level credential manipulation is not something DigitalOcean customers would ordinarily do (although there is nothing insecure about it).   Having these features show up in help texts would be distracting and might lead to the impression that this feature had an important mainstream use case.

### DIFF
--- a/commands/serverless.go
+++ b/commands/serverless.go
@@ -16,6 +16,7 @@ package commands
 import (
 	"bufio"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


ddebarros (approved):

commands/serverless.go
```diff
@@ -361,6 +373,24 @@ func RunServerlessStatus(c *CmdConfig) error {
 	if err != nil || checkNS != creds.Namespace {
 		return do.ErrServerlessNotConnected
 	}
+	// Before doing the normal display, check for the --credentials flag, which requests return of the
+	// 'creds' value as JSON.
+	credentials, _ := c.Doit.GetBool(c.NS, "credentials")
+	if credentials {
+		type showCreds struct {
+			Auth      string `json:auth`
```
ddebarros, line 381: @joshuaauerbachwatson can we also include the path to the credentials here ? It would be very useful when running the deployer tests.

joshuaauerbachwatson: @ddebarros I have added the path to the credentials output.  To do this cleanly I added to the `ServerlessService` which in turn required regenerating the mocks.  This led me to thinking whether we should expand unit tests to cover these paths, so I may do that also before requesting review again.

andrewsomething (approved):
:+1:  LGTM

ddebarros (approved):

PR #1313, opened 2022-11-17
Support http metric alerts for load balancers

StephenVarela:
Supporting http metric related alerts for load balancers


andrewsomething (approved):
:+1:  LGTM!

PR #1316, opened 2022-11-22
Adds subcommands to enable/disable a trigger

ddebarros:
- Adds subcommands to enable a disable a trigger.

### DIFF
--- a/commands/triggers.go
+++ b/commands/triggers.go
@@ -43,11 +43,8 @@ when events from that source type occur.  Currently, only the ` + "`" + `schedul
 		Writer, aliasOpt("ls"), displayerType(&displayers.Triggers{}))
 	AddStringFlag(list, "function", "f", "", "list only triggers for the chosen function")
 
-	enable := CmdBuilder(cmd, RunTriggerToggle(true), "enable <triggerName>", "Enable a trigger", "Use `doctl serverless triggers enable <triggerName>` to enable a trigger", Writer)
-	AddStringFlag(enable, "output", "o", "", "Desired output format [text|json] (default \"text\")")
-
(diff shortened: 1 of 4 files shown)


commands/triggers.go
```diff
@@ -42,6 +43,12 @@ when events from that source type occur.  Currently, only the ` + "`" + `schedul
 		Writer, aliasOpt("ls"), displayerType(&displayers.Triggers{}))
 	AddStringFlag(list, "function", "f", "", "list only triggers for the chosen function")
 
+	enable := CmdBuilder(cmd, RunTriggerToggle(true), "enable <triggerName>", "Enable a trigger", "Use `doctl serverless triggers enable <triggerName>` to enable a trigger", Writer)
+	AddStringFlag(enable, "output", "o", "", "Desired output format [text|json] (default \"text\")")
+
+	disable := CmdBuilder(cmd, RunTriggerToggle(false), "disable <triggerName>", "Disable a trigger", "Use `doctl serverless triggers disable <triggerName>` to disable a trigger", Writer)
+	AddStringFlag(disable, "output", "o", "", "Desired output format [text|json] (default \"text\")")
```
ddebarros, line 50: doctl has a global `output` flag but I able to get it to work with this command, so I added another flag specific to this command. Any guidance here would be great.

joshuaauerbachwatson (approved):
LGTM 👍 .
I don't know the answer to your query about the output flag.  Someone from the APICLI team will hopefully comment on that.

andrewsomething (approved):
:+1:  LGTM

It doesn't need to block this PR, but it would be great if we could start thinking about writing integration tests for new serverless features using the DO API directly (rather than the plugin) that follow the pattern here:

https://github.com/digitalocean/doctl/tree/main/integration

Unlike the mocked unit tests, these test the actual output of the doctl command.

PR #1317, opened 2022-11-28
function triggers: Use displayer for get sub-command and set displayer type.

andrewsomething:
This changes the triggers get sub-command to use the displayer rather than marshaling JSON itself. By default, it will use the table display and the output flag is supported for JSON. It also sets the displayer type so that the `--format` flag is now supported.

### DIFF
--- a/commands/triggers.go
+++ b/commands/triggers.go
@@ -15,8 +15,6 @@ package commands
 
 import (
 	"context"
-	"encoding/json"
-	"fmt"
 
 	"github.com/digitalocean/doctl"
 	"github.com/digitalocean/doctl/commands/displayers"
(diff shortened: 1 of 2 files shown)


ddebarros:
This may be okay, but `--output=json` prints out a list of triggers, where as the original behavior prints out a single item.

bentranter (approved):

andrewsomething:
@ddebarros  That's true. The intention here is to make this command consistent with other doctl commands. That is how the JSON output is generally formatted. I suppose this might be considered a breaking change, but the trigger commands are all still hidden from the menus/documentation. I'd rather make these consistent now than carry the inconsistency into GA. Are there any beta docs that would need updating?

ddebarros:
> @ddebarros That's true. The intention here is to make this command consistent with other doctl commands. That is how the JSON output is generally formatted. I suppose this might be considered a breaking change, but the trigger commands are all still hidden from the menus/documentation. I'd rather make these consistent now than carry the inconsistency into GA. Are there any beta docs that would need updating?

I think that should be fine. Yes, this feature is still hidden so it shouldn't break anything for the customer.

ddebarros (approved):

PR #1321, opened 2022-12-06
add "singular" aliases for "plural" serverless commands

joshuaauerbachwatson:
This change simply adds aliases, allowing the singular words `activation`, `function`, `namespace` and `trigger` to be used for their plural counterparts.


danaelhe (approved):
thank you! LGTM

PR #1328, opened 2022-12-14
auth: Use OAuth token info endpoint to validate token.

andrewsomething:
Rather than hitting `/v2/account` to validate a token, we should use the OAuth info endpoint. Requests may not be authorized for  `/v2/account` while the OAuth info endpoint only validates the token.


scotchneat (dismissed):

scotchneat (approved):
Looks good. 👍🏽

PR #1336, opened 2023-01-04
commands: Load Balancer updates

anitgandhi:
Only ID is actually required; requiring name and region is inconvenient.


ChiefMateStarbuck (approved):
👍🏾

PR #1337, opened 2023-01-05
Modernize auth init prompt.

andrewsomething:
This change brings some improvements to the `auth init` prompt. It makes use of the input charm to provide feedback to the user as they enter their token. The input is masked but now allows you to see the number of charters you type. It also does some basic input validation. The token is now never echoed back to the terminal.

https://asciinema.org/a/ESnY3RN6Y4XUFg1aKOtiIs5yA


ChiefMateStarbuck (approved):
🥇

scotchneat (dismissed):
Good stuff 👍🏽

scotchneat (approved):

PR #1346, opened 2023-02-08
APICLI-1692: Add backup-restore functionality to db create

danaelhe:
Adds the following command:
`doctl db create new_db_name --restore-from-cluster=original-cluster-name —engine="pg1" —num_nodes=100 —seize=biggest —region="nyc3"`

Note: This does not mimic the control panel behavior. This is parity to the API. To mimic the control panel behavior, we'd do something more like this:
`doctl databases fork new_db_name --restore-from-cluster=original-cluster-name`

which will be in another PR.

### DIFF
--- a/commands/databases.go
+++ b/commands/databases.go
@@ -258,12 +258,10 @@ func buildDatabaseCreateRequestFromArgs(c *CmdConfig) (*godo.DatabaseCreateReque
 			return nil, err
 		}
 		if restoreFromTimestamp != "" {
-			// accepts UTC time format from user (to match db list output) and converts it to ISO8601 for api parity.
-			date, error := time.Parse("2006-01-02 15:04:05 +0000 UTC", restoreFromTimestamp)
-			if error != nil {
-				return nil, fmt.Errorf("Invalid format for --restore-from-timestamp. Must be in UTC format: 2006-01-02 15:04:05 +0000 UTC")
+			dateFormatted, err := convertUTCtoISO8601(restoreFromTimestamp)
+			if err != nil {
+				return nil, err
 			}
-			dateFormatted := date.Format(time.RFC3339)
 			backUpRestore.BackupCreatedAt = dateFormatted
 		}
 		r.BackupRestore = backUpRestore
@@ -272,6 +270,17 @@ func buildDatabaseCreateRequestFromArgs(c *CmdConfig) (*godo.DatabaseCreateReque
 	r.PrivateNetworkUUID = privateNetworkUUID
 
 	return r, nil
+}
+
+func convertUTCtoISO8601(restoreFromTimestamp string) (string, error) {
+	// accepts UTC time format from user (to match db list output) and converts it to ISO8601 for api parity.
+	date, error := time.Parse("2006-01-02 15:04:05 +0000 UTC", restoreFromTimestamp)
+	if error != nil {
+		return "", fmt.Errorf("Invalid format for --restore-from-timestamp. Must be in UTC format: 2006-01-02 15:04:05 +0000 UTC")
+	}
+	dateFormatted := date.Format(time.RFC3339)
+
+	return dateFormatted, nil
 }
 
 // RunDatabaseDelete deletes a database cluster
--- a/commands/databases.go
+++ b/commands/databases.go
@@ -78,6 +78,8 @@ There are a number of flags that customize the configuration, all of which are o
 	AddStringFlag(cmdDatabaseCreate, doctl.ArgDatabaseEngine, "", defaultDatabaseEngine, "The database engine to be used for the cluster. Possible values are: `pg` for PostgreSQL, `mysql`, `redis`, and `mongodb`.")
 	AddStringFlag(cmdDatabaseCreate, doctl.ArgVersion, "", "", "The database engine version, e.g. 14 for PostgreSQL version 14")
 	AddStringFlag(cmdDatabaseCreate, doctl.ArgPrivateNetworkUUID, "", "", "The UUID of a VPC to create the database cluster in; the default VPC for the region will be used if excluded")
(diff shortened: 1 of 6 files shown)


commands/databases.go
```diff
@@ -78,6 +78,8 @@ There are a number of flags that customize the configuration, all of which are o
 	AddStringFlag(cmdDatabaseCreate, doctl.ArgDatabaseEngine, "", defaultDatabaseEngine, "The database engine to be used for the cluster. Possible values are: `pg` for PostgreSQL, `mysql`, `redis`, and `mongodb`.")
 	AddStringFlag(cmdDatabaseCreate, doctl.ArgVersion, "", "", "The database engine version, e.g. 14 for PostgreSQL version 14")
 	AddStringFlag(cmdDatabaseCreate, doctl.ArgPrivateNetworkUUID, "", "", "The UUID of a VPC to create the database cluster in; the default VPC for the region will be used if excluded")
+	AddStringFlag(cmdDatabaseCreate, doctl.ArgDatabaseRestoreFromCluster, "", "", "The name of an existing database cluster from which the backup will be restored.")
+	AddStringFlag(cmdDatabaseCreate, doctl.ArgDatabaseRestoreFromTimestamp, "", "", "The timestamp of an existing database cluster backup in ISO8601 combined date and time format. The most recent backup will be used if excluded.")
```
scotchneat, line 82: Just a note about doclt (not your change).

Unfortunately, doctl lists backups with a different date format. So where this command expects a value like `2023-02-01T17:32:15Z` (which is what the API returns), doctl displays:

```
Size in Gigabytes    Created At
0.039608             2023-02-01 17:32:15 +0000 UTC
0.039649             2023-02-02 17:32:08 +0000 UTC
0.039690             2023-02-03 17:32:15 +0000 UTC
```

Maybe we could add a doctl config setting that would return all dates in ISO8601 format at some point.

commands/databases_test.go
```diff
@@ -36,6 +36,11 @@ var (
 		Hour: "10:00",
 	}
 
+	testBackUpRestore = &godo.DatabaseBackupRestore{
+		DatabaseName:    "sunny-db-cluster",
+		BackupCreatedAt: time.Now().UTC().Format("2006-01-02T15:04:05Z07:00"),
```
scotchneat, line 41: You can use a predefined format here.

```suggestion
		BackupCreatedAt: time.Now().UTC().Format(time.RFC3339),
```

commands/databases.go
scotchneat: It might actually be a good idea to accept the format doctl outputs and convert it to ISO8601 for the user.

integration/database_create_backup_test.go
```diff
@@ -0,0 +1,146 @@
...
+				if err != nil {
+					t.Fatal("failed to dump request")
+				}
+
+				t.Fatalf("received unknown request: %s", dump)
+			}
+		}))
+	})
+
+	when("the minimum required flags are provided", func() {
```
scotchneat, line 78: If we end up doing the date format conversion as mentioned in my previous comment, we should also add a case that verifies the format is converted.

scotchneat (commented):
This looks great. 
I'm happy to approve as is but I left a couple of non-blocking comments for discussion first.

commands/databases.go
danaelhe: Oh good catch! Will accept in UTC format (`2023-02-01 17:32:15 +0000 UTC`) and will convert to ISO861 to pass on to API. Will add to the next commit.

danaelhe:
> This looks great. I'm happy to approve as is but I left a couple of non-blocking comments for discussion first.

Great idea Cesar and good catch- I've updated the implementation to accept UTC format from the user and convert it into ISO861 in the background before passing it to the godo function. Thanks for the thoughtful feedback!

commands/databases.go
```diff
@@ -243,9 +245,44 @@ func buildDatabaseCreateRequestFromArgs(c *CmdConfig) (*godo.DatabaseCreateReque
...
+	}
+
+	r.PrivateNetworkUUID = privateNetworkUUID
+
 	return r, nil
 }
 
+func convertUTCtoISO8601(restoreFromTimestamp string) (string, error) {
+	// accepts UTC time format from user (to match db list output) and converts it to ISO8601 for api parity.
+	date, error := time.Parse("2006-01-02 15:04:05 +0000 UTC", restoreFromTimestamp)
```
scotchneat, line 277: So just for consistency, this can use the `RFC822Z` const format since the time.UTC() method just uses that format and converts the given time to UTC timezone.

```suggestion
	date, error := time.Parse(time.RFC822Z, restoreFromTimestamp)
```

danaelhe: I was reading in a stackoverflow thread that one need to use that specific string date to parse the correct date, as seen below:

Using the `2006-01-02 15:04:05 +0000 UTC` string in `time.Parse()`:
```
	dateString := "2023-02-01 17:32:15 +0000 UTC"
	date, _ := time.Parse("2006-01-02 15:04:05 +0000 UTC", dateString)
	dateFormatted := date.Format(time.RFC3339)

	fmt.Println("The time converted to iso format ", dateFormatted)
```
outputs: 
`The time converted to iso format  2023-02-01T17:32:15Z`


Using the `time.RFC822Z` string in `time.Parse()`:
```
	dateString := "2023-02-01 17:32:15 +0000 UTC"
	date, _ := time.Parse(time.RFC822Z, dateString)
	dateFormatted := date.Format(time.RFC3339)

	fmt.Println("The time converted to iso format ", dateFormatted)
```
outputs: 
`The time converted to iso format  0001-01-01T00:00:00Z`

🤷

scotchneat: 😮 nice catch.

PR #1347, opened 2023-02-08
Bump godo to v1.96.0

danaelhe:
Bump godo to v1.86.0 to support database's promote a read replica (clustered database) to become a primary db cluster


andrewsomething (approved):
:+1:  lgtm!

PR #1348, opened 2023-02-10
Added promote replica support

markpaulson:
Recently, databases added a feature in the API to be able to promote a read replica to be a primary cluster.  This terminates the "replicating connection" to the original primary cluster being replicated and takes the replica and turns it into a stand-alone cluster that can be used as a primary, as could be beneficial in the case where the primary cluster fails and the customer needs to failover to the read replica and to do so requires this ability to promote it.

This PR adds the doctl new `replica promote` command as well as the unit tests required.  At present, I did not see any integration tests around the use of replicas so I did not add an integration test for this new function based on that precedent.


danaelhe:
Thanks for this PR! Could you include an integration test as well? (in` /integrations`)

ChiefMateStarbuck:
Added necessary integration test

danaelhe (approved):
Thanks Sam! Looks great.

PR #1350, opened 2023-02-13
App Platform: fix race on Create/Update --wait flag

jcodybaker:
This PR fixes a potential race which can occur when using the --wait flag on the App Platform create and update operations.  Previously, the wait logic would list deployments to determine the deployment to track. In some cases the wrong deployment would be selected and the --wait would block indefinitely.

APPS-6813

### DIFF
--- a/commands/apps.go
+++ b/commands/apps.go
@@ -490,35 +490,18 @@ func waitForActiveDeployment(apps do.AppsService, appID string, deploymentID str
 			}
 		}
 
-		if deploymentID == "" {
-			deployments, err := apps.ListDeployments(appID)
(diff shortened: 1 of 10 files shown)


andrewsomething (commented):
Nice! Now that we have reliably have an ID for deployments started ny app creates and/or updates, I think we can completely remove this logic that was meant to be a best  effort at  finding the ID.

https://github.com/digitalocean/doctl/blob/490cdfbdb16f979833f101386a3e2b8881688bce/commands/apps.go#L493-L507

kamaln7 (approved):

Issue #992, opened 2021-06-14
doctl databases fork existing cluster


richardwu:
**What is the problem this feature would solve? Please describe.**
I would like to fork an existing database cluster using `doctl`, similar to how you can do this in the web UI.

**Describe the solution you'd like**
Some command of the form `doctl databases fork <existing-name> <new-name> [flags]`

**Additional context**
Would be more convenient than doing a manual copy using postgres.

nibblesnbits:
Any movement on this?  It's the one thing preventing some current automation efforts.

danaelhe:
Addressed in #1352 

Will be available in the next doctl release.

Thank you for your write up and request!

nibblesnbits:
Is there an ETA for that next release?

danaelhe:
We can cut one next week. I’ll let comment here when we do.

danaelhe:
> Is there an ETA for that next release?

We have cut a new release of doctl- v1.93.0:
https://github.com/digitalocean/doctl/releases/tag/v1.93.0

😄

PR #1352, opened 2023-02-28
databases: add "fork" feature

danaelhe:
Add the following to doctl:

`doctl databases fork new_db_name --restore-from-cluster-id=original-cluster-id`

This is similar behavior to the control panel experience. We retrieve the db details and pass them in to the database create request for the user. 

Addresses #992 

APICLI-1856

### DIFF
--- a/args.go
+++ b/args.go
@@ -328,8 +328,10 @@ const (
 	// ArgProjectResource is a flag for your resource URNs
 	ArgProjectResource = "resource"
 
-	// ArgDatabaseRestoreFromCluster is a flag for specifying the name of an existing database cluster from which the backup will be restored.
-	ArgDatabaseRestoreFromCluster = "restore-from-cluster"
+	// ArgDatabaseRestoreFromClusterName is a flag for specifying the name of an existing database cluster from which the backup will be restored.
+	ArgDatabaseRestoreFromClusterName = "restore-from-cluster-name"
+	// ArgDatabaseRestoreFromClusterID is a flag for specifying the id of an existing database cluster from which the backup will be restored.
+	ArgDatabaseRestoreFromClusterID = "restore-from-cluster-id"
 	// ArgDatabaseRestoreFromTimestamp is a flag for specifying the timestamp of an existing database cluster backup in ISO8601 combined date and time format. The most recent backup will be used if excluded.
 	ArgDatabaseRestoreFromTimestamp = "restore-from-timestamp"
 	// ArgDatabaseEngine is a flag for specifying which database engine to use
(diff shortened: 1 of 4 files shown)


args.go
```diff
@@ -328,8 +328,10 @@ const (
 	// ArgProjectResource is a flag for your resource URNs
 	ArgProjectResource = "resource"
 
-	// ArgDatabaseRestoreFromCluster is a flag for specifying the name of an existing database cluster from which the backup will be restored.
-	ArgDatabaseRestoreFromCluster = "restore-from-cluster"
+	// ArgDatabaseRestoreFromClusterName is a flag for specifying the name of an existing database cluster from which the backup will be restored.
+	ArgDatabaseRestoreFromClusterName = "restore-from-cluster-name"
```
danaelhe, line 332: The "restore-from-cluster" flag was introduced in #1346. This would be a breaking change. However, #1346 has not been included in a release yet. Does this still make it a breaking change? If so, I can revert back to "restore-from-cluster"

ChiefMateStarbuck (dismissed):
some small nits. Looking good Dana!

args.go
danaelhe: Ok cool! That's what I was thinking too. Thanks Sam!

commands/databases.go
```diff
@@ -78,8 +78,8 @@ There are a number of flags that customize the configuration, all of which are o
 	AddStringFlag(cmdDatabaseCreate, doctl.ArgDatabaseEngine, "", defaultDatabaseEngine, "The database engine to be used for the cluster. Possible values are: `pg` for PostgreSQL, `mysql`, `redis`, and `mongodb`.")
 	AddStringFlag(cmdDatabaseCreate, doctl.ArgVersion, "", "", "The database engine version, e.g. 14 for PostgreSQL version 14")
 	AddStringFlag(cmdDatabaseCreate, doctl.ArgPrivateNetworkUUID, "", "", "The UUID of a VPC to create the database cluster in; the default VPC for the region will be used if excluded")
-	AddStringFlag(cmdDatabaseCreate, doctl.ArgDatabaseRestoreFromCluster, "", "", "The name of an existing database cluster from which the backup will be restored.")
-	AddStringFlag(cmdDatabaseCreate, doctl.ArgDatabaseRestoreFromTimestamp, "", "", "The timestamp of an existing database cluster backup in UTC combined date and time format (2006-01-02 15:04:05 +0000 UTC). The most recent backup will be used if excluded.")
```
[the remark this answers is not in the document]
danaelhe: I was thinking that the word "existing" would suffice, but since the doc uses "backup" too, I'm going to put it back in. Thanks!

scotchneat (approved):
LGTM 🚀

PR #1327, opened 2022-12-13
Implement tokens sub-commands.

andrewsomething:
Opening as a draft until https://github.com/digitalocean/godo/pull/591 is released.

PR #1369, opened 2023-04-24
Remove beta support for tokens API

bentranter:
Removes the beta support for the tokens API. I should also do a bump to godo, but it looks like the vendored deps are a bit out of date, so I figured it's better to do that in a separate PR.

Reverts [#1327](https://github.com/digitalocean/doctl/pull/1327), except for the `hiddenFlag` helper since it seems useful, and some formatting fixes from that PR.


Anna-Katona:
why?
any plans to implement this function later?

PR #1370, opened 2023-04-25
Update go version and deps

bentranter:
Updates godo, and the Go version in the go.mod file, tests, and release action.


andrewsomething (approved):
:+1:  lgtm 

Think we'll need to make a change to the branch protection rules to land this as well.

PR #1376, opened 2023-05-15
doit: respect json output flag

bentranter:
Updates the `version` command to respect the `-o json` flag. Opening as a draft since this is missing a test and the output is ugly.

Example output:

``` 
$ go run cmd/doctl/main.go version -o json

{
  "version": "doctl version 0.0.0-dev",
  "notification": "release 1.94.0 is available, check it out!"
}
```

### DIFF
--- a/doit.go
+++ b/doit.go
@@ -139,6 +139,37 @@ func (v Version) Complete(lv LatestVersioner) string {
 	return buffer.String()
 }
 
+// CompleteJSON is the complete version for doit, formatted as JSON.
+func (v Version) CompleteJSON(lv LatestVersioner) string {
+	versionInfo := &struct {
+		Version      string `json:"version,omitempty"`
+		Commit       string `json:"commit,omitempty"`
+		Notification string `json:"notification,omitempty"`
+	}{
+		Version: fmt.Sprintf("doctl version %s", v.String()),
+	}
+
+	if v.Build != "" {
+		versionInfo.Commit = fmt.Sprintf("\nGit commit hash: %s", v.Build)
+	}
+
+	if tagName, err := lv.LatestVersion(); err == nil {
+		v0, err1 := semver.Make(tagName)
+		v1, err2 := semver.Make(v.String())
+
+		if len(v0.Build) == 0 {
+			v0, err1 = semver.Make(tagName + "-release")
+		}
+
+		if err1 == nil && err2 == nil && v0.GT(v1) {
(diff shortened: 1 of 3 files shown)


doit.go
```diff
@@ -139,6 +139,37 @@ func (v Version) Complete(lv LatestVersioner) string {
...
+	versionInfo := &struct {
+		Version      string `json:"version,omitempty"`
+		Commit       string `json:"commit,omitempty"`
+		Notification string `json:"notification,omitempty"`
+	}{
+		Version: fmt.Sprintf("doctl version %s", v.String()),
+	}
+
+	if v.Build != "" {
+		versionInfo.Commit = fmt.Sprintf("\nGit commit hash: %s", v.Build)
```
ElanHasson, line 153: should this just be `v.Build`?

doit.go
```diff
@@ -139,6 +139,37 @@ func (v Version) Complete(lv LatestVersioner) string {
...
+	if tagName, err := lv.LatestVersion(); err == nil {
+		v0, err1 := semver.Make(tagName)
+		v1, err2 := semver.Make(v.String())
+
+		if len(v0.Build) == 0 {
+			v0, err1 = semver.Make(tagName + "-release")
+		}
+
+		if err1 == nil && err2 == nil && v0.GT(v1) {
+			versionInfo.Notification = fmt.Sprintf("release %s is available, check it out!", tagName)
```
ElanHasson, line 165: should this be `tagName` ? 

I think we should eliminate the need to do any string parsing 100% for json.

doit.go
bentranter: Yeah, it probably should, I just started by copying this verbatim from the regular output, and then sticking it in fields that roughly made sense. Definitely don't want a newline in there though haha

bentranter:
@ElanHasson Pushed an update so that the output is now,
```json
{
  "version": "0.0.0-dev",
  "commit": "66cd4a38",
  "notification": "release 1.94.0 is available, check it out!"
}
```
but based on your comments, would you rather than the latest tag is available in its own field? I could do something like:
```json
{
  "version": "0.0.0-dev",
  "commit": "66cd4a38",
  "latestTag": "1.94.0",
  "notification": "release 1.94.0 is available, check it out!"
}
```
instead.

ElanHasson:
@bentranter yep!

I think `latestTag` makes sense here, allowing for easier programmatic access to the json.

ChiefMateStarbuck (approved):

PR #1378, opened 2023-05-17
databases: prevent potential panic on db create

bentranter:
In the command to create a database, there's an error that we're currently ignoring. If that err is not nil and the next line runs that tries to access a field on a nil database, doctl will panic. This PR checks that error, which will prevent the panic.


andrewsomething (approved):
:+1:

danaelhe (approved):
nice

ChiefMateStarbuck (approved):

Issue #1375, opened 2023-05-13
Cannot get read-only creds with `doctl registry login`


iansltx:
**Describe the Issue:**

There's no command line flag to log into the DO registry read-only, though that's the default on `doctl registry config`. This means that logins always request read-write access, which fails if doctl is logged in with a read-only token.

**Environment:**

 - doctl version:

doctl version 1.94.0-release
Git commit hash: d51d12cc

 - OS: Ubuntu 22.04 (Docker marketplace image)
 - Installation method: Snap

bentranter:
Hey @iansltx, thanks for the report. Support has been added for logging into the DO registry read-only as of [the most recent release](https://github.com/digitalocean/doctl/releases/tag/v1.96.0).

You can now do,
```sh
$ doctl registry login --read-only tru
```

to log into the registry read-only.

PR #1379, opened 2023-05-18
allow read-only creds for container registry login

bentranter:
Addresses #1375. Adds a new flag to login to the container registry generating a read-only API token.


andrewsomething (approved):
:+1:  Good stuff!

PR #1381, opened 2023-05-26
databases: fix replica promote doc string

bentranter:
Previously, the docs for `replica promote` was the same as that of `replica delete`. This was probably from a copy-paste error. This change fixes that, and updates the command alias so that it's not also `rm` (it's now `p`).

Not sure how I feel about the `p` alias, so if you have any ideas on a better one, let me know and I can change it.


bentranter:
cc @digitalocean/product-docs

danaelhe (approved):

PR #1383, opened 2023-06-09
Doctl

Dadudidas:
Kontrolle ist besser


danaelhe:
Thanks for the PR. We'd like users to report vulnerabilities via Github issues so no need to document otherwise.

PR #1386, opened 2023-06-30
uptime checks: add crud methods

bentranter:
Tried to see how far I could get adding support for uptime checks and alerts in a day, and this is what I could do.

This is still missing:

  * Integration tests,
  * More in-depth tests for the `create` check operation, since I added a bit of extra functionality over the default API interaction,
  * The `get state` method for checks. It has a different displayable type, so it was too much work for today,
  * All of the uptime alerts, since it's a separate set of endpoints.

Not sure if we want to merge as-is, but I figured it's probably better than not having the functionality. I'll try to complete it on the next free Friday.


danaelhe (approved):
👍

commands/uptime_checks.go
```diff
@@ -0,0 +1,205 @@
...
+func UptimeCheck() *Command {
+	cmd := &Command{
+		Command: &cobra.Command{
+			Use:   "uptime",
+			Short: "Display commands to manage uptime checks",
+			Long: `The sub-commands of ` + "`" + `doctl uptime` + "`" + ` manage your uptime checks.
+
+DigitalOcean Uptime Checks provide the ability to monitor your endpoints from around the world,
+and alert you when they're slow, unavailable, or SSL certificates are expiring.`,
+		},
```
andrewsomething, line 37:
```suggestion
			GroupID: manageResourcesGroup,
		},
```

After https://github.com/digitalocean/doctl/pull/1387 landed, we need to remember to `GroupID` to get them categorized right. As is, it shows up under `Additional Commands`.

Also, should this maybe be nested under the existing `monitoring` command? That would mirror the control panel experience.

PR #1388, opened 2023-07-05
registry: add default expiry to login command

bentranter:
Adds a default expiry to the login command. By default, the API tokens created by logging into the registry will now expire after 30 days. To create a token that doesn't expire, you must pass the `--never-expire true` flag.


danaelhe (approved):

danaelhe:
Looks good!

andrewsomething (commented):
Looks good!

One suggestion: Since we are changing the behavior, I think it might be a good idea to print something that makes clear what happens by default so that there is less surprise. Might even be a good way to highlight that you can even set something shorter for CI, etc. 

Something like:

```go
	if expirySeconds == 0 && !neverExpire {
		notice("Login valid for 30 days. Use the --expiry-seconds flag to set a shorter expiration or --never-expire for no expiration.")
	}
```

bentranter:
@andrewsomething Thanks for the suggestion, totally forgot we'd talked about adding that when discussing the change before. I added that message right after the `Logging Docker in...` message appears.

andrewsomething (approved):
:+1:  LGTM!

PR #1393, opened 2023-07-24
Fix typo causing incorrect token highlighting

brandoncc:
I think the triple backtick that I changed is the reason that [this page](https://docs.digitalocean.com/reference/doctl/reference/compute/droplet/create/) has this content:

image


danaelhe (approved):
Thank you and good catch!

Issue #1394, opened 2023-07-24
--access-token flag ignored when also providing --context flag


toby-griffiths:
**Describe the Issue:**

I am attempting to create an auth context with a custom name, with a token from my env vars.

If I use…

```shell
$ doctl auth init --access-token "${DIGITALOCEAN_TOKEN}"
```

… it works just fine, but if I use…

```shell
$ doctl auth init --access-token "${DIGITALOCEAN_TOKEN}" --context=another
```

if prompts for my access token.

**Environment:**
 - doctl version: doctl version 1.97.1-release
 - OS: MacOS 13.2.1 (22D68)
 - Installation method: Homebrew

**Additional Details:**

No trace to provide as it's not making any requests

PR #1395, opened 2023-07-24
auth: init recognizes --access-token

danaelhe:
Addresses #1394

Before:
```
go run cmd/doctl/main.go auth init --access-token "$token" --context=anotherone
Please authenticate doctl for use with your DigitalOcean account. You can generate a token in the control panel at https://cloud.digitalocean.com/account/api/tokens

❯ Enter your access token:
```

After:
```
go run cmd/doctl/main.go auth init --access-token "$token" --context=anotherone
Using token for context anotherone

Validating token... ✔
```

### DIFF
--- a/commands/auth.go
+++ b/commands/auth.go
@@ -146,7 +146,14 @@ To create new contexts, see the help for `+"`"+`doctl auth init`+"`"+`.`, Writer
 // XDG_CONFIG_HOME is not set, use $HOME/.config. On Windows use %APPDATA%/doctl/config.
 func RunAuthInit(retrieveUserTokenFunc func() (string, error)) func(c *CmdConfig) error {
 	return func(c *CmdConfig) error {
-		token := c.getContextAccessToken()
+		token := viper.GetString(doctl.ArgAccessToken)
+
+		// if --access-token is passed, set the context.
+		if token != "" {
+			c.setContextAccessToken(token)
+		}
+
+		token = c.getContextAccessToken()
 		context := strings.ToLower(Context)
 		if context == "" {
 			context = strings.ToLower(viper.GetString("context"))


commands/auth.go
```diff
@@ -146,7 +146,14 @@ To create new contexts, see the help for `+"`"+`doctl auth init`+"`"+`.`, Writer
 // XDG_CONFIG_HOME is not set, use $HOME/.config. On Windows use %APPDATA%/doctl/config.
 func RunAuthInit(retrieveUserTokenFunc func() (string, error)) func(c *CmdConfig) error {
 	return func(c *CmdConfig) error {
-		token := c.getContextAccessToken()
+		token := viper.GetString(doctl.ArgAccessToken)
```
andrewsomething, line 149: Unfortunately this has some unintended consequences.  As the top level `access-token` item in the config file is bound to the flag value, there is no way for us to determine if this is passed on the command line or sourced from the config file for the default context. 

So this breaks initializing a new non-default context without passing the flag. If a default context is already configured, it's token just gets used:

```
$ rm ~/.config/doctl/config.yaml 
$ ./builds/doctl auth init
Please authenticate doctl for use with your DigitalOcean account. You can generate a token in the control panel at https://cloud.digitalocean.com/account/api/tokens

❯ Enter your access token:  ●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●

Validating token... ✔

$ ./builds/doctl auth init --context test
Using token for context test

Validating token... ✔
```

I'm not sure if there is a simple solution here. This may need to wait until we can do deeper changes to how contexts work.

andrewsomething (changes requested):

jerofasion (approved):

PR #1397, opened 2023-08-01
Adds local build support for arm64

jonfriesen:
This change updates the Makefile to introduce arm64 support for local builds.


danaelhe (approved):
was able to test this on my machine and looks good 👍  Ty!

PR #1399, opened 2023-08-01
Adds support for heroku-22 stack

jonfriesen:
This change adds support for locally building app platform apps that use both ubuntu 18 base (default) and also ubuntu-22 stack which is behind an app spec feature flag.


danaelhe (approved):
👍

PR #1405, opened 2023-08-18
Add support for retrying failed requests.

andrewsomething:
This enables support for retrying failed requests. A few things to note:

- I've chosen to expose the `http-retry-max` flag to users to give them insight into the new default behavior. I've marked the `http-retry-wait-max` and `http-retry-wait-min` flags as hidden as the defaults should be fine for most users, but it still allows for configuration if needed.
- For the wait settings, I've used an `int` rather than a `float64` as I feel that it is friendlier on the command line.
- I've used the `http` prefix on these so that the environment variables end up consistent with the same settings ion Terraform, e.g. `DIGITALOCEAN_HTTP_RETRY_MAX`.
- I've bypassed retries for the `doctl account ratelimit` command. The purpose of the command is to return the rate-limit headers regardless of response status. So retrying here just hides the current status and delays returning a response for no real gain.

### DIFF
--- a/commands/account.go
+++ b/commands/account.go
@@ -14,7 +14,10 @@ limitations under the License.
 package commands
 
 import (
+	"fmt"
+
 	"github.com/digitalocean/doctl/commands/displayers"
+	"github.com/digitalocean/doctl/do"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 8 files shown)


bentranter (approved):
LGTM!

PR #1406, opened 2023-08-21
Update docker to v24

fumblehool:
## Issue

Running command; `doctl app dev build`
Gives `Error: attaching to container exec: http: invalid Host header`

## Fix
The https://github.com/moby/moby/issues/45935 issue was recently fixed in https://github.com/moby/moby/pull/45942 so this PR updates the Docker to the latest version.


jcodybaker (approved):

andrewsomething (approved):
:+1:  LGTM

I was able to reproduce the issue with the most recent doctl release. It was build against Go 1.20.7.  It would be great to get some basic coverage for these `app dev` commands into our integration test suite. It would help catch something like this earlier.

PR #1408, opened 2023-08-29
Support LB type param on doctl

asaha2:


commands/load_balancers.go
```diff
@@ -63,6 +63,7 @@ With the load-balancer command, you can list, create, or delete load balancers,
 		fmt.Sprintf("The load balancer's size, e.g.: `lb-small`. Only one of %s and %s should be used", doctl.ArgSizeSlug, doctl.ArgSizeUnit))
 	AddIntFlag(cmdRecordCreate, doctl.ArgSizeUnit, "", 0,
 		fmt.Sprintf("The load balancer's size, e.g.: 1. Only one of %s and %s should be used", doctl.ArgSizeUnit, doctl.ArgSizeSlug))
+	AddStringFlag(cmdRecordCreate, doctl.ArgLoadBalancerType, "", "", "The type of load balancer")
```
andrewsomething, line 66: It would be great if the help string provided a hint of what the acceptable values are. 

If this feature will not be available to all customers for some time still, we may want to consider hiding it from the help output. You could do that with:

```
cmdRecordCreate.Flags().MarkHidden(doctl.ArgLoadBalancerType)
```

commands/load_balancers.go
```diff
@@ -445,6 +446,12 @@ func buildRequestFromArgs(c *CmdConfig, r *godo.LoadBalancerRequest) error {
 	}
 	r.SizeUnit = uint32(sizeUnit)
 
+	lbType, err := c.Doit.GetString(c.NS, doctl.ArgLoadBalancerType)
+	if err != nil {
+		return err
+	}
+	r.Type = lbType
```
andrewsomething, line 453: From the consts in godo, it looks like this value needs to be in all capital letters:

https://pkg.go.dev/github.com/digitalocean/godo#LoadBalancerTypeGlobal

Unless the backend is already doing something similar, it might be nice for users if we do something like:

```suggestion
	r.Type = strings.ToUpper(lbType)
```

commands/load_balancers.go
asaha2: Yep this feature will still remain behind a feature flipper for now: will update as suggested.

commands/load_balancers.go
asaha2: Backend is already doing something similar: ie it accepts case-insensitive versions of `GLOBAL` and `REGIONAL`, and godo consts have been created to make it more intuitive to the users.

asaha2: Addressed in 8c8a89ccaf1ac055d751b98ca380f68330f12f99.

andrewsomething (approved):
:+1:  LGTM

One thing to note that can be addressed in a future PR (perhaps for GA?) is while this allows for setting that option on create, it doesn't display it in the output of the list and get commands. It may be useful information that  users will want if they have LBs of multiple types.

PR #1412, opened 2023-09-01
databases: Support displaying private connection.

andrewsomething:
While doing https://github.com/digitalocean/doctl/pull/1411, I noticed another database feature we do not support well. Database clusters now return connection details for both their public hostnames and private hostnames only accessible via VPC. This adds a `--private` flag to the existing `databases connection` command that will return the private connection details instead of the public ones when passed.


danaelhe (approved):
Nice 🚢

PR #1429, opened 2023-10-05
Alerts: Expand on Monitoring Alerts Doc

danaelhe:
Expanding on `doctl monitoring alert -h` docs to mitigate confusion between `doctl monitoring alert` and `doctl monitoring uptime alert`

### DIFF
--- a/commands/monitoring.go
+++ b/commands/monitoring.go
@@ -55,7 +55,7 @@ func alertPolicies() *Command {
 An alert policy can be applied to resource(s) (currently Droplets)
 in order to alert on resource consumption. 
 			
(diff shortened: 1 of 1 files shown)


mikeberezowski (approved):
Good catch for clarification. Approved. 👍

MrCuPper:
Thanks for solve this case

MrCuPper:
> Good catch for clarification. Approved. 👍

Congratulations 🫶

MrCuPper (commented):
I don't know what you doing about it but it's just resolved with your code, thanks

PR #1448, opened 2023-10-22
docs(cdn): add examples for `compute cdn flush` command

### LINKED ISSUE #1156: Add examples other than [*] to compute cdn flush --files in documentation and/or add output showing which files are flushed

Currently there aren't any examples showing what a path should look like. Does it start with a leading slash? Do you need the bucket name like you do in s3cmd? Can you start with a wildcard to flush everything with a given file extension?

The fact that doctl always returns an empty success makes it even more difficult to determine whether you've entered the path correctly.

Documentation showing a few examples of how you'd format a request, like:

```
doctl compute cdn flush <cdn-id> [/spacename/foldername/*]
doctl compute cdn flush <cdn-id> [*/*.mp4]
doctl compute cdn flush <cdn-id> [/spacename/*.obj]
```

It would be great if doctl could also return a list of the files that had their cache flushed.

Thanks!

rustiever:
aims to fix #1156 

Before: 
``` 
gor cmd/doctl/main.go compute cdn flush -h
This command flushes the cache of a Content Delivery Network (CDN), which:

- purges all copies of the files in the cache
- re-caches the files
- retrieves files from the origin server for any requests that hit the CDN endpoint until all the files are re-cached

This is useful when you need to ensure that files which were recently changed on the origin server are immediately available via the CDN.

To purge specific files, you can use the `--files` flag and supply a path. The path may be for a single file or may contain a wildcard (`*`) to recursively purge all files under a directory. When only a wildcard is provided, or no path is provided, all cached files will be purged.

Usage:
  doctl compute cdn flush <cdn-id> [flags]

```

After:
```
gor cmd/doctl/main.go compute cdn flush -h
This command flushes the cache of a Content Delivery Network (CDN), which:

- purges all copies of the files in the cache
- re-caches the files
- retrieves files from the origin server for any requests that hit the CDN endpoint until all the files are re-cached

This is useful when you need to ensure that files which were recently changed on the origin server are immediately available via the CDN.

To purge specific files, you can use the `--files` flag and supply a path. The path may be for a single file or may contain a wildcard (`*`) to recursively purge all files under a directory. When only a wildcard is provided, or no path is provided, all cached files will be purged.
Examples:
 doctl compute cdn flush <cdn-id>  --files /path/to/assets/*
 doctl compute cdn flush <cdn-id>  --files "/path/to/file.one, /path/to/file.two"
 doctl compute cdn flush <cdn-id>  --files /path/to/file.one --files /path/to/file.two
 doctl compute cdn flush <cdn-id>  --files *

Usage:
  doctl compute cdn flush <cdn-id> [flags]
```

### DIFF
--- a/commands/cdns.go
+++ b/commands/cdns.go
@@ -87,7 +87,12 @@ Currently, you can only update the custom domain and its certificate ID with thi
 
 This is useful when you need to ensure that files which were recently changed on the origin server are immediately available via the CDN.
 
-To purge specific files, you can use the `+"`"+`--files`+"`"+` flag and supply a path. The path may be for a single file or may contain a wildcard (`+"`"+`*`+"`"+`) to recursively purge all files under a directory. When only a wildcard is provided, or no path is provided, all cached files will be purged.`, Writer,
+To purge specific files, you can use the `+"`"+`--files`+"`"+` flag and supply a path. The path may be for a single file or may contain a wildcard (`+"`"+`*`+"`"+`) to recursively purge all files under a directory. When only a wildcard is provided, or no path is provided, all cached files will be purged.
+Examples:		
+ doctl compute cdn flush <cdn-id>  --files /path/to/assets/*
+ doctl compute cdn flush <cdn-id>  --files "/path/to/file.one, /path/to/file.two"
+ doctl compute cdn flush <cdn-id>  --files /path/to/file.one --files /path/to/file.two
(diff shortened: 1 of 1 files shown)


rustiever:
Hi @andrewsomething, sorry for long delay for PR. Pls review and let me know

andrewsomething (approved):
This is great! Thanks! :pray:

PR #1450, opened 2023-10-22
Add app spec validate-offline (Resolves #1449)

### LINKED ISSUE #1449: Ability to validate App Spec Schema in Offline Mode

This feature would solve the following problems & issues:

- No ability to run the following command in "offline" mode. For example:

      doctl app spec validate .do/app.yml --schema-only
      Error: Unable to initialize DigitalOcean API client: access token is required. (hint: run 'doctl auth init')

  - digitalocean/action-doctl#67
  - LyraPhase/docker-phasik.tv#5
- Security Risk: GitHub Actions using [digitalocean/action-doctl][1] for App Schema validation always requires an API `token`
  - The current action & implementation for `doctl app schema validate --schema-only ...` makes it impossible to follow the [principle of least priviledge][2]
  - This prevents [a proposed `no_auth` feature for digitalocean/action-doctl][3] from being currently possible.

[1]: https://github.com/digitalocean/action-doctl
[2]: https://securitylab.github.com/research/github-actions-building-blocks/#following-the-principle-of-least-privilege
[3]: https://github.com/trinitronx/action-doctl/pull/1

Some way to validate a DigitalOcean App Spec in offline mode should be provided.  A separate command (e.g. `validate-offline`) should be decoupled from the `godo` API client & `auth` initialization to support this use case.

**Additional context**

Currently, the `doctl app spec validate` command is tightly coupled to require `godo` init and `doctl auth`.  The `validateCommand` is defined using [`CmdBuilder()`][4], which calls [`cmdBuilderWithInit(..., true, options...)`][5] with [`initCmd`][6] set to [`true` **always**][7] (This is passed to [`NewCmdConfig()`][8] as the [`initGodo` parameter which runs `cmdConfig.initServices()`][9]).

As such, the `--schema-only` flag still requires `doctl auth` and a `godo` client is initialized.  Unfortunately due to the way Cobra commands appear to be set up, the `--schema-only` flag is not available in the context where the command creation happens.  Therefore, it seems that we must implement a new sub-command for this feature (e.g. `doctl app spec validate-offline`). The usage of [`cmdBuilderWithInit` similar to this command PR][10] seems to be the proper method to setup this new command.

[4]: https://github.com/digitalocean/doctl/blob/f6d249bcc18b0712bc8c3048c76addea25ec6b7d/commands/command.go#L66-L68
[5]: https://github.com/digitalocean/doctl/blob/f6d249bcc18b0712bc8c3048c76addea25ec6b7d/commands/command.go#L70
[6]: https://github.com/digitalocean/doctl/blob/f6d249bcc18b0712bc8c3048c76addea25ec6b7d/commands/command.go#L81C5-L81C12
[7]: https://github.com/digitalocean/doctl/blob/f6d249bcc18b0712bc8c3048c76addea25ec6b7d/commands/command.go#L67C75-L67C79
[8]: https://github.com/digitalocean/doctl/blob/f6d249bcc18b0712bc8c3048c76addea25ec6b7d/commands/command_config.go#L78
[9]: https://github.com/digitalocean/doctl/blob/f6d249bcc18b0712bc8c3048c76addea25ec6b7d/commands/command_config.go#L193-L197
[10]: https://github.com/digitalocean/doctl/pull/225

trinitronx:
This PR implements a new sub-command, as described in #1449:

    doctl  apps spec validate-offline

I followed a similar method to #225

When merged, this enables the following GitHub Actions feature to work:

- trinitronx/action-doctl/pull/1
  - Note: this is a draft PR against my fork to demonstrate the changes
  - Fixes digitalocean/action-doctl#67

Notes: My work on this PR was done as part of [DigitalOcean + MLH's Hacktoberfest][1] + [Global Hack Week: Open Source][2]

[1]: https://news.mlh.io/mlh-partners-with-digitaloceans-10th-hacktoberfest-for-global-hack-week-open-source-09-01-2023
[2]: https://ghw.mlh.io/events/open-source


Ultratanques:
wget -qO- https://repos-droplet.digitalocean.com/install.sh | sudo bash

Ultratanques:
true

Ultratanques:
ExecStart=/opt/digitalocean/bin/droplet-agent -syslog -sshd_config="/path/to/your/custom/sshd_config"

Ultratanques:
ExecStart=/opt/digitalocean/bin/droplet-agent -syslog -sshd_port=123

andrewsomething (commented):
Thanks so much for this great contribution! 

You provided a lot of great context in the issue. I also prefer your original design of the `--schema-only` flag being used to determine if the token is needed or not. Reading over the problems you ran into, you are right that the `--schema-only` flag is not available when command creation happens. But we can workaround this! 

We can call `initServices` directly if needed:

```go
	if err := c.initServices(c); err != nil {
		return err
	}
```

So rather than adding a new sub-command we can use the `--schema-only` flag for this by:

1) Creating the command with `cmdBuilderWithInit` like you did with the `validate-offline` one.
1)  [Between the `if schemaOnly` block and the `c.Apps().Propose` call](https://github.com/digitalocean/doctl/blob/f6d249bcc18b0712bc8c3048c76addea25ec6b7d/commands/apps.go#L781-L796), add the call to `initServices`

This way, if the `--schema-only` flag is passed, we exit before initializing the services. If not, we still initialize the services and check for the API token before making the API call.

trinitronx:
> Thanks so much for this great contribution!
> 
> You provided a lot of great context in the issue. I also prefer your original design of the `--schema-only` flag being used to determine if the token is needed or not. Reading over the problems you ran into, you are right that the `--schema-only` flag is not available when command creation happens. But we can workaround this!
> 
> We can call `initServices` directly if needed:
> 
> ```
> [...SNIP...]
> ```

Thanks for that suggestion! I've refactored the code to use that method, and tested it against some proposed changes to `action-doctl`.  It's now tested & validated as working!

This is ready for review again :grinning:

andrewsomething (approved):
:+1:  LGTM! Thanks again!

PR #1452, opened 2023-10-25
Load balancer update clarification

dbrian57:
This PR adds additional clarity to the `doctl compute load-balancer update` command.


andrewsomething (approved):
Thanks!

PR #1454, opened 2023-10-26
Support scalable storage for PostgreSQL/MySQL databases

dweinshenker:
**What**

- Adds the ability to provide `--storage-size-mib` when doing CreateCluster or ResizeCluster operations on a database
- Includes the StorageSizeMib as a column in the databases display output

### DIFF
--- a/commands/displayers/database.go
+++ b/commands/displayers/database.go
@@ -45,6 +45,7 @@ func (d *Databases) Cols() []string {
 			"Region",
 			"Status",
 			"Size",
+			"StorageMib",
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


commands/displayers/database.go
```diff
@@ -59,34 +60,37 @@ func (d *Databases) Cols() []string {
...
-			"Size":     "Size",
+			"ID":         "ID",
+			"Name":       "Name",
+			"Engine":     "Engine",
+			"Version":    "Version",
+			"NumNodes":   "Number of Nodes",
+			"Region":     "Region",
+			"Status":     "Status",
+			"Size":       "Size",
+			"StorageMib": "StorageMib",
```
andrewsomething, line 78: The "short" version is used when listing all databases. It provides a bit less information, but the column names should match the "full" version below.

commands/displayers/database.go
```diff
@@ -59,34 +60,37 @@ func (d *Databases) Cols() []string {
...
-		"Created":  "Created At",
+		"ID":         "ID",
+		"Name":       "Name",
+		"Engine":     "Engine",
+		"Version":    "Version",
+		"NumNodes":   "Number of Nodes",
+		"Region":     "Region",
+		"Status":     "Status",
+		"Size":       "Size",
+		"StorageMib": "Storage Size - MiB",
```
andrewsomething, line 91: I wonder if we could truncate this just a bit? This table is already a bit big for most terminals, so saving a few characters might be nice. 

```suggestion
		"StorageMib": "Storage (MiB)",
```

How does that look to you?

andrewsomething (changes requested):
Looks great. Just a bit of a nitpick about the displayer.

commands/displayers/database.go
dweinshenker: this looks good to me, thanks for the additional context on matching short/long versions. I made the changes @andrewsomething

andrewsomething (approved):
LGTM!

PR #1466, opened 2023-12-04
Add check to prevent focused tests being committed.

andrewsomething:
This helps to prevent merging branches with tests that have been marked `Focused`. Marking a test `Focused` can be useful in development, but allowing a test to be checked in to main while focused means other tests are not actually run in CI.


bentranter (approved):
Nice solution, LGTM!

PR #1470, opened 2023-12-07
Resolve conflict with alert configurations by overriding command namespace

andrewsomething:
This resolves an issue with the `doctl monitoring alerts create/update` commands introduced when we added the uptime alert commands. The way doctl's config file currently works, settings can only be three levels deep. This led to the alert commands conflicting with each other.  For example, the configuration for the `type` flag for both ended up at `alert.create.type`. 

This PR resolves the problem adding a `cmdOption` to override the config namespace for a command without needing to change the sub-command layout. The monitoring alert config keeps its existing location (`alert.create.<flag>`) while the uptime alerts are now at `uptime-alert.create.<flag>`.


danaelhe (approved):
good idea!! 🚀

PR #1480, opened 2023-12-26
Docs audit: Firewalls

dbrian57:
Updates and clarifies doc strings for Firewall management commands. This also adds examples to each command.


danaelhe (approved):
thank you!

PR #1483, opened 2023-12-26
Docs audit: Images and Image Actions

dbrian57:
Updates and clarifies the doc strings for the `image` and `image-action` commands. This also adds examples for each command.


danaelhe (approved):
thank you!

PR #1484, opened 2023-12-26
Docs Audit: Invoices

dbrian57:
Updates and clarifies doc strings for Invoice command. Additionally, this adds examples for each of those commands.


danaelhe (approved):
thank you!

PR #1487, opened 2023-12-29
Docs audit: Regions and Sizes

dbrian57:
This updates the doc string for the region and size commands, and adds an example for each.


andrewsomething (approved):
:+1:

PR #1493, opened 2023-12-31
Docs audit: Tags

dbrian57:
Updates and clarifies doc strings for Tag commands. Also add examples for each command.


commands/tags.go
```diff
@@ -31,33 +31,36 @@ func Tags() *Command {
 			Short: "Display commands to manage tags",
 			Long: `The sub-commands of ` + "`" + `doctl compute tag` + "`" + ` manage the tags on your account.
 
-A tag is a label that can be applied to a resource (currently Droplets, images,
-volumes, volume snapshots, and database clusters) in order to better organize or
```
andrewsomething, line 35: The fact that only Droplets, images, volumes, volume snapshots, and database clusters are taggable resources seems like useful info that we should retain somewhere. Maybe in the `apply` command's help?

andrewsomething (commented):
Looks good, but I did have one question inline.

commands/tags.go
dbrian57: You're absolutely right. I'll work that back in there.

andrewsomething (approved):
:pray: Thanks Dan!

PR #1496, opened 2024-01-02
LBAAS-2899: doctl not removing LB firewall rules on update

jvasilevsky:
Currently, when you pass an empty value to the the load balancer allow/deny rules flag we are not updating the LB firewall rules. To match the behavior of our public API we should allow an empty flag allow/deny rules flag to remove the firewall rules from the LB.

At current passing an empty flag does not populates the LB firewall field. We should be separating the behavior of the allow/deny rules flag not being specified and the flag being specified but empty

This change allows an empty value in the fw rules flags to update the LB firewall to empty.
When not specifying the fw rules flags the LB firewall will not be updated.

### DIFF
--- a/commands/load_balancers.go
+++ b/commands/load_balancers.go
@@ -125,9 +125,9 @@ With the load-balancer command, you can list, create, or delete load balancers, 
 		"disable automatic DNS record creation for Let's Encrypt certificates that are added to the load balancer")
 	AddStringFlag(cmdRecordUpdate, doctl.ArgProjectID, "", "",
 		"Indicates which project to associate the Load Balancer with. If not specified, the Load Balancer will be placed in your default project.")
-	AddStringSliceFlag(cmdRecordUpdate, doctl.ArgAllowList, "", []string{},
+	AddStringSliceFlag(cmdRecordUpdate, doctl.ArgAllowList, "", nil,
 		"A comma-separated list of ALLOW rules for the load balancer, e.g.: `ip:1.2.3.4,cidr:1.2.0.0/16`")
-	AddStringSliceFlag(cmdRecordUpdate, doctl.ArgDenyList, "", []string{},
+	AddStringSliceFlag(cmdRecordUpdate, doctl.ArgDenyList, "", nil,
 		"A comma-separated list of DENY rules for the load balancer, e.g.: `ip:1.2.3.4,cidr:1.2.0.0/16`")
 
 	CmdBuilder(cmd, RunLoadBalancerList, "list", "List load balancers", "Use this command to get a list of the load balancers on your account, including the following information for each:\n\n"+lbDetail, Writer,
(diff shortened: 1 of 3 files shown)


guptado (approved):
LGTM

guptado:
can you share how are we testing it out?

commands/load_balancers.go
```diff
@@ -557,17 +557,17 @@ func buildRequestFromArgs(c *CmdConfig, r *godo.LoadBalancerRequest) error {
...
 	}
 
-	denyRules, err := c.Doit.GetStringSlice(c.NS, doctl.ArgDenyList)
+	denyRules, err := c.Doit.GetStringSlicePtr(c.NS, doctl.ArgDenyList)
 	if err != nil {
 		return err
 	}
 
-	if len(allowRules) > 0 || len(denyRules) > 0 {
+	if allowRules != nil || denyRules != nil || len(allowRules) > 0 || len(denyRules) > 0 {
```
asaha2, line 570: Nit: length of `nil` will still be `0`. 
```suggestion
	if len(allowRules) > 0 || len(denyRules) > 0 {
```

asaha2 (approved):

commands/load_balancers.go
bbassingthwaite: I don't think this logic is quite right either. If the `Firewall` object that is supplied to the API is missing, we will not update the rules. So we need to construct a zero Firewall struct when `allowRules` or `denyRules` has been set but is empty. I think there are three examples when a customer updates a firewall:

1) Neither `--allow-list` or `--deny-list` is provided, in this case we should have a `nil` Firewall which would ensure that we don't update the firewall.
2) `--allow-list` or `--deny-list` is provided but no values are provided. In this case, we should send along a zero Firewall object`
3) `--allow-list` or `--deny-list` is provided with values and we should pass a Firewall object with them provided.

I think `GetStringSlicePtr` will need to return a bool to indicate that the param was provided on the CLI call.

jvasilevsky: > 1. Neither --allow-list or --deny-list is provided, in this case we should have a nil Firewall which would ensure that we don't update the firewall.

If the `--allow-list `or `--deny-list` is not provided the `GetStringSlicePtr` should return nil which should ignore updating the firewall.

> 2. --allow-list or --deny-list is provided but no values are provided. In this case, we should send along a zero Firewall object`

If the `--allow-list `or `--deny-list` is provided but empty the `GetStringSlicePtr` should return an empty slice that is defined which will pass the `!= nil` check causing the firewall to be updated

> 3. --allow-list or --deny-list is provided with values and we should pass a Firewall object with them provided.

If the `--allow-list `or `--deny-list` is provided the `GetStringSlicePtr` should return a populated slice that is defined which will pass the `!= nil` check causing the firewall to be updated

The original GetStringSlice always returns at a minimum a defined but empty slice. The GetStringSlicePtr allows a nil slice to return when the flag is not provided. This allows us to differentiate between a defined by empty slice and a nil slice. I might be missing something.

jvasilevsky:
Add allow firewall rule
```
doctl compute load-balancer update a5faf024-bb6d-4b68-80c9-c987d504783d --name="nyc3-load-balancer-01" --vpc-uuid="32e26fa6-86b7-45d8-9719-1bc4b7edbed4" --size="lb-small" --size-unit="1" --region="nyc3" --forwarding-rules="entry_protocol:http,entry_port:80,target_protocol:http,target_port:80 entry_protocol:https,entry_port:443,target_protocol:https,target_port:443,tls_passthrough:true" --allow-list="cidr:1.2.0.0/16"
```

image


No firewall flag specified: No action
```
doctl compute load-balancer update a5faf024-bb6d-4b68-80c9-c987d504783d --name="nyc3-load-balancer-01" --vpc-uuid="32e26fa6-86b7-45d8-9719-1bc4b7edbed4" --size="lb-small" --size-unit="1" --region="nyc3" --forwarding-rules="entry_protocol:http,entry_port:80,target_protocol:http,target_port:80 entry_protocol:https,entry_port:443,target_protocol:https,target_port:443,tls_passthrough:true"
```

image

Empty allow flag: remove firewall rules
```
doctl compute load-balancer update a5faf024-bb6d-4b68-80c9-c987d504783d --name="nyc3-load-balancer-01" --vpc-uuid="32e26fa6-86b7-45d8-9719-1bc4b7edbed4" --size="lb-small" --size-unit="1" --region="nyc3" --forwarding-rules="entry_protocol:http,entry_port:80,target_protocol:http,target_port:80 entry_protocol:https,entry_port:443,target_protocol:https,target_port:443,tls_passthrough:true" --allow-list=""
```
image

andrewsomething (approved):
:+1:  LGTM!

Issue #1003, opened 2021-07-07
`doctl compute droplet create --wait` is not waiting long enough 


samuelstroschein:
**Describe the Issue:**

I am using GitHub workflows to provision new droplets. After creating the droplet and trying to SSH into it, I always got a connection refused' error. Trying to SSH into the newly created droplet on my local machine worked. Therefore, I thought it had to do with the GitHub environment. Turns out the --wait flag simply does not wait till the server is ready to accept a connection which I expected. It might be no bug but either clearly mention it in the documentation or change the --wait flag to test whether the droplet is ready to accept a connection. The workaround is to naively wait for X seconds after the droplet has been created. 

**Environment:**

 - digitalocean/action-doctl@v2.1.0
 - runs-on: ubuntu-latest

danaelhe:
Hi there,

Thank you for the write up. We merged #1498  PR that offers a work around with the `doctl compute ssh <droplet-id>`command by adding a `--ssh-retry-max` flag . This flag will retry for a successful connection to the droplet a certain amount of times before returning control to the user. I tested it out with a bash script and it consistently worked well. If you need anything else, please open another issue.

PR #1498, opened 2024-01-04
ssh: add retries

danaelhe:
A work around for #1003

Passing in the `--wait` flag for `doctl compute droplet create` will return to user before the droplet is ready to accept ssh connections. This PR adds a `--ssh-retry-max` flag for `doctl compute ssh` command to retry for a successful connection to the droplet a certain amount of times before returning to user:

An example of what a CI script would look like:
```
#!/bin/bash
DROPLET_ID=$(go run cmd/doctl/main.go compute droplet create --image ubuntu-20-04-x64 --size s-1vcpu-1gb --region nyc1 wb2 --ssh-keys 40563907 --wait -o json | jq '.[] | .id')
go run cmd/doctl/main.go compute ssh $DROPLET_ID --ssh-retry-max 15  
doctl compute droplet delete $DROPLET_ID
```

### DIFF
--- a/commands/ssh.go
+++ b/commands/ssh.go
@@ -21,7 +21,6 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
-	"time"
 
 	"github.com/digitalocean/doctl"
 	"github.com/digitalocean/doctl/do"
@@ -53,7 +52,7 @@ You may specify the user to login with by passing the `+"`"+`--%s`+"`"+` flag. T
 	AddBoolFlag(cmdSSH, doctl.ArgsSSHAgentForwarding, "", false, "Enable SSH agent forwarding")
 	AddBoolFlag(cmdSSH, doctl.ArgsSSHPrivateIP, "", false, "SSH to Droplet's private IP address")
 	AddStringFlag(cmdSSH, doctl.ArgSSHCommand, "", "", "Command to execute on Droplet")
-	AddBoolFlag(cmdSSH, doctl.ArgCommandRetry, "", false, "Retry for a successful SSH connection with Droplet")
+	AddIntFlag(cmdSSH, doctl.ArgSSHRetryMax, "", 0, "Max number of retries for a successful SSH connection to a Droplet (default is 0)")
 
 	return cmdSSH
 }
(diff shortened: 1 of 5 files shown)


commands/ssh.go
```diff
@@ -50,6 +53,7 @@ You may specify the user to login with by passing the `+"`"+`--%s`+"`"+` flag. T
 	AddBoolFlag(cmdSSH, doctl.ArgsSSHAgentForwarding, "", false, "Enable SSH agent forwarding")
 	AddBoolFlag(cmdSSH, doctl.ArgsSSHPrivateIP, "", false, "SSH to Droplet's private IP address")
 	AddStringFlag(cmdSSH, doctl.ArgSSHCommand, "", "", "Command to execute on Droplet")
+	AddBoolFlag(cmdSSH, doctl.ArgCommandRetry, "", false, "Retry for a successful SSH connection with Droplet")
```
andrewsomething, line 56: I wonder if it would make sense to just use the global `--http-retry-max` flag? On the one hand conflating HTTP and SSH retries might be a bit unintuitive. On the other, it's would be respecting the general intention to retry transient failures.

commands/ssh.go
```diff
@@ -156,7 +160,32 @@ func RunSSH(c *CmdConfig) error {
 		return errors.New("Could not find Droplet address")
 	}
 
+	retry, err := c.Doit.GetBool(c.NS, doctl.ArgCommandRetry)
+	if err != nil {
+		return err
+	}
+
 	runner := c.Doit.SSH(user, ip, keyPath, port, opts)
```
andrewsomething, line 168: It might be interesting to push the retry logic into the SSH runner itself. You can pass configuration options on the command line to SSH with the `-o` flag, e.g. `-o ConnectionAttempts=5`

> **ConnectionAttempts**
> Specifies the number of tries (one per second) to make before exiting. The argument must be an integer. This may be  useful in scripts if the connection sometimes fails. The default is 1.

https://linux.die.net/man/5/ssh_config

SSH itself probably has a a better understanding of what error conditions are retryable.

danaelhe: Oh great catch. Pushing it down to the SSH runner is working out well actually. Thanks for the suggestion!

commands/ssh.go
danaelhe: Ya I can see that. We could meet somewhere in the middle with `ssh-retry-max`. That makes it also consistent with the rest of the `ssh` flags i.e. `ssh-keys`,  `ssh-agent-forwarding` , `ssh-private-ip` etc.

andrewsomething (approved):
:+1:  LGTM!

PR #1499, opened 2024-01-25
droplet create: assign to project after create

### LINKED ISSUE #630: create droplet in a specific project
digitalocean can group droplets into "projects". 

"doctl compute droplet create" will create droplets into the default project.

how can I create a droplet in a specific project?

bentranter:
Fixes #630

Adds a new optional flag to the `doctl compute droplet create` command, `--project-id`, that allows you to pass a project UUID. If a project UUID is passed, that Droplet will be assigned to that project after it is created.

This pattern of create-then-assign can probably be expanded to any resource that returns an URN, but since we've received quite a few requests for this functionality for Droplets, I figured this is a good starting point.


danaelhe (approved):
Nice 👍

andrewsomething (approved):

PR #1504, opened 2024-02-05
Update cnb builder image

ShivaniKumar1:
https://do-internal.atlassian.net/browse/APPS-8334
Update cn builder image to new apps-images tag


andrewsomething (approved):
:+1:  LGTM

PR #1510, opened 2024-03-06
kubernetes: remove backquotes from flag descriptions

bhcleek:
Remove backquotes from flag descriptions in the kubernetes subcommands, because those backquotes cause pflag's usage parser to extract them and treat them as a flag name to try to determine the data type of the flag. The practical effect of this was that the help was borderline unreadable because the description was forced so far right and narrow.

### DIFF
--- a/commands/kubernetes.go
+++ b/commands/kubernetes.go
@@ -264,46 +264,46 @@ Retrieves a list of slugs representing Kubernetes upgrade versions you can use t
 
 	cmdKubeClusterCreate := CmdBuilder(cmd, k8sCmdService.RunKubernetesClusterCreate(defaultKubernetesNodeSize,
 		defaultKubernetesNodeCount), "create <name>", "Create a Kubernetes cluster", `
-Creates a Kubernetes cluster given the specified options and using the specified name. Before creating the cluster, you can use `+"`"+`doctl kubernetes options`+"`"+` to see possible values for the various configuration flags.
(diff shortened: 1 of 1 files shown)


andrewsomething:
Requested a review from @dbrian57 for impact on pdocs.

andrewsomething:
This does improve on the existing experience on the CLI. Top pane is the most recent release and the bottom one is built from this PR:

image

commands/kubernetes.go
```diff
@@ -264,46 +264,46 @@ Retrieves a list of slugs representing Kubernetes upgrade versions you can use t
 
 	cmdKubeClusterCreate := CmdBuilder(cmd, k8sCmdService.RunKubernetesClusterCreate(defaultKubernetesNodeSize,
 		defaultKubernetesNodeCount), "create <name>", "Create a Kubernetes cluster", `
-Creates a Kubernetes cluster given the specified options and using the specified name. Before creating the cluster, you can use `+"`"+`doctl kubernetes options`+"`"+` to see possible values for the various configuration flags.
+Creates a Kubernetes cluster given the specified options and using the specified name. Before creating the cluster, you can use doctl kubernetes options to see possible values for the various configuration flags.
```
timoreimann, line 267: Could we use regular quotes instead of backtick quotes to retain the command delimiters?

bhcleek:
@timoreimann I refactored and put up #1511 as an alternative PR that I think is a better change, because it provides an even more useful usage message without dropping the backquotes around some commands.

andrewsomething:
Closing as this was superseded by https://github.com/digitalocean/doctl/pull/1511

PR #1511, opened 2024-03-08
kubernetes: improve usage message

bhcleek:
Ensure that the first backquoted string in kubernetes create flags is a string that should be used as the type in the usage message, because those backquotes cause pflag's usage parser to extract them and treat them as a flag name to try to determine the data type of the flag. The practical effect of this was that the help was borderline unreadable because the description was forced so far right and narrow.

Here's a screenshot of the effect of these changes:

image


andrewsomething (approved):
:+1:  I like it!

timoreimann (approved):

Issue #1516, opened 2024-03-21
doctl kubernetes cluster node-pool create flave-development --auto-scale --min-nodes 0 --name test --size s-1vcpu-2gb --max-nodes 1 --count 0 says the node pool needs at least min 1


h0jeZvgoxFepBQ2C:
In your documentation (https://docs.digitalocean.com/reference/doctl/reference/kubernetes/cluster/node-pool/create/) you write:

> The minimum number of nodes in the node pool when autoscaling is enabled
> Default: 0

wheres when I write following command: 

```shell
doctl kubernetes cluster node-pool create mycluster123 --auto-scale --min-nodes 0 --name test --size s-1vcpu-2gb --max-nodes 1 --count 0
```

It returns me:
```
Error: POST https://api.digitalocean.com/v2/kubernetes/clusters/123123-123123123-123123123-123123/node_pools: 422 (request "123-123-123-123") validation error: worker_node_pool_spec.min nodes must be greater than or equal to 1
```

Is it not possible to create a nodepool with 0 min nodes? We have a pretty valid usecase: we have some special tasks, which needs huge amounts of RAM/disk space, but we dont want to spend money on a node, which is not utilized at all, when the task is not running. 
So we would like to have a nodepool with 0 min-nodes, which then autoscales when a job is running - and deletes the node again afterwards.

danaelhe:
Thank you for this write up.

Your use case sounds very valid, but unfortunately our Kubernetes product doesn't support scale-to-zero on any node pools as of right now. I've relayed your use case and feedback to our internal Kubernetes team. 

You can only set a minimum number of nodes to 0 if `auto-scale` is set to false. I'm going to update the doctl documentation to include this vital piece of information.

MrCuPper:
Actually I don't know what I does,
I just interesting programming experience but I couldn't get required
construction for learning or pay for college, so I just trying to make
something at all,

PR #1517, opened 2024-03-21
K8s: Update Node Pool documentation

danaelhe:
Addressing #1516

Adding explicit documentation that minimum node pool size can't not be zero if autoscale is true.


commands/kubernetes.go
```diff
@@ -301,8 +301,8 @@ Format: `+"`"+`"name=your-name;size=size_slug;count=5;tag=tag1;tag=tag2;label=ke
 - `+"`"+`tag`+"`"+`: A comma-separated list of tags to apply to nodes in the pool
 - `+"`"+`label`+"`"+`: A label in `+"`"+`key=value`+"`"+` notation. Repeat to add multiple labels at once.
 - `+"`"+`taint`+"`"+`: Taint in `+"`"+`key=value:effect`+"`"+` notation. Repeat to add multiple taints at once.
-- `+"`"+`auto-scale`+"`"+`: Enables cluster auto-scaling on the node pool (boolean)
-- `+"`"+`min-nodes`+"`"+`: The minimum number of nodes that the cluster can be auto-scaled to.
+- `+"`"+`auto-scale`+"`"+`: Enables cluster auto-scaling on the node pool (boolean). Scale-to-zero is not supported.
```
danaelhe, line 304: Not sure if this is necessary.

andrewsomething (commented):
Looking at https://github.com/digitalocean/doctl/issues/1516, they're referencing the `node-pool create` sub-command. That's where the default of `0` is specified in the docs.

https://github.com/digitalocean/doctl/blob/338e5dc629ccfa02ed0626ed88d7b9ca09f86bba/commands/kubernetes.go#L492

So we probably want to update there as well.

danaelhe:
> Looking at #1516, they're referencing the `node-pool create` sub-command. That's where the default of `0` is specified in the docs.
> 
> https://github.com/digitalocean/doctl/blob/338e5dc629ccfa02ed0626ed88d7b9ca09f86bba/commands/kubernetes.go#L492
> 
> So we probably want to update there as well.

good catch 👍

andrewsomething (approved):

PR #1518, opened 2024-03-25
Update cnb run-image tag

fumblehool:
https://do-internal.atlassian.net/browse/APPS-8465

The run-image being used in `doctl apps dev build` is outdated.
Local builds are failing with: `failed to export: get run image top layer SHA: image <image-1> has no layers`
This PR updates the tag to fix the issue.

### DIFF
--- a/commands/apps_dev.go
+++ b/commands/apps_dev.go
@@ -516,8 +516,8 @@ func appDevPrepareEnvironment(ctx context.Context, ws *workspace.AppDev, cli bui
 		}
 
 		// TODO: get stack run image from builder image md after we pull it, see below
-		images = append(images, "digitaloceanapps/apps-run:heroku-18_db5978a")
-		images = append(images, "digitaloceanapps/apps-run:heroku-22_db5978a")
+		images = append(images, "digitaloceanapps/apps-run:heroku-18_df7e351")
+		images = append(images, "digitaloceanapps/apps-run:heroku-22_df7e351")
 	}
 
 	if componentSpec.GetType() == godo.AppComponentTypeStaticSite {


MrCuPper:
hi thanks for your update, but there is another issue happening that it's
noticed I don't have access to it when I try to login to my account

commands/apps_dev.go
```diff
@@ -516,8 +516,8 @@ func appDevPrepareEnvironment(ctx context.Context, ws *workspace.AppDev, cli bui
 		}
 
 		// TODO: get stack run image from builder image md after we pull it, see below
-		images = append(images, "digitaloceanapps/apps-run:heroku-18_db5978a")
-		images = append(images, "digitaloceanapps/apps-run:heroku-22_db5978a")
+		images = append(images, "digitaloceanapps/apps-run:heroku-18_df7e351")
+		images = append(images, "digitaloceanapps/apps-run:heroku-22_df7e351")
```
andrewsomething, line 520: Looking at the available tags for this image, even the new ones are four months old. Should we consider using a new one? Is there a changelog for the image? 

https://hub.docker.com/r/digitaloceanapps/apps-run/tags

fumblehool: Actually, the `apps-run` image is pinned with `cnb-local-image` https://github.com/digitalocean/doctl/blob/32566d9ea889b07dd9cb1e2e07b537633028a688/internal/apps/builder/cnb.go#L29

i.e. `cnb-local-image v0.65.0` uses `apps-run image: df7e351`.

The newer tags are for unreleased dev versions (which are still unstable).

ShivaniKumar1 (approved):

commands/apps_dev.go
andrewsomething: :+1:  Thanks for the additional info!

elskwid:
Hi Folks, I am seeing the error referenced in the description above when I try to build my apps locally today. It looks like the [CNB builder images were updated](https://github.com/digitalocean/doctl/commit/8c0151342530cfa550d4df539e4850b1861d1c0d). Is it possible that the run-image tag update was missed?

elskwid:
I just rolled back to `doctl v1.109.1` and the problem went away. Does that make it more likely to be related to a run-image tag and builder image mismatch?

andrewsomething:
@elskwid Thanks for flagging this. It look like you were correct. I've just merged https://github.com/digitalocean/doctl/pull/1573 and will cut a new release shortly.

PR #1521, opened 2024-04-02
Add config options for creating global load balancers

asaha2:


commands/load_balancers.go
```diff
@@ -92,6 +92,14 @@ With the load-balancer command, you can list, create, or delete load balancers,
 		"A comma-separated list of ALLOW rules for the load balancer, e.g.: `ip:1.2.3.4,cidr:1.2.0.0/16`")
 	AddStringSliceFlag(cmdLoadBalancerCreate, doctl.ArgDenyList, "", []string{},
 		"A comma-separated list of DENY rules for the load balancer, e.g.: `ip:1.2.3.4,cidr:1.2.0.0/16`")
+	AddStringFlag(cmdLoadBalancerCreate, doctl.ArgLoadBalancerDomains, "", "",
+		"A comma-separated list of domains required to ingress traffic to a global load balancer")
```
andrewsomething, line 96: I don't believe GLBs are available to most customers at the moment. Is that correct? If that's the case we may want to consider hiding these flags from the help output and published documentation until it is more widely available. 

Adding the `hiddenFlag()` option would do that:

```suggestion
	AddStringFlag(cmdLoadBalancerCreate, doctl.ArgLoadBalancerDomains, "", "",
		"A comma-separated list of domains required to ingress traffic to a global load balancer", hiddenFlag())
```

commands/load_balancers.go
```diff
@@ -92,6 +92,14 @@ With the load-balancer command, you can list, create, or delete load balancers,
...
+	AddStringFlag(cmdLoadBalancerCreate, doctl.ArgGlobalLoadBalancerSettings, "", "",
+		"Global load balancer settings")
+	AddStringFlag(cmdLoadBalancerCreate, doctl.ArgGlobalLoadBalancerCDNSettings, "", "",
+		"Global load balancer CDN settings")
```
andrewsomething, line 100: Is there any more detail you can provide on the format that these settings should be in? This help output doesn't really explain to the customer how they can use this flag.

commands/load_balancers.go
asaha2: 👋 We're planning a limited/beta release soon, hence it may be prudent to keep these options visible? Wdyt?

commands/load_balancers.go
asaha2: ACK, will add example config string to the helper text to make it more explicit.

commands/load_balancers.go
```diff
@@ -92,6 +92,14 @@ With the load-balancer command, you can list, create, or delete load balancers,
 		"A comma-separated list of ALLOW rules for the load balancer, e.g.: `ip:1.2.3.4,cidr:1.2.0.0/16`")
 	AddStringSliceFlag(cmdLoadBalancerCreate, doctl.ArgDenyList, "", []string{},
 		"A comma-separated list of DENY rules for the load balancer, e.g.: `ip:1.2.3.4,cidr:1.2.0.0/16`")
+	AddStringFlag(cmdLoadBalancerCreate, doctl.ArgLoadBalancerDomains, "", "",
```
bbassingthwaite, line 95: Should be `AddStringSliceFlag`?

commands/load_balancers.go
```diff
@@ -574,6 +644,49 @@ func buildRequestFromArgs(c *CmdConfig, r *godo.LoadBalancerRequest) error {
 		r.Firewall = firewall
 	}
 
+	dms, err := c.Doit.GetString(c.NS, doctl.ArgLoadBalancerDomains)
```
bbassingthwaite, line 647: Why not use `GetStringSlice`?

commands/load_balancers.go
```diff
@@ -359,6 +389,31 @@ func RunLoadBalancerRemoveForwardingRules(c *CmdConfig) error {
...
+		return err
+	}
+	lbID := c.Args[0]
+
+	force, err := c.Doit.GetBool(c.NS, doctl.ArgForce)
+	if err != nil {
+		return err
+	}
+
+	if force || AskForConfirmDelete("load balancer", 1) == nil {
```
andrewsomething, line 405: This generates this message: 

```
❯ Are you sure you want to delete this load balancer?  yes ▸no
```

`AskForConfirm` gives you more flexibility for the message if we want to craft something more relevant.

commands/load_balancers.go
```diff
@@ -92,6 +92,18 @@ With the load-balancer command, you can list, create, or delete load balancers,
...
 	AddStringSliceFlag(cmdLoadBalancerCreate, doctl.ArgDenyList, "", []string{},
 		"A comma-separated list of DENY rules for the load balancer, e.g.: `ip:1.2.3.4,cidr:1.2.0.0/16`")
+	AddStringSliceFlag(cmdLoadBalancerCreate, doctl.ArgLoadBalancerDomains, "", []string{},
+		"A comma-separated list of domains required to ingress traffic to a global load balancer, e.g.: `name:test-domain,is_managed:true,certificate_id:test-cert-id` "+
+			"(NOTE: this is a closed beta feature, contact DigitalOcean support to review its public availability.)")
+	AddStringFlag(cmdLoadBalancerCreate, doctl.ArgGlobalLoadBalancerSettings, "", "",
+		"Target protocol and port settings for ingressing traffic to a global load balancer, e.g.: `target_protocol:http,target_port:80` "+
+			"(NOTE: this is a closed beta feature, contact DigitalOcean support to review its public availability.)")
+	AddStringFlag(cmdLoadBalancerCreate, doctl.ArgGlobalLoadBalancerCDNSettings, "", "",
+		"CDN cache settings global load balancer, e.g.: `is_enabled:true` "+
```
andrewsomething, line 102: Are additional CDN configuration options likely? A bool flag might be more straight forward if not.

andrewsomething (commented):
Looking good. A couple small comments in line. Additionally, it would be great to get a GLB test case in the integration style tests:

https://github.com/digitalocean/doctl/blob/main/integration/lb_create_test.go

commands/load_balancers.go
```diff
@@ -378,12 +433,29 @@ func extractForwardingRules(s string) (forwardingRules []godo.ForwardingRule, er
...
 
-func fillStructFromStringSliceArgs(obj any, s string) error {
+func extractDomains(s []string) (domains []*godo.LBDomain, err error) {
+	if len(s) == 0 {
+		return domains, err
+	}
+
+	for _, v := range s {
+		domain := new(godo.LBDomain)
+		if err := fillStructFromStringSliceArgs(domain, v, " "); err != nil {
```
andrewsomething, line 443: Sorry for all the back and forth here, but I did have one more question.

If feels odd to break the pattern used by other flags for `domains`. `glb-settings` takes a comma-separated list of key-value pairs as does `forwarding-rules`.

```
--domains "name:test-domain-1 is_managed:true certificate_id:test-cert-id-1"
--glb-settings target_protocol:http,target_port:80
--forwarding-rules entry_protocol:tcp,entry_port:3306,target_protocol:https,target_port:443
```

Can we stick to the existing pattern? I think it will be a more consistent experience for users.

andrewsomething (changes requested):

commands/load_balancers.go
asaha2: 👋 The `glb-settings` are _flat_ config params which could be delimited by commas and we're parsing as expected. The `forwarding-rules` are a _slice_ of config rules: we actually have a limitation to how we handle this parameter in `doctl` currently. If you try passing multiple forwarding rules for an LB (as normally expected), only the latest rule will be used. E.g.:
```
./doctl \
...
--forwarding-rules "entry_protocol:http,entry_port:80,target_protocol:http,target_port:80" \
--forwarding-rules "entry_protocol:tcp,entry_port:50002,target_protocol:tcp,target_port:50002"
```

Running the above we're _only_ able to set the last forwarding rule specified (TCP). With `domains` the expectation should be the same, it takes a _slice_ of params and this adjustment to using a space delimiter makes it possible to parse multiple elements for a slice param appropriately.

asaha2: So basically TLDR:

- `forwarding-rules` accepts arg through `AddStringFlag` thus parses the param as a flat string, which isn't correct considering customers will have multiple forwarding rules
- `domains` accepts arg through `AddStringSliceFlag` thus parses the param as a slice of string, which is logically correct, except comma separator cannot be used to distinguish components in the same element

andrewsomething (approved):

PR #1522, opened 2024-04-10
Document App Platform crash log value

dbrian57:
Resolves PDOCS-2445. Documents the `run_restarted` value for `doctl apps logs $app_id --type`.


andrewsomething (approved):
LGTM!

PR #1535, opened 2024-05-28
Corrects DOCR list manifests example

dbrian57:
Resolves PDOCS-2515. This PR corrects an example where the registry name was incorrectly prepended to the repository name. The example should be correct now.


andrewsomething (approved):
:+1:  Thanks Dan!

PR #1548, opened 2024-07-15
Fix typo in DOKS 1-click example text

jkpe:
👋 

Just a very small typo correction in the DOKS 1-click example text when running `doctl kubernetes 1-click install --help`

It removes the extra `>` which should not be there. Running the command with the extra `>` would result in `Notice: No addons specified for cluster:`


andrewsomething (approved):
:+1:  Thanks Jack!

PR #1549, opened 2024-07-16
Improve description of Firewall arguments related to tags

brunograsselli:
The previous descriptions suggested that the firewall would be tagged. However, what really happens is that the droplets with those tags will be placed behind the firewall.


andrewsomething (approved):
:+1:  Thanks Bruno!

PR #1550, opened 2024-07-16
Fix spelling in firewall documentation

beardicus:
Also tweaks one description to be more similar to existing text.


andrewsomething (approved):
:+1:  Thanks Brian! Should have caught that :man_facepalming:

PR #1551, opened 2024-07-16
ci: Configure crate-ci/typos and auto fix what I can

halkeye:
Can setup a github action too

https://github.com/crate-ci/typos/blob/master/docs/github-action.md

### DIFF
--- a/args.go
+++ b/args.go
@@ -429,8 +429,8 @@ const (
 	ArgDatabaseTopicMaxCompactionLagMS = "max-compaction-lag-ms"
 	// ArgDatabaseTopicMaxMessageBytes is the maximum size, in bytes, of the largest record batch that can be sent to the server
 	ArgDatabaseTopicMaxMessageBytes = "max-message-bytes"
-	// ArgDatabaseTopicMesssageDownConversionEnable determines whether brokers should convert messages for consumers expecting older message formats
-	ArgDatabaseTopicMesssageDownConversionEnable = "message-down-conversion-enable"
+	// ArgDatabaseTopicMessageDownConversionEnable determines whether brokers should convert messages for consumers expecting older message formats
+	ArgDatabaseTopicMessageDownConversionEnable = "message-down-conversion-enable"
 	// ArgDatabaseTopicMessageFormatVersion is the version used by the broker to append messages to the kafka topic logs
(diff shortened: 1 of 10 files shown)


args.go
```diff
@@ -429,8 +429,8 @@ const (
 	ArgDatabaseTopicMaxCompactionLagMS = "max-compaction-lag-ms"
 	// ArgDatabaseTopicMaxMessageBytes is the maximum size, in bytes, of the largest record batch that can be sent to the server
 	ArgDatabaseTopicMaxMessageBytes = "max-message-bytes"
-	// ArgDatabaseTopicMesssageDownConversionEnable determines whether brokers should convert messages for consumers expecting older message formats
-	ArgDatabaseTopicMesssageDownConversionEnable = "message-down-conversion-enable"
+	// ArgDatabaseTopicMessageDownConversionEnable determines whether brokers should convert messages for consumers expecting older message formats
+	ArgDatabaseTopicMessageDownConversionEnable = "message-down-conversion-enable"
```
andrewsomething, line 433: As this is a "code aware" spell checker, any  chance you happen to know if there is a way to exclude existing exported vars/consts? That's not really an issue for doctl itself, but if we were to bring this into say godo, we'd want to avoid breaking changes.

andrewsomething (commented):
Great stuff! I've been meaning to get to adding something like this. Thanks :pray:  

Could you add a note somewhere in the CONTRIBUTING.md around how to run/set this up for local dev?

args.go
halkeye: so best i've found in the past was `default.extend-identifiers` in the config, but nothing automated
I believe typos also has a cli flag for listing all the identifiers it finds

that being said, fixing the spelling is an easy breaking change to handle for end users, its not like they have to figure out how to rewrite code because a function no longer exists.

halkeye:
I can when I get a minute. Do you want a PR github action too?

andrewsomething:
> Do you want a PR github action too?

If you're up for it. That would be great. We can add that latter too to no block this PR.

halkeye:
> > Do you want a PR github action too?
> 
> If you're up for it. That would be great. We can add that latter too to no block this PR.

https://github.com/digitalocean/doctl/pull/1552

decided to do it seperately

andrewsomething (approved):
:+1:  Thanks again Gavin!

PR #1555, opened 2024-07-18
Enable control plane firewall pointer arguments and trim spaces

llDrLove:
- Ensure that the `enable-control-plane-firewall` flag is a optional boolean and only pass the configuration when needed.
- Ensure that the IPs or CIDRs are trimmed from whitespaces when building the Update or Create request.


commands/kubernetes.go
```diff
@@ -1679,6 +1675,9 @@ func buildClusterCreateRequestFromArgs(c *CmdConfig, r *godo.KubernetesClusterCr
 		if r.ControlPlaneFirewall == nil {
 			r.ControlPlaneFirewall = &godo.KubernetesControlPlaneFirewall{}
 		}
+		for i := range controlPlaneFirewallAllowedAddresses {
+			controlPlaneFirewallAllowedAddresses[i] = strings.ReplaceAll(controlPlaneFirewallAllowedAddresses[i], " ", "")
```
timoreimann, line 1679: I'm inclined to prefer [`strings.TrimSpace()`](https://pkg.go.dev/strings#TrimSpace) which also handles all sorts of unicode variants.

Similar for the other usage elsewhere.

llDrLove: Sure, I've refactored it to use strings.TrimSpace()

andrewsomething (approved):
:+1:  LGTM!

PR #1556, opened 2024-07-19
APPS-8731: Update CNB Builder image

ShivaniKumar1:
APPS-8731
Update CNB Builder image from apps-images.


andrewsomething (approved):
:+1:  LGTM!

Issue #1553, opened 2024-07-17
Add ability to change connection pool size for connection pools


h0jeZvgoxFepBQ2C:
**What is the problem this feature would solve? Please describe.**
Change the conneciton pool size via doctl

**Describe the solution you'd like**
doctl databases pool update database123id poolname123id --size 2

In the WebUI you can change the pool size, so i dont understand why this is not working via doctl itself?

bentranter:
@h0jeZvgoxFepBQ2C We must've missed this when adding connection pool support, so thanks for calling this out.

I've fixed this in the latest release [v1.110.0](https://github.com/digitalocean/doctl/releases/tag/v1.110.0), so as of that release you can run the command as you described (`doctl databases pool update database123id poolname123id --size 2`) and it will work as expected.

h0jeZvgoxFepBQ2C:
hmmm it seems that this does something unintended, and removes the username?

Before:

<img width="1206" alt="Bildschirmfoto 2024-09-06 um 11 13 33" src="https://github.com/user-attachments/assets/9332dd80-bd99-406e-9ffc-e083b47a943a">

> doctl databases pool update 123-123-123-123 p_danube22_22_07_13__1300 --size 1

<img width="1213" alt="Bildschirmfoto 2024-09-06 um 11 14 12" src="https://github.com/user-attachments/assets/58e82843-3f3e-407e-b4cd-f0b8ba266c50">

?

h0jeZvgoxFepBQ2C:
@bentranter any feedback on this? Right now we can't use this command since it breaks our connections :(

andrewsomething:
Hi @h0jeZvgoxFepBQ2C,

By default, the connection pool uses the connecting user's privileges. By excluding the `--user` flag when updating, you reset it to the default. To use a specific user, you need to pass the flag as well:

```
doctl databases pool update <database-uuid> <pool-name> --size <size> --user <username>
```

h0jeZvgoxFepBQ2C:
Phu, this is a bit cumbersome.. wouldnt it be better to leave it as it is? otherwise you always have to make a second call to get the username first

PR #1557, opened 2024-07-23
databases: update connection pool

bentranter:
This adds the missing update functionality for database connection pools, as identified in https://github.com/digitalocean/doctl/issues/1553.


commands/databases.go
```diff
@@ -1202,6 +1215,62 @@ func buildDatabaseCreatePoolRequestFromArgs(c *CmdConfig) (*godo.DatabaseCreateP
...
+	req := &godo.DatabaseUpdatePoolRequest{}
+	size, err := c.Doit.GetInt(c.NS, doctl.ArgDatabasePoolSize)
+	if err != nil {
+		return err
+	}
+	if size != 0 {
+		req.Size = size
+	} else {
+		req.Size = pool.Size
+	}
```
andrewsomething, line 1241: This is a nice pattern for handling PUTs. We might want to look to do this in other commands where the API endpoint expects a full representation of the resource in the PUT body (i.e. doesn't act like a patch).

andrewsomething (approved):

PR #1558, opened 2024-07-24
go: upgrade to 1.22

bentranter:
Upgrades the GitHub CI workflows and `go.mod` entry for Go 1.22. Just in time for Go 1.23 to be released 😅

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
   test:
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


danaelhe (approved):

bentranter:
When I click the GitGuardian ID it says, "not found", so I'm guessing this is getting falsely flagged from the change in #1557 prior to it being marked as a test value. Going to merge as-is and see if it's fine.

PR #1569, opened 2024-08-13
Droplet: Add SizeSlug as format flag

danaelhe:
SizeSlug as a format flag was supported but wasn't documented.

### DIFF
--- a/commands/displayers/droplet.go
+++ b/commands/displayers/droplet.go
@@ -34,7 +34,7 @@ func (d *Droplet) JSON(out io.Writer) error {
 
 func (d *Droplet) Cols() []string {
 	cols := []string{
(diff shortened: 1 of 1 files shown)


danaelhe:
Ooop need to update some tests too, one sec

andrewsomething:
Adding new columns isn't necessarily a breaking change. For scripting, we should be encouraging usage of the `--format` flag to pick the specific values the user cares about and provide a stable order. Though we also need to think about readability of the default output. Unfortunately, the output here is already too wide to fit on a single line in most cases. While the slug is useful, it does duplicate some of the info already shown.

I wonder if there is some more general solution for non-default columns? There are a lot of commands that could benefit.

The values shown in the help output come from here:

https://github.com/digitalocean/doctl/blob/9282784517cb25d3cb6b06a4895626b173177603/commands/command_option.go#L35

It might be as simple as:

```diff
-               c.fmtCols = d.Cols()
+               c.fmtCols = maps.Keys(d.ColMap())
```

Though we need to check that the change doesn't have any undesired side effects.

PR #1576, opened 2024-09-10
misc: use 'go install' in _install_github_release_notes make target

andrewsomething:
Fix for:

```
$ make _install_github_release_notes
go: modules disabled by GO111MODULE=off; see 'go help modules'
make: *** [Makefile:163: _install_github_release_notes] Error 1
```


loosla (approved):
Thank you :fire:

PR #1579, opened 2024-09-11
[DBAAS] | Add API endpoint for applying cluster patches

vsharma6855:
Add 'doctl databases install_update <database_uuid>' command and unit tests.

### DIFF
--- a/commands/databases.go
+++ b/commands/databases.go
@@ -147,6 +147,7 @@ For PostgreSQL and MySQL clusters, you can also provide a disk size in MiB to sc
 
 	cmd.AddCommand(databaseReplica())
 	cmd.AddCommand(databaseMaintenanceWindow())
+	cmd.AddCommand(databaseInstallUpdate())
 	cmd.AddCommand(databaseUser())
 	cmd.AddCommand(databaseDB())
 	cmd.AddCommand(databasePool())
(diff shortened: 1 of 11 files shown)


loosla:
Thanks for your contribution :rocket:  

I believe we are using hyphen instead of underscores (install_update) in commands names. 
Maybe just `update` can be clear enough as a name? 
I'd rely on @andrewsomething as he may share more thoughts about the naming.

commands/databases.go
```diff
@@ -599,6 +600,21 @@ To see a list of your databases and their IDs, run `+"`"+`doctl databases list`+
...
 
+func databaseInstallUpdate() *Command {
+	cmd := &Command{
+		Command: &cobra.Command{
+			Use:   "install_update",
+			Short: "Display commands for starting installation of updates",
+			Long:  `The ` + "`" + `doctl databases install_update` + "`" + ` commands allow you to starts installation of updates for your databases`,
+		},
+	}
+	cmdDatabaseInstallUpdate := CmdBuilder(cmd, RunDatabaseInstallUpdate, "install_update <database-cluster-id>", "Starts installation of updates ", "Starts installation of updates ", Writer)
```
andrewsomething, line 611: Here you create a new `install_update` sub-group under `databases` and add an `install_update` command to it. So to use this command, you must run:

    doctl databases install_update install_update <database-cluster-id> 

Instead, I think grouping this with the existing maintenance window commands might make a lot of sense. It keeps sub-commands related to applying updates together. So we'd have something like:

```
$ doctl databases maintenance-window -h

# ...

Available Commands:
  get         Retrieve details about a database cluster's maintenance windows
  update      Update the maintenance window for a database cluster
  install     Start installation of updates for a database cluster outside of a maintenance window
```

If we attach it to the top-level `databases` command, please use `install-update` with a hyphen not an underscore for consistency with other commands.

andrewsomething (changes requested):

loosla:
Looks good to me :+1:

andrewsomething (approved):
:+1:  LGTM

Issue #1512, opened 2024-03-12
Access the link to download CA certificate for managed databases


AAverin:
**What is the problem this feature would solve? Please describe.**

To fully automate managed database cluster setup within my infrastructure, including access to the database by other services, I would like to have a way to access CA database certificate download link.
This would allow me to download certificate and pass it through the environment to configure SSL access.
Additionally I would love that to be accessible with Terraform too.

**Describe the solution you'd like**
Doctl implementation + Terraform exposed output parameter for the `digitalocean_database_cluster` resource: https://registry.terraform.io/providers/digitalocean/digitalocean/latest/docs/resources/database_cluster

**Additional context**

MrCuPper:
Hello, God, I don't know what I did, I don't know programming at all, I'm
just very interested in these programs, but I don't have the money to
spend, nor the possibility to study, that's why I got stuck, I don't know
what I did.

andrewsomething:
On the Terraform side, the `digitalocean_database_ca` data source has been available since v2.13.0 (released October 7, 2021).

https://registry.terraform.io/providers/digitalocean/digitalocean/latest/docs/data-sources/database_ca

loosla:
Implemented in doctl [v1.114.0](https://github.com/digitalocean/doctl/releases/tag/v1.114.0)

PR #1580, opened 2024-09-11
databases: add get-ca to retrieve db certificate

loosla:
This PR allows users to obtain a certificate for connecting to the database.

Related to https://do-internal.atlassian.net/browse/APICLI-2506, https://github.com/digitalocean/doctl/issues/1512

### DIFF
--- a/commands/databases.go
+++ b/commands/databases.go
@@ -204,7 +204,7 @@ func RunDatabaseGetCA(c *CmdConfig) error {
 		return err
 	}
 
-	return displayDatabaseCA(c, *dbCA)
+	return displayDatabaseCA(c, dbCA)
 }
 
 // RunDatabaseCreate creates a database cluster
@@ -925,8 +925,8 @@ func displayDatabaseUsers(c *CmdConfig, users ...do.DatabaseUser) error {
 	return c.Display(item)
 }
 
-func displayDatabaseCA(c *CmdConfig, dbCAs ...do.DatabaseCA) error {
-	item := &displayers.DatabaseCAs{DatabaseCAs: dbCAs}
+func displayDatabaseCA(c *CmdConfig, dbCA *do.DatabaseCA) error {
+	item := &displayers.DatabaseCA{DatabaseCA: *dbCA}
 	return c.Display(item)
 }
 
--- a/commands/databases.go
+++ b/commands/databases.go
@@ -70,6 +70,10 @@ func Databases() *Command {
 - The date and time when the database cluster was created`+databaseListDetails, Writer, aliasOpt("g"), displayerType(&displayers.Databases{}))
 	cmdDatabaseGet.Example = `The following example retrieves the details for a database cluster with the ID ` + "`" + `f81d4fae-7dec-11d0-a765-00a0c91e6bf6` + "`" + ` and uses the ` + "`" + `--format` + "`" + ` flag to return only the database's ID, engine, and engine version: doctl databases get f81d4fae-7dec-11d0-a765-00a0c91e6bf6`
 
+	// TODO: update draft descriptions, examples when ready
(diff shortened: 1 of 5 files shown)


commands/displayers/database.go
```diff
@@ -155,6 +155,41 @@ func (db *DatabaseBackups) KV() []map[string]any {
...
+		"Certificate": "Certificate",
+	}
+}
+
+func (dc *DatabaseCAs) KV() []map[string]any {
+	out := make([]map[string]any, 0, len(dc.DatabaseCAs))
+
+	for _, c := range dc.DatabaseCAs {
+		o := map[string]any{
+			"Certificate": string(c.Certificate),
```
loosla, line 185: TODO: run with -o text, test, encode

commands/displayers/database.go
```diff
@@ -155,6 +155,36 @@ func (db *DatabaseBackups) KV() []map[string]any {
...
+func (dc *DatabaseCA) ColMap() map[string]string {
+	return map[string]string{
+		"Certificate": "Certificate",
+	}
+}
+
+func (dc *DatabaseCA) KV() []map[string]any {
+	return []map[string]any{
+		{
+			"Certificate": string(dc.DatabaseCA.Certificate),
```
loosla, line 183: `json` and `text` formats have different certificate values. 
users won't be able to connect with the cert taken from json :crying_cat_face:

andrewsomething: The value is base64 encoded. As that matches the API response, I think that is OK. Though we should probably mention it in the help output for the command.

This should show the same value as the text output:

```
./builds/doctl databases get-ca <uuid> -o json | jq -r .certificate | base64 --decode
```

loosla: Thanks a lot, Andrew :pray: This really helps :+1: 

I'm struggling to explain it in a simple way to our users, but I'll try my best :slightly_smiling_face: 
I'm afraid that this difference can be confusing and will lead to more open issues down the road as people may expect json and text to use the same encoding...

commands/databases.go
```diff
@@ -70,6 +70,10 @@ func Databases() *Command {
 - The date and time when the database cluster was created`+databaseListDetails, Writer, aliasOpt("g"), displayerType(&displayers.Databases{}))
 	cmdDatabaseGet.Example = `The following example retrieves the details for a database cluster with the ID ` + "`" + `f81d4fae-7dec-11d0-a765-00a0c91e6bf6` + "`" + ` and uses the ` + "`" + `--format` + "`" + ` flag to return only the database's ID, engine, and engine version: doctl databases get f81d4fae-7dec-11d0-a765-00a0c91e6bf6`
 
+	cmdDatabaseGetCA := CmdBuilder(cmd, RunDatabaseGetCA, "get-ca <database-cluster-id>", "Provides the CA certificate for a DigitalOcean database", `Retrieves a database certificate`, Writer, aliasOpt("gc"), displayerType(&displayers.DatabaseCA{}))
+	cmdDatabaseGetCA.Example = `Retrieves the database certificate for the cluster with the ID ` + "`" + `f81d4fae-7dec-11d0-a765-00a0c91e6bf6` + "`" + `: doctl databases get-ca f81d4fae-7dec-11d0-a765-00a0c91e6bf6
+With the -o json flag, decode the certificate to connect to the database: doctl databases get-ca <database-cluster-id> -o json | jq -r .certificate | base64 --decode`
```
bentranter, line 75: Small nitpick here – just wanted to tweak the wording, and add formatting to the example flag and commands so that it gets rendered more nicely in the generated documentation.
```suggestion
With the ` + "`" + `-o json flag` + "`" + `, the certificate to connect to the database is base64 encoded. To decode it: ` + "`" + `doctl databases get-ca <database-cluster-id> -o json | jq -r .certificate | base64 --decode` + "`"
```

loosla: Good idea :+1:  Thanks

bentranter (approved):
LGTM!

PR #1589, opened 2024-10-04
Update godo that uses opensearch ism_history_max_docs correct int64 type

loosla:
This PR updates godo version. 
A new godo version uses a correct `int64` type for Opensearch `ism_history_max_docs` field.


loosla:
Already in `main` branch.

Issue #1585, opened 2024-09-25
Add --wait flag to databases resize command


loosla:
**What is the problem this feature would solve? Please describe.**
doctl aims to be useful in scripting. One way it does this is by providing a --wait flag for long running, async operations to block until they are complete rather than making the user implement polling in their own script. Some examples include:

- `doctl database create --wait`
- `doctl compute droplet create --wait`
- `doctl kubernetes cluster create --wait`
- `doctl apps create --wait`

Currently the `doctl databases resize` command does not have a `--wait` flag.

**Describe the solution you'd like**
Add a --wait flag to the `doctl databases resize` command. It should poll the [/v2/databases/$DATABASE_ID/resize](https://docs.digitalocean.com/reference/api/api-reference/#operation/databases_update_clusterSize), blocking until the database cluster's status attribute moves from resizing to online.

**Additional context**
Following the code path starting here demonstrates how `doctl database create`'s  `--wait` flag is implemented:

https://github.com/digitalocean/doctl/blob/de1918ad1ee4b1e523c335fa9254e85d415d18aa/commands/apps.go#L282

Yordaniss:
Hello @loosla This one sounds interesting. Can you please assign this issue to me? 🙂 Thanks

SkySingh04:
This looks like a fun issue, Will be working on this!

Yordaniss:
Hey @SkySingh04 sorry i already started working on this and already created a PR ^^

PR #1590, opened 2024-10-08
ISSUE: 1585 | Add --wait flag to databases resize command

Yordaniss:
ISSUE #1585 
* added bool flag into resize command
* added wait functionality into resize
* adjusted test

### DIFF
--- a/commands/databases.go
+++ b/commands/databases.go
@@ -136,7 +136,7 @@ For PostgreSQL and MySQL clusters, you can also provide a disk size in MiB to sc
 	AddStringFlag(cmdDatabaseResize, doctl.ArgSizeSlug, "", "", nodeSizeDetails, requiredOpt())
 	AddIntFlag(cmdDatabaseResize, doctl.ArgDatabaseStorageSizeMib, "", 0, storageSizeMiBDetails)
 	AddBoolFlag(cmdDatabaseResize, doctl.ArgCommandWait, "", false,
-		"Boolean that specifies whether to wait for an app to complete before returning control to the terminal")
+		"Boolean that specifies whether to wait for the resize to complete before returning control to the terminal")
 	cmdDatabaseResize.Example = `The following example resizes a PostgreSQL or MySQL database to have two nodes, 16 vCPUs, 64 GB of memory, and 2048 GiB of storage space: doctl databases resize ca9f591d-9999-5555-a0ef-1c02d1d1e352 --num-nodes 2 --size db-s-16vcpu-64gb --storage-size-mib 2048000 --wait true`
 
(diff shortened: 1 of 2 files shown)


commands/databases.go
```diff
@@ -511,13 +513,57 @@ func RunDatabaseResize(c *CmdConfig) error {
...
+	err = dbs.Resize(id, r)
+	if err != nil {
+		return err
+	}
+
+	// Retrieve the database object after resize (to check status or further actions)
+	db, err := dbs.Get(id)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve the database after resize: %v", err)
+	}
```
andrewsomething, line 533: I think that we could probably simplify this a bit further. Compared to the create command, the resize command only returns success or error. It doesn't output the database information. So we can remove this call to `Get` here as well as the one below in the `wait` block.

commands/databases.go
```diff
@@ -511,13 +513,57 @@ func RunDatabaseResize(c *CmdConfig) error {
...
+				"failed to retrieve the resized database: %v",
+				err,
+			)
+		}
+		db.Connection = connection
+	}
+
+	notice("Database resized successfully")
+
+	return displayDatabases(c, false, *db)
```
andrewsomething, line 566:
```suggestion
		notice("Database resized successfully")
	}

	return nil
```

Let's also move the notice inside the wait block. That way the current behavior remains the same if you do not pass the `--wait` flag.

andrewsomething (changes requested):
Hi @Yordaniss. Thank you so much for the pull request!

I've left a few comments in-line with suggestions. In addition to that, it would be great to add a new test case to the `TestDatabaseResize` test. 
https://github.com/digitalocean/doctl/blob/6804c7d62dfd7ff142ddb3ac1abed3893f9cdcab/commands/databases_test.go#L618

Something along the line of:

```go
	// Success with wait flag
		tm.databases.EXPECT().Resize(testDBCluster.ID, r).Return(nil)
		tm.databases.EXPECT().Get(testDBCluster.ID).Return(&testDBCluster, nil)
		config.Args = append(config.Args, testDBCluster.ID)
		config.Doit.Set(config.NS, doctl.ArgSizeSlug, testDBCluster.SizeSlug)
		config.Doit.Set(config.NS, doctl.ArgDatabaseNumNodes, testDBCluster.NumNodes)
		config.Doit.Set(config.NS, doctl.ArgDatabaseStorageSizeMib, testDBCluster.StorageSizeMib)
		config.Doit.Set(config.NS, doctl.ArgCommandWait, true)

		err := RunDatabaseResize(config)
		assert.NoError(t, err)
	})
```

Comparing that to the existing tests, you'll see it passes the `wait` flag and tests that the `Get` method is called by `waitForDatabaseReady` when the flag is passed.

Let us know if you have any questions!

Yordaniss:
Hi @andrewsomething thanks a lot for a feedback. I simplified logic and added test case. 🙂

andrewsomething (approved):
:+1:  Great work! Thanks again for the contribution!

PR #1592, opened 2024-10-16
doks: fix panic during create

gottwald:
If a user creates a DOKS cluster through doctl and deletes it through some other means while doctl is waiting for the creation the command panics. This fixes the panic and returns an appropriate error.


lfundaro-do (approved):

andrewsomething (approved):
:+1:  Good catch

PR #1593, opened 2024-10-17
fix(app): #1536 fix logging format to work with pino-pretty

### LINKED ISSUE #1536: doctl apps logs doesn't output JSON

On our apps services, we use JSON logging (pino) and it'd be really helpful to be able to do `doctl apps logs <id> -f | pino-pretty` but I can't because the log lines include the service name and the log timestamp, not just the log message:

```
webapp 2024-05-29T20:14:14.996999111Z {"level":30,"time":1717013654993,  ....
```

I'd like to just get what I logged, i.e., the JSON, without the `webapp 2024-05-29T20:14:14.996999111Z ` prefix.

**Additional context**

This would greatly improve my experience debugging apps on DO's App platform (currently we use papertrail, but it's not a particularly good log querier (doesn't parse the log lines correctly)

mohit-nagaraj:
## Summary
This PR introduces a new --output-json flag for the RunAppsGetLogs function in the DigitalOcean CLI (doctl). When this flag is used, the logs are outputted in JSON format without the service name and timestamp prefix.

### Changes Made
- Added --output-json flag to the RunAppsGetLogs function.
- Modified the log processing logic to remove the service name and timestamp prefix when the --output-json flag is used.

Closes #1536

### DIFF
--- a/args.go
+++ b/args.go
@@ -54,6 +54,8 @@ const (
 	ArgAppLogFollow = "follow"
 	// ArgAppLogTail tail logs.
 	ArgAppLogTail = "tail"
+	// ArgNoPrefix no prefix to json logs
+	ArgNoPrefix = "no-prefix"
 	// ArgAppForceRebuild forces a deployment rebuild
 	ArgAppForceRebuild = "force-rebuild"
 	// ArgAppAlertDestinations is a path to an app alert destination file.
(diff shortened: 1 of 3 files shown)


andrewsomething (changes requested):
Thanks for this great contribution @mohit-nagaraj! I do have some feedback:

For your new flag to work as expected, it needs to be registered with the command. You can see how that was done with other flags for the `logs` command here:

https://github.com/digitalocean/doctl/blob/7f7a3c090e485bf7717310810df06c5a35fe1b06/commands/apps.go#L167-L188

That is also where you add the information that shows in the help output.

I also think `output-json` might not be the best name for this flag. The output is only in JSON if your application logs in JSON. If you log in plain text, using the `output-json` would still produce plain text. Maybe something like `no-prefix` as it removes the prefixes from the logs? We could provide a hint in the help text that it useful for JSON structured logs.

mohit-nagaraj:
Sure got it fixing them

mohit-nagaraj:
Hey I updated code as specified, Could you re-review them will implement any of the changes you say

andrewsomething (approved):
:+1:  LGTM! Thanks again for the great contribution!

Issue #1594, opened 2024-10-18
Option descriptions not printed correctly.


zikalino:
**Describe the Issue:**

I have noticed that option descriptions are not printed exactly as defined in the code. For instance, the code is:

	AddStringFlag(cmdKubeClusterCreate, doctl.ArgClusterVersionSlug, "", "latest",
		"A `slug` indicating which Kubernetes version to use when creating the cluster. Use the `doctl kubernetes options versions` command for a list of options")

and we get following output:

      --version slug                                       A slug indicating which Kubernetes version to use when creating the cluster. Use the `doctl kubernetes options versions` command for a list of options (default "latest")

Or incase of registry, **slug** is not quoted in the source code:

		"A slug indicating which datacenter region the registry reside in. For a list of supported region slugs, use the `doctl registry options available-regions` command")
	cmdRunRegistryCreate.Example = `The following example creates a registry named ` + "`" + `example-registry` + "`" + ` in the NYC3 region: doctl registry create example-registry --region=nyc3`

and instead quotes around **doctl registry options available-regions** are removed as below:

      --region doctl registry options available-regions               A slug indicating which datacenter region the registry reside in. For a list of supported region slugs, use the doctl registry options available-regions command

It looks like first pair of quotes `...` is always removed from every option.

**Environment:**

doctl version 0.0.0-dev
release 1.115.0 is available, check it out!

**Additional Details:**

zikalino:
Ok, I figured it out, it seems this is intentional and first item in `` quotes is copied just after the param list.

So in some cases if **slug** is not quoted, wrong thing is appended to parameters, like here:

--region **doctl registry options available-regions** which is obviously not intentional

I have created a PR fixing it in a few places: https://github.com/digitalocean/doctl/pull/1595

andrewsomething:
Thanks again!

Closed via: https://github.com/digitalocean/doctl/pull/1595

PR #1595, opened 2024-10-18
Fixing not properly displayed options

zikalino:
This fixes improperly displayed options as described here:

https://github.com/digitalocean/doctl/issues/1594


zikalino:
@zikalino thanks for merging!
I think i located all the places where **slug** was not quoted, if i found anything else causing similar problems i will make another PR :-)

PR #1596, opened 2024-10-18
Add Database Replica Integration Tests

### LINKED ISSUE #1359: Database Replica Functions Missing Integration Tests
**Describe the Issue:**

When working on #1348 to add `promote` functionality it was discovered that other endpoints under `replica` are missing integration tests. `replica create`, `replica delete`, `replica get`, `replica list`, and `replica getReplicaConnection` are all missing integration tests.

**Environment:**
Local Dev

gsquire:
This patch adds the remaining database replica integration tests. It also corrects some of the parameters for the replica promote test.

Tested with `make test_integration`.

Closes #1359

### DIFF
--- a/integration/database_replica_create_test.go
+++ b/integration/database_replica_create_test.go
@@ -2,6 +2,7 @@ package integration
 
 import (
 	"fmt"
+	"io"
 	"net/http"
 	"net/http/httptest"
 	"net/http/httputil"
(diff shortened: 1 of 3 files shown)


gsquire:
I do want to call out the fact that I couldn't find the corresponding API for the [`replica connection`](https://docs.digitalocean.com/reference/doctl/reference/databases/replica/connection/) command. So there isn't an integration test for that. I searched for 'replica' [here](https://docs.digitalocean.com/reference/api/api-reference/). If someone could help point me in the right direction, I can add it in. Thanks!

andrewsomething:
@gsquire  Thanks for this great contribution!

> I do want to call out the fact that I couldn't find the corresponding API for the replica connection command.

The `replica connection` command is a bit of an abstraction. There is a not an endpoint that only returns that information. If you look at the implementation, you'll see that we're calling the [`GET  /v2/databases/$DATABASE_ID/replicas/$REPLICA_NAME`](https://docs.digitalocean.com/reference/api/api-reference/#operation/databases_get_replica) endpoint and just surfacing the connection information.

https://github.com/digitalocean/doctl/blob/062646883a3a9e3c24cce7d1a43a030a089abc4c/do/databases.go#L622-L631

andrewsomething (approved):
:+1:

gsquire:
> The replica connection command is a bit of an abstraction.

Ah, okay. Thanks for reviewing!

PR #1600, opened 2024-10-25
Fixing positional resource id descriptions

zikalino:
I have fixed some other places and replaced **<id>** with **<resource-id>** in command descriptions as it's more common and descriptive pattern.

### DIFF
--- a/commands/firewalls.go
+++ b/commands/firewalls.go
@@ -70,7 +70,7 @@ Use a quoted string of space-separated values for multiple rules.`
 	dropletIDRulesTxt := "A comma-separated list of Droplet IDs to place behind the cloud firewall, for example: `386734086,391669331`"
 	tagNameRulesTxt := "A comma-separated list of existing tags, for example: `frontend,backend`. Droplets with these tags will be placed behind the cloud firewall"
 
-	cmdFirewallGet := CmdBuilder(cmd, RunFirewallGet, "get <id>", "Retrieve information about a cloud firewall", `Retrieves information about an existing cloud firewall, including:`+fwDetail, Writer, aliasOpt("g"), displayerType(&displayers.Firewall{}))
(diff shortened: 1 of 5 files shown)


commands/registry.go
```diff
@@ -80,12 +80,12 @@ func Registry() *Command {
 	cmdRunRegistryCreate.Example = `The following example creates a registry named ` + "`" + `example-registry` + "`" + ` in the NYC3 region: doctl registry create example-registry --region=nyc3`
 
 	getRegDesc := "Retrieves details about a private container registry, including its name and the endpoint used to access it."
-	cmdRegistryGet := CmdBuilder(cmd, RunRegistryGet, "get", "Retrieve details about a container registry",
+	cmdRegistryGet := CmdBuilder(cmd, RunRegistryGet, "get <registry-id>", "Retrieve details about a container registry",
```
andrewsomething, line 83: At the moment, only one container registry is supported per account. So this command is a bit of an outlier. There is no positional argument, just `doctl registry get`.

commands/registry.go
```diff
@@ -80,12 +80,12 @@ func Registry() *Command {
...
 		getRegDesc, Writer, aliasOpt("g"), displayerType(&displayers.Registry{}))
 	cmdRegistryGet.Example = `The following example retrieves details about a registry named ` + "`" + `example-registry` + "`" + `: doctl registry get example-registry`
 
 	deleteRegDesc := "Permanently deletes a private container registry and all of its contents. This is irreversible."
-	cmdRunRegistryDelete := CmdBuilder(cmd, RunRegistryDelete, "delete",
+	cmdRunRegistryDelete := CmdBuilder(cmd, RunRegistryDelete, "delete <registry-id>",
```
andrewsomething, line 88: Same thing here.

andrewsomething (changes requested):
Thanks for another great contribution! I left a couple comments inline regarding the container registry commands. They do not actually require a positional argument.

commands/registry.go
zikalino: oops.. you're right..
it actually works without registry id. i was specifying registry id as first parameter but seems it was actually ignored.....

zikalino:
ok, removed registry changes. i left other improvements - probably still worth merging.

andrewsomething (approved):

andrewsomething:
Definetly! Thanks again!

PR #1601, opened 2024-10-25
droplets: support listing GPU Droplets.

andrewsomething:
Adds a `--gpus` flag to the `doctl compute droplets list` command.


loosla (approved):
:+1: :rocket:

PR #1603, opened 2024-10-28
Don't set APP_CACHE_DIR if --no-cache is passed with apps dev build

fumblehool:
## Description

It mounts the`/cnb/cache` directory for caching the build.
This fails to build in Github Actions since user doesn't have permission to write to /
Error message is : `Error: failed to chown volumes: chown /cnb/cache: operation not permitted`

`doctl apps dev build` has a flag `--no-cache` which disables the caching [docs-link](https://docs.digitalocean.com/reference/doctl/reference/apps/dev/build/).
But this is not being used for buildpack-based builds.

This PR fixes this behavior to disable caching when `--no-cache` is set.

### DIFF
--- a/internal/apps/builder/cnb.go
+++ b/internal/apps/builder/cnb.go
@@ -92,7 +92,7 @@ func (b *CNBComponentBuilder) Build(ctx context.Context) (res ComponentBuilderRe
 		})
 	}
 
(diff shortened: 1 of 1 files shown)


markusthoemmes (approved):

andrewsomething (approved):
:+1:

PR #1604, opened 2024-10-30
LBAAS: remove beta notes on LBaaS fields

jvasilevsky:
Now that global load balancer is in GA and internal load balancer is in EA, remove the beta notes on these associated load balancer fields


andrewsomething (approved):
:+1:  LGTM!

PR #1608, opened 2024-11-05
APPS-9765 Add command to restart apps

blesswinsamuel:
This PR adds a command to restart apps, with an option to restart specific components only.

~TODO: update godo version in go.mod once godo is released with Restart method.~ (this is done)


andrewsomething (approved):
:+1:  LGTM

PR #1612, opened 2024-11-13
APPS-9859 Add doctl apps console command to start a console session to a component

blesswinsamuel:
This PR adds a `doctl apps console` command to start a console session to a component.

To try this out:
```
go build -o /tmp/doctl cmd/doctl/main.go
/tmp/doctl apps console <app-id> <component-name>
```

Note: The tests for this PR will pass only after https://github.com/digitalocean/godo/pull/754 is released and go.mod updated here.


commands/apps.go
```diff
@@ -695,6 +714,125 @@ func RunAppsGetLogs(c *CmdConfig) error {
...
+		}
+	})
+
+	grp.Go(func() error {
+		defer func() {
+			// An extra key press is needed to exit the console session because os.Stdin.Read blocks until a key is pressed.
+			// So an extra keystroke is required to unblock os.Stdin.Read and exit that goroutine.
+			fmt.Fprintln(c.Out)
+			fmt.Fprint(c.Out, "Press any key to exit.")
+		}()
```
blesswinsamuel, line 815: If we don't want the user to hit an extra keystroke to exit, we could make the `listener.ReadRawStdin(ctx, stdinCh)` goroutine outside of the errgroup. But it'd be a leaky goroutine, which is probably okay here since the CLI process exits after this function exits.

pkg/listen/listen.go
```diff
@@ -27,34 +30,32 @@ type Listener struct {
...
+	InCh <-chan []byte
 }
 
 // ListenerService listens to a websocket connection and outputs to the provided io.Writer
 type ListenerService interface {
-	Start() error
-	Stop()
+	Listen(ctx context.Context) error
+	ReadRawStdin(ctx context.Context, stdinCh chan<- string) error
+	MonitorResizeEvents(ctx context.Context, resizeEvents chan<- TerminalSize) error
```
blesswinsamuel, line 41: This is probably not the right place for these 2 functions. I'm open to ideas on refactoring this.

andrewsomething: They do feel a bit out of place. A few options:

1. Create a new package named something like `terminal` to hold utils related to `golang.org/x/term` usage
2. The listen package is only used by App Platform related commands. We can rename/redefine it to be more tightly coupled with the App Platform use case rather than trying to be a generic way to listen on a websocket. 
3. Leave it as is. They are used along side the listen package. So living along side that code keeps things quickly understandable

I'm leaning towards the first option. Thoughts?

andrewsomething (commented):
Looks great. Let's sort out the question about the stdin/resize functions, and we'll be good to go.

pkg/listen/listen.go
blesswinsamuel: Thanks. I liked option 1 as well. I have updated the PR. I had to add it to `doit` as well to mock it.

I'd have liked to add an integration test as well. But since the tests run in a non-interactive environment, executing this command from the integration test returns this error while it tries to set the terminal to raw mode:
```
Error: error setting terminal to raw mode: operation not supported by device
```

commands/apps.go
blesswinsamuel: I removed the need to hit an extra keystroke as it felt annoying from user perspective. I think the leaky goroutine is okay here since the CLI process exits after this function exits.

andrewsomething (approved):
LGTM! :rocket:

PR #1616, opened 2024-11-21
Fix conflicting aliases in apps commands

blesswinsamuel:
This PR changes the alias for:
- `console` command to `cs` (`l` is already used for `logs` command)
- `propose` command to `p` (`c` is already used for `create` command)


andrewsomething (approved):
:+1:  Good catch!

PR #1620, opened 2024-11-27
LBAAS: display LBs IPv6

jvasilevsky:
- update godo version
- display lb ipv6


loosla (approved):
:rocket: Looks good to me.

PR #1621, opened 2024-12-03
LBAAS: display LBs firewall rules

jvasilevsky:
display LBs firewall rules


loosla (approved):
:tada: Looks great :rocket: 
Thanks!

PR #1622, opened 2024-12-04
APPS: Add an --update-sources parameter to create and update app

markusthoemmes:
This adds a parameter named `--update-sources` to the `apps create` and `apps update` commands. If set, the respective update of the app spec also causes the underlying sources to be reevaluated and their respective latest state to be reflected in the created deployment.

On the `apps create` flow, this only applies to `--upsert` cases.


ElanHasson (approved):

loosla (approved):
:raised_hands: Looks good to me. Thanks for the update :+1:

PR #1625, opened 2024-12-05
Updates Firewalls create example with tags

dbrian57:
Addresses [user feedback](https://ideas.digitalocean.com/documentation/p/more-detail-needed-in-the-create-firewall-docs) about how to use tags when creating a firewall.


loosla:
Looks great! Thanks a lot for the update :rocket: 
Could you please also update [L71](https://github.com/digitalocean/doctl/blob/f49585eb0d79111d91d9d55852bd78acb861d8c1/commands/firewalls.go#L71) :pray: with

```
		tagNameRulesTxt := "A comma-separated list of existing tags, for example: frontend,backend,env:prod. Droplets with these tags will be placed behind the cloud firewall"
```
It may be useful to include an example with ":" to the `--tag-names` doc.
Many thanks.

loosla (approved):

PR #1626, opened 2024-12-07
fix: remove duplicate config init

wilkice:
1. there is `cobra.OnInitialize(initConfig)` in function `init` to do the configure init process
2. call of `initConfig` in this line makes no sense, as the `cfgFile` is blank at this time.


loosla (approved):
:raised_hands: Thanks for catching that and creating a PR to fix it!

Issue #1615, opened 2024-11-17
Reference an app by name instead of ID


dblock:
**What is the problem this feature would solve? Please describe.**

Apps in the UI are listed by name but in doctl I always have to figure out the UUID by going to the UX and copy-pasting it. 

Add support to using names instead of IDs in doctl.

**Describe the solution you'd like**

Currently to tail my logs I do this:

```
doctl apps logs uuid -f
```

I'd like to do this :

```
doctl apps logs app-name -f
```

**Additional context**

```
$ doctl version
doctl version 1.117.0-release
```

PR #1627, opened 2024-12-07
Added support getting logs for an app by name.

dblock:
I host [a bunch of apps](https://www.vestris.com/) on DO which is working really great. 

Trying to use the UX less. This change lists apps if you specify anything other than a UUID, I use this as `doctl apps logs slack-strava` vs. `doctl apps logs 4bf2666d-8fa9-4edd-8013-298ce3a8d52a` that I always have to get from the command line or UI.

Closes https://github.com/digitalocean/doctl/issues/1615.


andrewsomething (approved):
Thanks so much for this great contribution! :raised_hands:

PR #1630, opened 2024-12-11
refactor: move assign/unassign actions to reserved-ipv6 command

amanoj-do:
This PR removes `reserved-ipv6-actions` command and moves assign and unassign actions to `reserved-ipv6` command


loosla (approved):
:rocket: Looks good to me. Thanks for the update.

PR #1646, opened 2025-01-27
create partner interconnect attachment support

apinonformoso:
#### Verification

```sh
doctl network --type "partner" interconnect-attachment create --name "doctl-pia" --connection-bandwidth-in-mbps 50 --naas-provider "MEGAPORT" --region "nyc" --vpc-ids "d35e5cb7-7957-xxxx-xxxx-xxxx"
```

image


loosla:
Notes: 
[Related godo PR](https://github.com/digitalocean/godo/pull/769/files)
[PR that makes BGP field a pointer](https://github.com/digitalocean/godo/pull/777)

commands/partner_interconnect_attachments.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/digitalocean/doctl/do"
+	"github.com/digitalocean/godo"
+	"github.com/spf13/cobra"
+)
+
+// Partner creates the partner interconnect attachments command.
+func Partner() *Command {
+	cmd := &Command{
+		Command: &cobra.Command{
+			Use:     "partner",
```
andrewsomething, line 28: I'm curious about `partner` as a top level command. Do we expect there to be other "partner" features or products in the future? As this is so closely related to VPCs, I wonder if it shouldn't live under there? Where will this functionality live in the control panel?

commands/interconnect_attachment.go
```diff
@@ -0,0 +1,150 @@
+/*
```
guptado, line 1: lets prefix the file name with "partner_"

commands/interconnect_attachment.go
```diff
@@ -0,0 +1,150 @@
...
+	cmd.PersistentFlags().String(doctl.ArgInterconnectAttachmentType, "partner", "Specify interconnect attachment type (e.g., partner)")
+	viper.BindPFlag(strings.Join([]string{cmd.Use, doctl.ArgInterconnectAttachmentType}, "."), cmd.PersistentFlags().Lookup(doctl.ArgInterconnectAttachmentType))
+
+	cmd.AddCommand(PartnerInterconnectAttachments())
+
+	return cmd
+}
+
+// PartnerInterconnectAttachments creates the interconnect attachments command.
+func PartnerInterconnectAttachments() *Command {
```
guptado, line 48: can we also add tests for this?

commands/partner_interconnect_attachment.go
```diff
@@ -0,0 +1,150 @@
...
+		Command: &cobra.Command{
+			Use:     "network",
+			Short:   "Display commands that manage network products",
+			Long:    `The commands under ` + "`" + `doctl network` + "`" + ` are for managing network products`,
+			GroupID: manageResourcesGroup,
+		},
+	}
+
+	cmd.PersistentFlags().String(doctl.ArgInterconnectAttachmentType, "partner", "Specify interconnect attachment type (e.g., partner)")
+	viper.BindPFlag(strings.Join([]string{cmd.Use, doctl.ArgInterconnectAttachmentType}, "."), cmd.PersistentFlags().Lookup(doctl.ArgInterconnectAttachmentType))
```
andrewsomething, line 40: I think it would be better to attach this flag to the relevant commands rather than placing it here. Adding it as a persistent flag here means that it will be on any commands in this category whether it is relevant or not. For commands where it is relevant, it's now displayed under the `Global Flags` flags section in help output rather than with the other flags for the command making it easy to miss.

```
$ ./builds/doctl network interconnect-attachment create -
h
Use this command to create a new Partner Interconnect Attachment on your account.

Usage:
  doctl network interconnect-attachment create [flags]

Aliases:
  create, c

Examples:
The following example creates a Partner Interconnect Attachment: doctl network interconnect-attachment create --name "example-pia" --connection-bandwidth-in-mbps 50 --naas-provider "MEGAPORT" --region "nyc" --vpc-ids "c5537207-ebf0-47cb-bc10-6fac717cd672"

Flags:
      --bgp-local-asn int                  BGP Local ASN
      --bgp-local-router-ip string         BGP Local Router IP
      --bgp-peer-asn int                   BGP Peer ASN
      --bgp-peer-router-ip string          BGP Peer Router IP
      --connection-bandwidth-in-mbps int   Connection Bandwidth in Mbps (required)
      --format ID                          Columns for output in a comma-separated list. Possible values: ID, `Name`, `State`, `ConnectionBandwidthInMbps`, `Region`, `NaaSProvider`, `VPCIDs`, `CreatedAt`, `BGPLocalASN`, `BGPLocalRouterIP`, `BGPPeerASN`, `BGPPeerRouterIP`.
  -h, --help                               help for create
      --naas-provider string               NaaS Provider (required)
      --name string                        Name of the Partner Interconnect Attachment (required)
      --no-header                          Return raw data with no headers
      --region string                      Region (required)
      --vpc-ids strings                    VPC network IDs (required)

Global Flags:
  -t, --access-token string   API V2 access token
  -u, --api-url string        Override default API endpoint
  -c, --config string         Specify a custom config file (default "/home/asb/.config/doctl/config.yaml")
      --context string        Specify a custom authentication context name
      --http-retry-max int    Set maximum number of retries for requests that fail with a 429 or 500-level error (default 5)
      --interactive           Enable interactive behavior. Defaults to true if the terminal supports it (default true)
  -o, --output string         Desired output format [text|json] (default "text")
      --trace                 Show a log of network activity while performing a command
      --type string           Specify interconnect attachment type (e.g., partner) (default "partner")
  -v, --verbose               Enable verbose output
```

commands/partner_interconnect_attachment.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+// Network creates the network command.
+func Network() *Command {
+	cmd := &Command{
+		Command: &cobra.Command{
+			Use:     "network",
+			Short:   "Display commands that manage network products",
+			Long:    `The commands under ` + "`" + `doctl network` + "`" + ` are for managing network products`,
+			GroupID: manageResourcesGroup,
```
andrewsomething, line 35:
```suggestion
			GroupID: manageResourcesGroup,
			Hidden:  true,
```

What's the release status on this? If it is in private preview and only available to a small set of customers, we should hide it from top-level help output to prevent confusion.

andrewsomething (changes requested):
Over all this is looking good. One comment inline about the usage of the persistent flag. 

It would be great to get an integration test as well, e.g.:

https://github.com/digitalocean/doctl/blob/main/integration/vpc_create_test.go

commands/partner_interconnect_attachment.go
apinonformoso: Yes, still available to a small set of customers.
Addressed

andrewsomething (approved):
:+1:  LGTM!

PR #1649, opened 2025-01-27
add cluster autoscaler configuration for cluster create and update

dhij:
This PR adds the `scale-down-utilization-threshold` and the `scale-down-unneeded-time` flags for users to configure their cluster autoscaler configuration.

### DIFF
--- a/commands/kubernetes.go
+++ b/commands/kubernetes.go
@@ -292,9 +292,9 @@ After creating a cluster, a configuration context is added to kubectl and made a
 	AddStringSliceFlag(cmdKubeClusterCreate, doctl.ArgControlPlaneFirewallAllowedAddresses, "", nil,
 		"A comma-separated list of allowed addresses that can access the control plane.")
 	AddStringFlag(cmdKubeClusterCreate, doctl.ArgClusterAutoscalerScaleDownUtilizationThreshold, "", "",
-		"The threshold value for the Cluster Autoscaler's scale-down-utilization-threshold. It is the maximum value between the sum of cpu requests and sum of memory requests of all pods running on the node divided by node's corresponding allocatable resource, below which a node can be considered for scale down.")
+		"The threshold value for the cluster autoscaler's scale-down-utilization-threshold. It is the maximum value between the sum of CPU requests and sum of memory requests of all pods running on the node divided by node's corresponding allocatable resource, below which a node can be considered for scale down. To set the scale-down-utilization-threshold to 50%, pass the floating point value 0.5.")
 	AddStringFlag(cmdKubeClusterCreate, doctl.ArgClusterAutoscalerScaleDownUnneededTime, "", "",
-		"The unneed time for the Cluster Autoscaler's scale-down-unneeded-time. It defines how long a node should be unneeded before it is eligible for scale down.")
+		"The unneed time for the cluster autoscaler's scale-down-unneeded-time. It defines how long a node should be unneeded before it is eligible for scale down. To set the scale-down-unneeded-time to a minute and 30 seconds for example, pass the string '1m30s'.")
 	AddStringSliceFlag(cmdKubeClusterCreate, doctl.ArgTag, "", nil,
 		"A comma-separated list of `tags` to apply to the cluster, in addition to the default tags of `k8s` and `k8s:$K8S_CLUSTER_ID`.")
 	AddStringFlag(cmdKubeClusterCreate, doctl.ArgSizeSlug, "",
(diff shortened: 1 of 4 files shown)


commands/kubernetes.go
```diff
@@ -291,6 +291,10 @@ After creating a cluster, a configuration context is added to kubectl and made a
 		"Creates the cluster with control plane firewall enabled. Defaults to false. To enable the control plane firewall, supply --enable-control-plane-firewall=true.")
 	AddStringSliceFlag(cmdKubeClusterCreate, doctl.ArgControlPlaneFirewallAllowedAddresses, "", nil,
 		"A comma-separated list of allowed addresses that can access the control plane.")
+	AddStringFlag(cmdKubeClusterCreate, doctl.ArgClusterAutoscalerScaleDownUtilizationThreshold, "", "",
+		"The threshold value for the Cluster Autoscaler's scale-down-utilization-threshold. It is the maximum value between the sum of cpu requests and sum of memory requests of all pods running on the node divided by node's corresponding allocatable resource, below which a node can be considered for scale down.")
+	AddStringFlag(cmdKubeClusterCreate, doctl.ArgClusterAutoscalerScaleDownUnneededTime, "", "",
+		"The unneed time for the Cluster Autoscaler's scale-down-unneeded-time. It defines how long a node should be unneeded before it is eligible for scale down.")
```
bentranter, line 297: A few nitpicks:

* The other commands don't title-case words like "cluster" or "autoscaler". We should lowercase them to be consistent.
* "CPU" is an acronym, so it should be uppercase.
* The formats for these values aren't obvious to me, so it would be nice to include examples of valid input.
```suggestion
	AddStringFlag(cmdKubeClusterCreate, doctl.ArgClusterAutoscalerScaleDownUtilizationThreshold, "", "",
		"The threshold value for the cluster autoscaler's scale-down-utilization-threshold. It is the maximum value between the sum of CPU requests and sum of memory requests of all pods running on the node divided by node's corresponding allocatable resource, below which a node can be considered for scale down. To set the scale-down-utilization-threshold to 50%, pass the floating point value 0.5.")
	AddStringFlag(cmdKubeClusterCreate, doctl.ArgClusterAutoscalerScaleDownUnneededTime, "", "",
		"The unneed time for the cluster autoscaler's scale-down-unneeded-time. It defines how long a node should be unneeded before it is eligible for scale down. To set the scale-down-unneeded-time to a minute and 30 seconds for dxample, pass the string '1m30s'.")
```

bentranter (changes requested):
A few nitpicks about the documentation strings, but implementation looks good

loosla:
Nice work :rocket: 
Curious if we want to return these fields on `get` :thinking:

dhij:
@loosla it looks like we make a direct godo call to retrieve the cluster so we wouldn't need to do anything on the doctl side 🙂

commands/kubernetes.go
dhij: Thank you for the feedback, I added the changes in a new commit 👍

bentranter:
@dhij Anna brings up a good point though – there's no way to see what these fields are set to after you've created a cluster using them.

You'll need to update the displayer for the Kubernetes cluster (https://github.com/digitalocean/doctl/blob/main/commands/displayers/kubernetes.go) to display the value of these fields somehow. Since it's only two values, you can add each value as its own fields in the full Kubernetes cluster displayer. I've added comments to point out the new fields to make it more obvious:

```go
func (clusters *KubernetesClusters) Cols() []string {
	if clusters.Short {
		return []string{
			"ID",
			"Name",
			"Region",
			"Version",
			"AutoUpgrade",
			"Status",
			"NodePools",
		}
	}
	return []string{
		"ID",
		"Name",
		"Region",
		"Version",
		"AutoUpgrade",
		"HAControlPlane",
		"Status",
		"Endpoint",
		"IPv4",
		"ClusterSubnet",
		"ServiceSubnet",
		"Tags",
		"Created",
		"Updated",
		"NodePools",
		"Autoscaler.UtilizationThreshold", // New field.
		"Autoscaler.UnneededTime", // New field.
	}
}

func (clusters *KubernetesClusters) ColMap() map[string]string {
	if clusters.Short {
		return map[string]string{
			"ID":          "ID",
			"Name":        "Name",
			"Region":      "Region",
			"Version":     "Version",
			"AutoUpgrade": "Auto Upgrade",
			"Status":      "Status",
			"NodePools":   "Node Pools",
		}
	}
	return map[string]string{
		"ID":                              "ID",
		"Name":                            "Name",
		"Region":                          "Region",
		"Version":                         "Version",
		"AutoUpgrade":                     "Auto Upgrade",
		"HAControlPlane":                  "HA Control Plane",
		"ClusterSubnet":                   "Cluster Subnet",
		"ServiceSubnet":                   "Service Subnet",
		"IPv4":                            "IPv4",
		"Endpoint":                        "Endpoint",
		"Tags":                            "Tags",
		"Status":                          "Status",
		"Created":                         "Created At",
		"Updated":                         "Updated At",
		"NodePools":                       "Node Pools",
		"Autoscaler.UtilizationThreshold": "Autoscaler Scale Down Utilization", // New field.
		"Autoscaler.UnneededTime":         "Autoscaler Scale Down Unneeded Time", // New field.
	}
}

func (clusters *KubernetesClusters) KV() []map[string]any {
	out := make([]map[string]any, 0, len(clusters.KubernetesClusters))

	for _, cluster := range clusters.KubernetesClusters {
		tags := strings.Join(cluster.Tags, ",")
		nodePools := make([]string, 0, len(cluster.NodePools))
		for _, pool := range cluster.NodePools {
			nodePools = append(nodePools, pool.Name)
		}
		if cluster.Status == nil {
			cluster.Status = new(godo.KubernetesClusterStatus)
		}

		o := map[string]any{
			"ID":             cluster.ID,
			"Name":           cluster.Name,
			"Region":         cluster.RegionSlug,
			"Version":        cluster.VersionSlug,
			"AutoUpgrade":    cluster.AutoUpgrade,
			"HAControlPlane": cluster.HA,
			"ClusterSubnet":  cluster.ClusterSubnet,
			"ServiceSubnet":  cluster.ServiceSubnet,
			"IPv4":           cluster.IPv4,
			"Endpoint":       cluster.Endpoint,
			"Tags":           tags,
			"Status":         cluster.Status.State,
			"Created":        cluster.CreatedAt,
			"Updated":        cluster.UpdatedAt,
			"NodePools":      strings.Join(nodePools, " "),
		}

                // Add the new fields to the display output if they're set.
		if cfg := cluster.ClusterAutoscalerConfiguration; cfg != nil {
			o["Autoscaler.UtilizationThreshold"] = fmt.Sprintf("%d%%", int(*cfg.ScaleDownUtilizationThreshold*100))
			o["Autoscaler.UnneededTime"] = *cfg.ScaleDownUnneededTime
		}

		out = append(out, o)
	}

	return out
}
```

The output is already pretty long so this isn't great, but I think it's acceptable for now. If we end up having more autoscaler fields in the future, we'll most likely need to split it off into its own command separate from the `get` and `list` ones.

bentranter (approved):
LGTM!

PR #1651, opened 2025-01-28
load-balancers: make network-stack flag consistent

andrewsomething:
Happened to notice this and wanted to get it in before a release. Changes `network_stack` to `network-stack` to be consistent with other flags.


bentranter (approved):
🙇 Very nice catch, LGTM!

PR #1654, opened 2025-02-04
k8s: fix autoscaler integration test cases

bentranter:
Fixes the broken projects list test for the K8s resource, and adds a new k8s cluster get test to check for the newly added autoscaler fields.

This should fix the test failures introduced in #1649

### DIFF
--- a/commands/displayers/kubernetes.go
+++ b/commands/displayers/kubernetes.go
@@ -100,26 +100,32 @@ func (clusters *KubernetesClusters) KV() []map[string]any {
 		}
 
 		o := map[string]any{
-			"ID":             cluster.ID,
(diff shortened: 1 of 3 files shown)


andrewsomething (approved):
:+1:  LGTM!

Do we understand how the test failure was only found after merging to main?

dhij (approved):
Thank you 🙏

bentranter:
> 👍 LGTM!
> 
> Do we understand how the test failure was only found after merging to main?

Yeah, it looks like it's from caching the integration test runs between commits in the same PR. Here's a screenshot from the integration test run on the PR that introduced the failure:
image

PR #1659, opened 2025-02-17
k8s: update upgrade command documentation

jkpe:
👋 Just a minor correction to the description of how to rerieve the list of available versions.

`doctl k8s get-upgrades k8s-lon1` should be `doctl k8s cluster get-upgrades k8s-lon1`


andrewsomething (approved):
Good catch! Thanks @jkpe! :pray:

PR #1687, opened 2025-04-25
LBAAS: add tls cipher policy as load balanacer arg

jvasilevsky:
add ability to set load balancer tls cipher policy via doctl

### DIFF
--- a/vendor/github.com/digitalocean/godo/load_balancers.go
+++ b/vendor/github.com/digitalocean/godo/load_balancers.go
@@ -26,6 +26,10 @@ const (
 	// Load Balancer network_stack types
 	LoadBalancerNetworkStackIPv4      = "IPV4"
 	LoadBalancerNetworkStackDualstack = "DUALSTACK"
+
+	// Supported TLS Cipher policies
+	LoadBalancerTLSCipherPolicyDefault = "DEFAULT"
+	LoadBalancerTLSCipherPolicyStrong  = "STRONG"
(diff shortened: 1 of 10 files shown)


vendor/github.com/digitalocean/godo/load_balancers.go
```diff
@@ -26,6 +26,10 @@ const (
 	// Load Balancer network_stack types
 	LoadBalancerNetworkStackIPv4      = "IPV4"
 	LoadBalancerNetworkStackDualstack = "DUALSTACK"
+
+	// Supported TLS Cipher policies
+	LoadBalancerTLSCipherPolicyDefault = "DEFAULT"
```
kishlay-singh-DO, line 31: Hey @jvasilevsky I was looking at the API docs for [Load balancers ](https://docs.digitalocean.com/reference/api/digitalocean/#tag/Load-Balancers/operation/loadBalancers_create) and it seems there we allow 3 values for tls cipher policy (STRONG, WEAK & DEFAULT) but in the code we are allowing only 2 values (STRONG & DEFAULT). Could you please look into this?

jvasilevsky: Let me update the API docs. The Weak value can be assigned, but it currently does not do anything/is equivalent to default. Let me remove it from the api docs.

kishlay-singh-DO (approved):

PR #1691, opened 2025-05-07
Enhancement: Pull ubuntu-18 builder image only when the app requires ubuntu-18

fumblehool:
## Details

For the command: `doctl apps dev build`, all builder images are pulled before the app is built.
This results in unnecessary bandwidth and storage usage, as it pulls builders for both `ubuntu-22` and `ubuntu-18` stacks, even though only one is actually used during the build.

With this change, doctl will pull the `ubuntu-18` builder image only if the AppSpec explicitly specifies this stack.

### DIFF
--- a/commands/apps_dev.go
+++ b/commands/apps_dev.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"strings"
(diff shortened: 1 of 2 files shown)


danaelhe (approved):
thank you!

PR #1697, opened 2025-05-16
Updates App create description with repo access instruction

dbrian57:
Resolves [PDOCS-2951](https://do-internal.atlassian.net/browse/PDOCS-2951). Updates App create command description to provide links that give AP access to various services.


harshmaru7:
LGTM ! let me check the failing tests once !

PR #1709, opened 2025-06-09
projects: Add support for retrieving floating IPv6 from URN

amanoj-do:
This minor change allow searching `floatingipv6` URNs searchable from `projects resources get` command


kishlay-singh-DO (approved):
LGTM. We'll try to work on a better error message to show when an invalid resource type is passed. RN we just show
`
Error: "xyz" is an invalid resource type, consult the documentation
`

PR #1714, opened 2025-06-25
Fix vpc nat gateway default timeout values

StephenVarela:
- Fix default tcp timeout for vpc-nat-gateways
- Make type flag optional as it defaults to public


commands/vpc_nat_gateways.go
```diff
@@ -29,7 +29,7 @@ You can use vpc-nat-gateway to perform CRUD operations on a VPC NAT Gateway.`,
 		cmdVPCNATGatewayUpdate,
 	} {
 		AddStringFlag(c, doctl.ArgVPCNATGatewayName, "", "", "Name of the VPC NAT Gateway", requiredOpt())
-		AddStringFlag(c, doctl.ArgVPCNATGatewayType, "", "PUBLIC", "Gateway type", requiredOpt())
+		AddStringFlag(c, doctl.ArgVPCNATGatewayType, "", "PUBLIC", "Gateway type")
```
StephenVarela, line 32: It is a required field on the API but since we are using a default to public since thats the only one supported to the public, i removed the required flag 

Let me know if you think otherwise

anup-deka (approved):
LGTM

PR #1717, opened 2025-07-11
Added Agent route operations

SSharma-10:
Implemented Agent Route operations:

1. Add Agent Route to an Agent
2. Update Agent Route for an Agent
3. Delete Agent Route for an Agent

### DIFF
--- a/args.go
+++ b/args.go
@@ -771,6 +771,6 @@ const (
 	// ArgAgentRouteName is the name of the route.
 	ArgAgentRouteName = "route-name"
 
-	// ArgAgentRouteIfCase is the type of the route.
+	// ArgAgentRouteIfCase is the if-case condition for the route.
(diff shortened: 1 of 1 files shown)


args.go
```diff
@@ -758,4 +758,19 @@ const (
...
+	ArgChildAgentId = "child-agent-id"
+
+	// ArgAgentRouteId is the UUID for the agent linkage.
+	ArgAgentRouteId = "route-id"
+
+	// ArgAgentRouteName is the name of the route.
+	ArgAgentRouteName = "route-name"
+
+	// ArgAgentRouteIfCase is the type of the route.
+	ArgAgentRouteIfCase = "if-case"
```
anup-deka, line 775: Change the comment for ArgAgentRouteIfCase

PR #1751, opened 2025-09-12
Added commands to get scheduled job invocations

fumblehool:
## Details:

This PR depends on godo: https://github.com/digitalocean/godo/pull/900

Changes - 
* It adds command `doctl apps list-job-invocations <app id>` to list job-invocations for an app.
* It adds command `doctl apps get-job-invocation <app id> <job invocation id>` to get a job-invocation.
* Updates command: `doctl apps logs <app id> <component name> --job-invocation` to get logs for a given job-invocation id.


commands/apps.go
```diff
@@ -188,6 +188,32 @@ Only basic information is included with the text output format. For complete app
...
+		RunAppsListJobInvocations,
+		"list-job-invocations <app id>",
+		"List all job invocations",
+		`List all job invocations for an app.
+
+Only basic information is included with the text output format. For complete app details including the app specs, use the JSON format.`,
+		Writer,
+		aliasOpt("lsji"),
+		displayerType(&displayers.JobInvocations{}),
+	)
```
blesswinsamuel, line 202: I believe we're missing some docs for job names and deployment id argument.

Something like this (see restart command):
```
AddStringSliceFlag(restartApp, doctl.ArgAppComponents, "", nil, "The components to restart. If not provided, all components are restarted.")
```

blesswinsamuel: would be good to also add an example

blesswinsamuel: I think we should also add some docs in logs command that it supports job invocation id

blesswinsamuel (approved):

kishlay-singh-DO (approved):

PR #1755, opened 2025-09-18
Adding integration tests as a pre requirement for releases

kishlay-singh-DO:
This change runs integration tests before we mark a release as latest.


harshmaru7 (approved):
LGTM

PR #1783, opened 2025-10-24
[WIP]  Implement --access-key flag and serverless keys commands

vrindavinod-do:
Adding the auth-key flag implementation:

## 1. Our access-key implementation
doctl serverless connect my-namespace --access-key "dof_v1_abc123:secretkey"
 → Validates access-key (OpenWhisk)
 → Stores in credentials.json: {"currentHost": "api.host", "currentNamespace": "my-namespace", "credentials": {"api.host": {"my-namespace": {"api_key": "dof_v1_abc123:secretkey"}}}}

## 2. All other commands automatically work  
doctl serverless functions list
 → CheckServerlessStatus() ✅ (finds credentials.json)
 → ReadCredentials() ✅ (gets YOUR access-key)
 → initWhisk() ✅ (creates OpenWhisk client with YOUR access-key)
 → API calls succeed ✅

doctl serverless deploy ./my-function  
 → Same credential flow ✅
 → Deploy uses YOUR access-key ✅

New commands:
doctl serverless key create --name "my-key"                    # Create new access key
doctl serverless key list                                     # List all keys  
doctl serverless key revoke <key-id>


commands/serverless.go
```diff
@@ -282,6 +341,41 @@ func RunServerlessConnect(c *CmdConfig) error {
 	return connectFromList(ctx, sls, list, c.Out)
 }
 
+// validateAccessKeyFormat validates that the access key follows the expected format
+func validateAccessKeyFormat(accessKey string) error {
```
logwolvy: Since, access-key will be sent by out backend, do we really need to validate its format?

commands/displayers/access_keys.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// KV implements Displayable.
+func (ak *AccessKeys) KV() []map[string]any {
+	out := make([]map[string]any, 0, len(ak.AccessKeys))
+
+	for _, key := range ak.AccessKeys {
+		// Show full secret during creation, hidden otherwise
+		secret := "<hidden>"
+		if key.Secret != "" && ak.ShowFullSecret {
```
logwolvy: In addition to this, I am assuming the list/get calls to access keys will not have the secret field, right?

commands/keys.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/digitalocean/doctl/do"
+	"github.com/spf13/cobra"
+)
+
+// Keys generates the serverless 'keys' subtree for addition to the doctl command
+func Keys() *Command {
+	cmd := &Command{
+		Command: &cobra.Command{
+			Use:   "key",
+			Short: "Manage access keys for functions namespaces",
```
logwolvy: Not immediately but eventually we could expose key update, kind of like https://docs.digitalocean.com/reference/doctl/reference/spaces/keys/update/

commands/displayers/access_keys.go
vrindavinod-do: yes, the secret will only be shown if create command is called

commands/serverless.go
logwolvy: I see this is being used in `connect` command. My initial thought was that we needed a stronger check to differentiate between legacy and new access keys. We simply don't want users to pass legacy access keys to the new connect command/flag and then suddenly break their environments due to deprecation. If this prevents that, let's keep it.

commands/keys.go
```diff
@@ -0,0 +1,192 @@
...
+  doctl serverless key list
+  doctl serverless key list --namespace fn-abc123`,
+		Writer, aliasOpt("ls"), displayerType(&displayers.AccessKeys{}))
+	AddStringFlag(list, "namespace", "", "", "target namespace (uses connected namespace if not specified)")
+
+	delete := CmdBuilder(cmd, RunAccessKeyDelete, "delete <key-id>", "Deletes an access key",
+		`Permanently deletes an existing access key. This action cannot be undone.
+
+Examples:
+  doctl serverless key delete dof_v1_a1b2c3d4e5f67890
```
logwolvy: Is it possible to specify <access_key_id> or equivalent in description or one of the examples?

vrindavinod-do:
> What will be the behaviour for other commands (deploy, fn list...) after deprecation of legacy keys, if no access keys were added by the user via doctl connect cmd?

if users haven't added access keys via the new [doctl serverless connect --access-key] command, all serverless operations will fail with:
Error: serverless support is not connected to a functions namespace (use `doctl serverless connect`)

do/serverless.go
```diff
@@ -1474,3 +1495,45 @@ func validateFunctionLevelFields(serverlessAction *ServerlessFunction) ([]string
...
+// DeleteNamespaceAccessKey deletes an access key from the specified namespace
+func (s *serverlessService) DeleteNamespaceAccessKey(ctx context.Context, namespace string, keyID string) error {
+	path := fmt.Sprintf("v2/functions/namespaces/%s/keys/%s", namespace, keyID)
+	req, err := s.client.NewRequest(ctx, http.MethodDelete, path, nil)
+	if err != nil {
+		return err
+	}
+	_, err = s.client.Do(ctx, req, nil)
+	return err
+}
```
DO-rrao: Hi Vrinda, can you please help me with understanding if we are asking for a confirmation from the user before dispaying access key? Or we are just displaying it once during creation, following which it is shown as hidden>?

PR #1827, opened 2026-03-30
droplets: add --public-networking flag to create command

anitgandhi:
## Notes

Requires:
  - https://github.com/digitalocean/godo/pull/977


kishlay-singh-DO (approved):
LGTM

PR #1831, opened 2026-04-14
NFRNC-702: Update request and response schema to support custom model in Dedicated Inference

divyanshgupta-dodeveloper:
NFRNC-702: Update request and response schema to support custom model in Dedicated Inference

Jira: https://do-internal.atlassian.net/browse/NFRNC-702


sjana-do (approved):
LGTM

Issue #1782, opened 2025-10-22
doctl apps spec get not returning authority correctly


stigrune:
**Describe the Issue:**
I'm using `doctl apps spec get <id> --format json > spec.json` 

I'm observing that the authority in the response is `{}`
```
  "ingress": {
    "rules": [
      {
        "match": {
          "path": {
            "prefix": "/"
          },
          "authority": {}
        },
        "component": {
          "name": "pixelate"
        }
      }
    ]
  },
```
But in the DigitalOcean Web UI there is an authority added on the app. 

<img width="476" height="334" alt="Image" src="https://github.com/user-attachments/assets/57878668-5e5f-4d47-a87a-46c26dea544e" />

I would expect the doctl apps spec get command to return the same values that are 
displayed in the web UI.

**Environment:**
 - doctl version: 1.146.0-release
 - OS: Tested both Windows 11 and ubuntu-latest via github actions.
 - Installation method: winget, and `uses: digitalocean/action-doctl@v2`

**Additional Details:**
Also tested yaml with same results, the authority is not set in the spec. 

This is a new app created by the Web UI, with deployment source “Container image”, pointing to a DigitalOcean Container Registry image. No manual changes have been done to the spec file.

anup-deka:
Thanks for raising this issue @stigrune

aupadhyay-shark:
Bug fix is released -
https://github.com/digitalocean/doctl/pull/1839

PR #1833, opened 2026-04-15
doctl apps spec get not returning authority correctly #1782

aupadhyay-shark:
Root Cause
The bug is in the AppIngressSpecRuleStringMatch struct in godo (v1.184.0), defined in vendor/github.com/digitalocean/godo/apps.gen.go:

type AppIngressSpecRuleStringMatch struct {
    Prefix string `json:"prefix,omitempty"`
    Exact  string `json:"exact,omitempty"`
}

Fix
Changed Prefix and Exact from string to *string in AppIngressSpecRuleStringMatch

Issue - https://github.com/digitalocean/doctl/issues/1782
https://github.com/digitalocean/godo/issues/937

### DIFF
--- a/commands/apps_test.go
+++ b/commands/apps_test.go
@@ -955,6 +955,289 @@ services:
 	})
 }
 
+func TestRunAppSpecGet_AuthoritySerialization(t *testing.T) {
+	exactEmpty := ""
+	exactDomain := "example.com"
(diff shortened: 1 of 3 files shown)


aupadhyay-shark:
Need to only update mod file

PR #1834, opened 2026-04-24
Prevent archive link-based path traversal in extract package

DO-rrao:
## Summary
A crafted tar archive containing a hardlink or symlink pointing outside the extraction root can be used to overwrite arbitrary files on the host filesystem. Since extract.Extract is called on archives downloaded from the network during doctl serverless install/upgrade, this is exploitable in practice if a malicious archive is served.
The existing HasPrefix path check only validates where a file is created, not where a link points to. This change adds validation of link targets for both hardlinks and symlinks before they are created on disk.

## Test plan

- [x] doctl serverless install-succeeds end-to-end with real archives
- [x] Malicious archives (hardlink escape, symlink escape, write-through chains) are rejected without modifying files outside the extraction root
- [x] Valid symlinks and hardlinks within the extraction root still work

## Ticket 
https://do-internal.atlassian.net/browse/VULN-3739


pkg/extract/extract.go
```diff
@@ -90,20 +120,32 @@ func extractTarGz(source, target string) error {
 
 			err = f.Chmod(info.Mode())
 			if err != nil {
+				f.Close()
```
sjana-do, line 123: f.close() file handling should also be replicated in extractZip() method as well
https://github.com/digitalocean/doctl/blob/c352a085e42392971e47ae3057f45a3302c42d9d/pkg/extract/extract.go#L187

sjana-do (changes requested):

pkg/extract/extract.go
DO-rrao: done

sjana-do (approved):
LGTM

PR #1835, opened 2026-04-24
DOKS SSO: per-cluster config, kubeconfig type

d-honeybadger:
A few related changes for DOKS SSO. The project is not available to the public yet, so, breaking changes in the sso-specific flags are okay.

1. DOKS now expects per-cluster SSO configuration (i.e. each cluster gets an SSO client ID and issuer url). Adding the corresponding flags
2. There's no more reason to differentiate between "enabled" and "required" values not being set by the customer vs being set to "false". Hence, these 2 fields are getting changed from bool pointer to regular bool
3. Customers can request kubeconfigs with SSO as the auth method, rather than the token-based kubeconfigs. Added a flag "type" to the commands that fetch kubeconfigs.


commands/nfs.go
```diff
@@ -256,8 +256,13 @@ func nfsSnapshotCreate(c *CmdConfig) error {
 		return err
 	}
 
+	actionID, err := strconv.Atoi(action.ID)
```
d-honeybadger, line 259: All changes to nfs command and tests are a drive-by: they were necessary b/c of a breaking change in godo: https://github.com/digitalocean/godo/pull/989

args.go
```diff
@@ -166,9 +166,13 @@ const (
 	// ArgEnableRDMASharedDevicePlugin enables automatic NVIDIA gpu device plugin installation.
 	ArgEnableRDMASharedDevicePlugin = "enable-rdma-shared-device-plugin"
 	// ArgKubernetesEnableSSO enables SSO as a method of authentication for the DOKS cluster.
-	ArgKubernetesEnableSSO = "enable-sso"
+	ArgKubernetesEnableSSO = "sso-enabled"
```
d-honeybadger, line 169: Renaming for consistency with the other sso flags

commands/kubernetes.go
```diff
@@ -1186,35 +1204,60 @@ func (s *KubernetesCommandService) RunKubernetesClusterListAssociatedResources(c
...
-	_, err = c.Out.Write(kubeconfig)
-	return err
+	kubeconfig, err := s.KubeconfigProvider.Remote(c.Kubernetes(), kubeconfigParams)
+	if err != nil {
+		return err
+	}
+	if kubeconfig == nil {
+		return fmt.Errorf("received nil kubeconfig")
+	}
+	kubecnfigBytes, err := clientcmd.Write(*kubeconfig)
```
thearyanahmed:
```suggestion
    kubeconfigBytes, err := clientcmd.Write(*kubeconfig)
```

commands/kubernetes.go
```diff
@@ -1186,35 +1204,60 @@ func (s *KubernetesCommandService) RunKubernetesClusterListAssociatedResources(c
...
+	if err != nil {
+		return fmt.Errorf("serializing kubeconfig: %w", err)
+	}
+	_, err = c.Out.Write(kubecnfigBytes)
```
thearyanahmed:
```suggestion
	_, err = c.Out.Write(kubeconfigBytes)
```

thearyanahmed (approved):

thearyanahmed (approved):

commands/kubernetes.go
```diff
@@ -2364,14 +2397,26 @@ func mergeKubeconfig(clusterID string, remote, local *clientcmdapi.Config, setCu
 		local.CurrentContext = remote.CurrentContext
 	}
 
+	remoteAuthInfo, ok := remote.AuthInfos[remoteCtx.AuthInfo]
+	if !ok {
+		// this is a bug in the backend, we received incomplete/non-sensical data
+		return fmt.Errorf("The remote config has no no auth info for user %q. This is a bug, please open a ticket with DigitalOcean.",
```
DO-rrao: "no auth info" instead of "no no auth info"

d-honeybadger: Fixed, thank you!

DO-rrao (approved):

PR #1840, opened 2026-05-07
kubernetes exec-credential command for OIDC Authorization flow with PKCE

d-honeybadger:
Adds a set of flags to `doctl cluster kubernetes kubeconfig exec-credential` command to be able to handle OIDC authorization.
Currently, `doctl kubernetes kubeconfig exec-credential` is used to fetch a DO PAT from DOKS API, and this PAT is used by kubectl to authenticate inside the k8s cluster.
The additional flags enable  `doctl kubernetes kubeconfig exec-credential` to instead receive a token (ID token) from an OIDC idendity provider such as Okta, Auth0.

The authorization flow implemented here is similar to how OSS kubectl plugins do it (e.g. https://github.com/int128/kubelogin) but with our own implementation we get seamless UX (users don't have to install and configure third-party plugins) and flexibility in adjusting & fixing any issies.

### DIFF
--- a/commands/doit.go
+++ b/commands/doit.go
@@ -140,7 +140,7 @@ func initConfig() {
 }
 
 // in case we ever want to change this, or let folks configure it...
-func defaultConfigHome() string {
+var defaultConfigHome = func() string {
 	cfgDir, err := os.UserConfigDir()
 	checkErr(err)
 
--- a/commands/kubernetes.go
+++ b/commands/kubernetes.go
@@ -488,7 +488,7 @@ Returns the raw YAML for the specified cluster's kubeconfig.`, Writer, aliasOpt(
 	cmdShowConfig.Example = `The following example shows the kubeconfig YAML for a cluster named ` + "`" + `example-cluster` + "`" + `: doctl kubernetes cluster kubeconfig show example-cluster`
 
 	execCredDesc := "INTERNAL: This hidden command is for printing a cluster's exec credential"
-	cmdExecCredential := CmdBuilder(cmd, k8sCmdService.RunKubernetesKubeconfigExecCredential, "exec-credential <cluster-id>", execCredDesc, execCredDesc, Writer) //, hiddenCmd())
+	cmdExecCredential := CmdBuilder(cmd, k8sCmdService.RunKubernetesKubeconfigExecCredential, "exec-credential <cluster-id>", execCredDesc, execCredDesc, Writer, hiddenCmd())
 	AddStringFlag(cmdExecCredential, doctl.ArgVersion, "", "", "")
 	AddStringFlag(cmdExecCredential, doctl.ArgKubernetesSSOIssuerURL, "", "", "")
 	AddStringFlag(cmdExecCredential, doctl.ArgKubernetesSSOClientID, "", "", "")
--- a/commands/kubernetes.go
+++ b/commands/kubernetes.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"log"
 	"os"
 	"path/filepath"
 	"sort"
@@ -28,12 +29,14 @@ import (
 	"github.com/blang/semver"
 	"github.com/digitalocean/godo"
 	"github.com/google/uuid"
+	"github.com/pkg/browser"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 
 	"github.com/digitalocean/doctl"
 	"github.com/digitalocean/doctl/commands/displayers"
 	"github.com/digitalocean/doctl/do"
+	"github.com/digitalocean/doctl/internal/kubernetes/sso"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kubeerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -60,6 +63,14 @@ A typical workflow is to use ` + "`" + `doctl kubernetes cluster create` + "`" +
 
 The commands under ` + "`" + `doctl kubernetes options` + "`" + ` retrieve values used while creating clusters, such as the list of regions where cluster creation is supported.`
 )
+
+func init() {
+	// Borrowed from https://github.com/int128/kubelogin:
+	// In credential plugin mode, some browser launcher writes a message to stdout
+	// and it may break the credential json for client-go.
+	// This prevents the browser launcher from breaking the credential json.
+	browser.Stdout = os.Stderr
+}
 
 var getCurrentAuthContextFn = defaultGetCurrentAuthContextFn
 
@@ -207,6 +218,9 @@ func (p *kubeconfigProvider) ConfigPath() string {
 // KubernetesCommandService is used to execute Kubernetes commands.
 type KubernetesCommandService struct {
 	KubeconfigProvider KubeconfigProvider
+
+	// to be used for stubbing in testss
+	ssoLogin func(ctx context.Context, clientID, issuerURL string, opts ...sso.LocalOIDCLoginOption) (string, time.Time, error)
 }
 
 func kubernetesCommandService() *KubernetesCommandService {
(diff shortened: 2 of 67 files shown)


internal/kubernetes/sso/auth_success.html
```diff
@@ -0,0 +1,162 @@
+<!DOCTYPE html>
```
d-honeybadger, line 1: AI-generated as I'm no html/css expert
This is what it looks like 

Image

internal/kubernetes/sso/auth_error.html
```diff
@@ -0,0 +1,164 @@
+<!DOCTYPE html>
```
d-honeybadger, line 1: AI-generated as I'm not an htlm/css expert
Here's what it looks like 

Image

commands/doit.go
```diff
@@ -140,7 +140,7 @@ func initConfig() {
 }
 
 // in case we ever want to change this, or let folks configure it...
-func defaultConfigHome() string {
+var defaultConfigHome = func() string {
```
d-honeybadger, line 143: So that it can be stubbed in tests.

commands/kubernetes.go
```diff
@@ -1264,10 +1282,13 @@ func cachedExecCredentialPath(id string) string {
 	return filepath.Join(kubeconfigCachePath(), id+".json")
 }
 
+func cachedSSOExecCredentialPath(id string) string {
```
d-honeybadger, line 1285: Separate files for SSO tokens

internal/kubernetes/sso/sso.go
```diff
@@ -0,0 +1,261 @@
...
+func (t *localOIDCLogin) loginURL() string {
+	return fmt.Sprintf("http://localhost:%d/login", t.port)
+}
+
+func (t *localOIDCLogin) getAuthCode(ctx context.Context) (string, error) {
+	var group errgroup.Group
+
+	server := &http.Server{
+		Handler: t.ssoServer,
+		Addr:    fmt.Sprintf(":%d", t.port),
```
sjana-do: ":8080" listens on every interface, the browser only needs to reach localhost. Suggested to use fmt.Sprintf("127.0.0.1:%d", t.port)

internal/kubernetes/sso/sso.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+func (t *localOIDCLogin) getIDToken(ctx context.Context) (string, time.Time, error) {
+	code, err := t.getAuthCode(ctx)
+	if err != nil {
+		return "", time.Time{}, fmt.Errorf("getting authorization code: %w", err)
+	}
+
+	t.logger.Println("Received an authorization code, exchanging for ID token")
+	token, err := t.oauth2Config.Exchange(ctx, code, oauth2.S256ChallengeOption(t.codeVerifier), oauth2.VerifierOption(t.codeVerifier))
```
sjana-do: Is CodeChallenge required to be passed in token exchange step?

internal/kubernetes/sso/sso.go
```diff
@@ -0,0 +1,261 @@
...
+	if err != nil {
+		return "", time.Time{}, fmt.Errorf("exchanging authorization code for ID token: %w", err)
+	}
+
+	idToken, ok := token.Extra("id_token").(string)
+	if !ok || idToken == "" {
+		return "", time.Time{}, errors.New("no ID token found")
+	}
+
+	return idToken, token.Expiry, nil
```
sjana-do, line 217: Can we verify the ID token locally before returning it. We are pulling id_token out of the token response and forward it to kubectl as-is — no signature, iss, aud, or exp check

internal/kubernetes/sso/sso.go
d-honeybadger: Good point, thanks!

internal/kubernetes/sso/sso.go
d-honeybadger: Yes. Maybe some IDPs don't expect it (code challenge should be enough?) but Auth0 throws me an error if the verifier is not provided:
`Error: Failed to get ID token: exchanging authorization code for ID token: oauth2: "invalid_request" "Parameter 'code_verifier' is required"`

internal/kubernetes/sso/sso.go
d-honeybadger: This command doesn't authorize the user to do anything though, it just returns a token that will be passed to kube-apiserver which will perform all the necessary verifications. I added a simple verification, but I think we can be pretty minimal here and not duplicate kube-apiserver's job - WDYT?

commands/kubernetes.go
```diff
@@ -473,8 +488,11 @@ Returns the raw YAML for the specified cluster's kubeconfig.`, Writer, aliasOpt(
 	cmdShowConfig.Example = `The following example shows the kubeconfig YAML for a cluster named ` + "`" + `example-cluster` + "`" + `: doctl kubernetes cluster kubeconfig show example-cluster`
 
 	execCredDesc := "INTERNAL: This hidden command is for printing a cluster's exec credential"
-	cmdExecCredential := CmdBuilder(cmd, k8sCmdService.RunKubernetesKubeconfigExecCredential, "exec-credential <cluster-id>", execCredDesc, execCredDesc, Writer, hiddenCmd())
+	cmdExecCredential := CmdBuilder(cmd, k8sCmdService.RunKubernetesKubeconfigExecCredential, "exec-credential <cluster-id>", execCredDesc, execCredDesc, Writer) //, hiddenCmd())
```
RohithSangati: If we’re making this command non-hidden, should we also update the execCredDesc accordingly?

internal/kubernetes/sso/sso.go
RohithSangati: @d-honeybadger  I agree the verifier is required, but as per the PKCE standard I don’t think we need to pass code_challenge and code_challenge_method in the token request, which are currently being passed as the third argument in this function call.

As per the spec, the token exchange request should include the code_verifier, while code_challenge and code_challenge_method are only used during the authorization request.

Reference for token request:
https://www.oauth.com/oauth2-servers/pkce/authorization-code-exchange/

What are your thoughts?

RohithSangati:
> Adds a set of flags to `doctl kubernetes kubeconfig exec-credential` command to be able to handle OIDC authorization. Currently, `doctl kubernetes kubeconfig exec-credential` is used to fetch a DO PAT from DOKS API, and this PAT is used by kubectl to authenticate inside the k8s cluster. The additional flags enable `doctl kubernetes kubeconfig exec-credential` to instead receive a token (ID token) from an OIDC idendity provider such as Okta, Auth0.
> 
> The authorization flow implemented here is similar to how OSS kubectl plugins do it (e.g. https://github.com/int128/kubelogin) but with our own implementation we get seamless UX (users don't have to install and configure third-party plugins) and flexibility in adjusting & fixing any issies.

Just a small clarification — should the command here be `doctl kubernetes cluster kubeconfig exec-credential` instead of `doctl kubernetes kubeconfig exec-credential`, since from the code it looks like `kubeconfig` is a child of the `cluster` command? I might be misunderstanding something, so wanted to confirm once.

commands/kubernetes.go
d-honeybadger: ! Thanks for catching this
I didn't mean to un-hide this command. I commented it out for debugging at some point and never removed the comment slashes.

internal/kubernetes/sso/sso.go
d-honeybadger: Indeed, verified that the flow works just fine without the challenge, thank you!

d-honeybadger:
> Just a small clarification — should the command here be `doctl kubernetes cluster kubeconfig exec-credential` instead of `doctl kubernetes kubeconfig exec-credential`, since from the code it looks like `kubeconfig` is a child of the `cluster` command? I might be misunderstanding something, so wanted to confirm once.

Correct, I forgot the `cluster` part in this PR's description (it's a loong subcommand :smile: ). Fixed!

sjana-do (approved):

centromere:
How do I disable these log messages?:
```
doctl: 2026/05/21 14:15:10 Using cached credential
```

PR #1846, opened 2026-05-21
Add serverless Inference endpoints support

SSharma-10:


commands/serverless_inference.go
```diff
@@ -0,0 +1,941 @@
...
+	var b strings.Builder
+	for _, choice := range completion.Choices {
+		if choice.Message.Content != nil {
+			b.WriteString(*choice.Message.Content)
+		}
+	}
+	return b.String()
+}
+
+func chatCompletionText(completion *godo.ChatCompletion) string {
```
anup-deka: Unused function

commands/serverless_inference.go
```diff
@@ -0,0 +1,941 @@
...
+func Inference() *Command {
+	cmd := &Command{
+		Command: &cobra.Command{
+			Use:     "inference",
+			Aliases: []string{"serverless-inference", "si"},
+			Short:   "Call DigitalOcean serverless inference APIs",
+			Long: `The subcommands of doctl inference call the serverless inference API at https://inference.do-ai.run.
+
+Authenticate using --access-token. The value may be a model access key or a DigitalOcean personal access token with full access; all scopes must be granted for the serverless inference API to work.`,
+			GroupID: manageResourcesGroup,
```
anup-deka: I was just wondering if inference needs another section like under Serverless Inference , not a blocker just a thought 😅 

Image

commands/serverless_inference.go
```diff
@@ -0,0 +1,941 @@
...
+	create := CmdBuilder(cmd, RunServerlessInferenceResponsesCreate, "create", "Create a response",
+		`Use --request for a full JSON body, or --model with --input.`, Writer)
+	AddStringFlag(create, doctl.ArgInferenceModel, "m", "", "Model ID")
+	AddStringFlag(create, doctl.ArgInferenceInput, "", "", "Input text (required unless --request is set)")
+	AddStringFlag(create, doctl.ArgInferenceRequest, "", "", "Path to JSON request body. Use \"-\" for stdin.")
+	AddBoolFlag(create, doctl.ArgInferenceStream, "", false, "Stream using server-sent events")
+	AddStringFlag(create, doctl.ArgInferenceInstructions, "", "", "Optional instructions")
+
+	create.Example = `doctl inference responses create --model MODEL --input "Hello"
+doctl inference responses create --model MODEL --input "Hello" --stream`
```
anup-deka: For `--model MODEL` can we provide an model example like you did above for eg  `--model openai-gpt-oss-20b` try doing it for all examples

commands/serverless_inference.go
```diff
@@ -0,0 +1,941 @@
...
+		Command: &cobra.Command{
+			Use:   "responses",
+			Short: "Create responses",
+			Long:  "POST /v1/responses.",
+		},
+	}
+
+	create := CmdBuilder(cmd, RunServerlessInferenceResponsesCreate, "create", "Create a response",
+		`Use --request for a full JSON body, or --model with --input.`, Writer)
+	AddStringFlag(create, doctl.ArgInferenceModel, "m", "", "Model ID")
```
anup-deka: `Model ID (required unless --request is set)` would be needed here as well, check for all occurrences

commands/serverless_inference.go
SSharma-10: Added a new `Inference` section and moved `Serverless Inference` under it. In the future, we will also add Agent Inference and Batch Inference, following the same pattern as openAPI

commands/serverless_inference.go
```diff
@@ -0,0 +1,937 @@
...
+		return err
+	}
+	jsonOut := strings.EqualFold(Output, "json")
+	return runServerlessInferenceStream(c, stream, func() error {
+		ev := stream.Current()
+		if jsonOut {
+			return writeServerlessInferenceJSON(c.Out, ev)
+		}
+		if ev.B64JSON != "" {
+			_, err := fmt.Fprintf(c.Out, "[%s] partial image index %d\n", ev.Type, ev.PartialImageIndex)
```
anup-deka: Can you check `--stream` for image generation once? It should work also with `--output someimage.png` right?

commands/serverless_inference.go
SSharma-10: Done

commands/serverless_inference.go
SSharma-10: added

commands/serverless_inference.go
SSharma-10: yes, it should work, added a fix. TY!

commands/serverless_inference.go
```diff
@@ -0,0 +1,1017 @@
...
+func serverlessInferenceImageParams(c *CmdConfig) (*godo.ImageGenerateParams, error) {
+	requestPath, err := serverlessInferenceRequestPath(c)
+	if err != nil {
+		return nil, err
+	}
+	if requestPath != "" {
+		params := new(godo.ImageGenerateParams)
+		if err := readServerlessInferenceJSON(requestPath, params); err != nil {
+			return nil, err
+		}
```
anup-deka, line 445: 🤔 
```suggestion
		}
		// Ensure N is at least 1 and less than equal to 10
		if params.N < 1 {
			params.N = 1
		}
		if params.N > 10 {
			return nil, fmt.Errorf("--%s must be between 1 and 10", params.N)
		}
```

commands/serverless_inference.go
```diff
@@ -0,0 +1,1017 @@
...
+	}
+	stream, err := serverlessInferenceStream(c)
+	if err != nil {
+		return err
+	}
+	// Also respect "stream": true embedded in a --request JSON body.
+	if !stream && params.Stream != nil && *params.Stream {
+		stream = true
+	}
+	if stream {
```
anup-deka, line 425: If someone adds `--stream` with the `--request` but request config doesn't have `"stream":true` should we still enable streaming? Based on the checks it looks like it, then the below changes should be there

```suggestion
	if stream {
	    params.Stream = godo.PtrTo(true)
```

commands/serverless_inference.go
```diff
@@ -0,0 +1,1017 @@
...
+	}
+
+	create := CmdBuilder(cmd, RunServerlessInferenceAsyncCreate, "create", "Start an async invocation",
+		`Starts an asynchronous job for a fal model and returns a request ID. Use --model and --prompt for image or audio generation, --text for text-to-speech, or --request for a full JSON body. Poll the result using the get subcommand.`, Writer)
+	AddStringFlag(create, doctl.ArgInferenceModel, "m", "", "fal model ID (maps to model_id in the API)")
+	AddStringFlag(create, doctl.ArgInferencePrompt, "", "", "Prompt for image or audio generation (input.prompt)")
+	AddStringFlag(create, doctl.ArgInferenceText, "", "", "Text for text-to-speech generation (input.text)")
+	AddIntFlag(create, doctl.ArgInferenceSecondsTotal, "", 0, "Audio duration in seconds (input.seconds_total)")
+	AddStringFlag(create, doctl.ArgInferenceRequest, "", "", "Path to JSON request body. Use \"-\" for stdin.")
+	AddStringSliceFlag(create, doctl.ArgTag, "", nil, "Tag in key=value form (repeatable)")
```
anup-deka: Can we update the messages, clarify which field is required for image, text-to-speech and image generation
Or maybe add this link like for more info - https://docs.digitalocean.com/reference/api/reference/serverless-inference/#inference_create_async_invoke

commands/serverless_inference.go
SSharma-10: good catch! Updated.

anup-deka (approved):

PR #1859, opened 2026-06-08
agents: decode SPI event stream, surface API errors, ack input

logwolvy:
doctl agent attach connected but printed nothing: the harness-api SSE stream emits the SPI canonical event envelope (type/data/timestamp/ tenant_id, dot-separated event names) while vendored godo decoded kind/payload/at/team_id, so every event fell through renderEvent.

- godo HostedAgentEvent: decode the SPI wire via UnmarshalJSON and switch the HostedAgentEventKind constants to the dot-separated SPI names.
- godo ErrorResponse: also read the nested {"error":{code,message}} envelope so failed calls show the server's reason, not a bare status.
- agents: align event payload structs/rendering and HITL tracking to the SPI data shapes, and acknowledge each submit with its run id so the agent's startup latency isn't mistaken for a hang (which caused duplicate submits).
- tests for the wire decode, rendering, nested-error surfacing, and ack.

The vendored godo edits mirror the fix being upstreamed to digitalocean/godo and are in-tree until godo is re-vendored.

### DIFF
--- a/commands/agents.go
+++ b/commands/agents.go
@@ -282,12 +282,12 @@ func RunAgentsAttach(c *CmdConfig) error {
 			case godo.HostedAgentEventKindHITLRequested:
 				var p hitlRequestedPayload
 				if err := json.Unmarshal(ev.Payload, &p); err == nil {
-					pending.set(p.Request.RequestID)
+					pending.set(p.HitlID)
 				}
 			case godo.HostedAgentEventKindHITLResolved:
 				var p hitlResolvedPayload
 				if err := json.Unmarshal(ev.Payload, &p); err == nil {
-					pending.clearIf(p.Decision.RequestID)
+					pending.clearIf(p.HitlID)
 				}
 			}
 			renderEvent(c.Out, ev)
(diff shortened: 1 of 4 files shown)


commands/agents.go
```diff
@@ -484,39 +498,40 @@ func renderEvent(w io.Writer, ev godo.HostedAgentEvent) {
...
+	if len(payload) > 0 {
+		// MarshalIndent sorts map keys, so output is stable run-to-run.
+		if b, err := json.MarshalIndent(payload, "  ", "  "); err == nil {
+			fmt.Fprintf(w, "  %s\n", b)
 		}
 	}
-	fmt.Fprintf(w, "  request_id: %s\n", req.RequestID)
-	fmt.Fprintf(w, "  (resolve with `/a %s`, `/r %s`, or `/d %s`)\n", req.RequestID, req.RequestID, req.RequestID)
+	fmt.Fprintf(w, "  hitl_id: %s\n", hitlID)
+	fmt.Fprintf(w, "  (resolve with `/a %s`, `/r %s`, or `/d %s`)\n", hitlID, hitlID, hitlID)
```
SSharma-10, line 536: does the event's hitl_id match what POST /v2/agents/sessions/<id>/hitl/<request_id> expects as <request_id>? 

the pre-vendor code used HostedAgentHITLRequest.RequestID, this PR feeds hitl_id straight into that path. If harness-api treats them as different ids, /a /r /d will 404 silently

logwolvy:
```
curl -X POST \
  https://<host>/v2/agents/sessions/sess_123/hitl/hitl_ab12cd34 \
  -H "Authorization: Bearer <token>" \
  -H "Content-Type: application/json" \
  -d '{
    "outcome": "HITL_OUTCOME_APPROVE",
    "reason": "looks safe",
    "source": "RESOLUTION_SOURCE_OUT_OF_BAND"
  }'
```

#### Valid enum values

```
outcome (wire.HITLOutcome):
HITL_OUTCOME_APPROVE
HITL_OUTCOME_REJECT
HITL_OUTCOME_DEFER
(HITL_OUTCOME_UNSPECIFIED is rejected)

source (wire.ResolutionSource):
RESOLUTION_SOURCE_INLINE_KEYSTROKE
RESOLUTION_SOURCE_OUT_OF_BAND (default)
```

#### Responses

204 No Content - decision accepted.
400 Bad Request - invalid JSON, or missing/unspecified outcome.
500 Internal Server Error - resolution failed downstream.

This is the request/response schema for HITL endpoint

PR #1862, opened 2026-06-11
fix --format and --no-header on gradient commands

m3co-code:
This PR contains two fixes:
- `--format` and `--no-header` was ignored on `doctl gradient list-models` and `list-regions`
- `--format ID` was not working on the list, even though the header was advertising ID, Id was expected which is inconsistent with the header and the rest of doctl, hence having moved to ID consistently

### DIFF
--- a/commands/gradientai_list_regions.go
+++ b/commands/gradientai_list_regions.go
@@ -4,8 +4,8 @@ import (
 	"github.com/digitalocean/doctl/commands/displayers"
 )
 
-func ListRegionsCmd() *Command {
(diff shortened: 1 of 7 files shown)


commands/gradientai_list_regions.go
```diff
@@ -4,8 +4,8 @@ import (
 	"github.com/digitalocean/doctl/commands/displayers"
 )
 
-func ListRegionsCmd() *Command {
```
DO-rrao, line 7: list-models and list-regions are now registered twice
After this PR, both commands appear duplicated in doctl gradient --help:

m3co-code: Thx @DO-rrao! Good catch. I'll spend a little bit more time on that.

PR #1865, opened 2026-06-25
nfs access point

v-amanjain-afk:


go.mod
```diff
@@ -135,3 +135,5 @@ require (
 	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 )
+
+replace github.com/digitalocean/doctl => ../doctl
```
rak16: Please remove this before merging

rak16 (commented):
Overall, when we `list` / `get` access points - it should show multiple VPC's that are attached to the default access point `/`

rak16 (changes requested):

go.mod
v-amanjain-afk: done

rak16 (approved):

DO-rrao:
Tested this end-to-end (built the binary, ran the commands against prod, and checked the real API responses). Two things came up that are worth a look:

**1. The default access point's VPC IDs only show up if you pass `--region`.**
The API doesn't return any VPC info for the default access point, so doctl works around that by looking up the share to fill in the `VPC IDs` column — but that share lookup needs a region. Since `--region` is optional on `get`/`list`, anyone who leaves it off just sees a blank `VPC IDs` column for the default AP, which is a little confusing. Could we either make `--region` required for `get`/`list`, or have the default AP return its VPCs directly so the extra lookup isn't needed?

**2. Small flag-naming inconsistency in the `nfs` group.**
The new access-point commands use `--vpc-id` (with a dash), while the existing `attach`/`detach` commands use `--vpc_id` (with an underscore). The dashed version is the usual doctl convention, so it's probably worth switching the older flags over now, before more folks start depending on them.

Neither is a blocker — #1 is the one I'd lean toward fixing before merge.

DO-rrao (approved):

PR #1873, opened 2026-07-10
added byoip support for load balancer and nat-gateway

rbhatia-code:


commands/load_balancers.go
```diff
@@ -104,6 +104,8 @@ With the load-balancer command, you can list, create, or delete load balancers,
 	AddStringFlag(cmdLoadBalancerCreate, doctl.ArgLoadBalancerNetworkStack, "", "", "The network stack type determines the allocation of ipv4/ipv6 addresses to the load balancer, e.g.: `IPV4` or `DUALSTACK`"+
 		" (NOTE: this feature is in private preview, contact DigitalOcean support to review its public availability.)")
```
DO-rrao, line 107: nit: Small inconsistency between the two new bits: the NAT gateway path guards with if ip != "" while the LB path just does r.IP = ip unconditionally. Both are correct today (the field is omitempty), but if someone ever wires --ip into the LB update command down the line, that unconditional assignment would quietly start sending ip on updates, which the API treats as create-only. A if ip != "" guard (or a one-line comment noting it's create-only) would make that intent obvious and future-proof.

DO-rrao (approved):
LGTM

PR #1878, opened 2026-07-15
doctl agents proxy w/codex - M0-M2 

julia-ye:
Implementation plan here (from Adil Hafeez): https://github.internal.digitalocean.com/digitalocean/cthulhu/pull/166147/files
Has milestones that this is based on (M0-M2)
Comment: there is no `turn/failed` method in the codex app-server protocol -> run.failed maps to the same turn/completed notification as run.completed, with turn.status == "failed" and turn.error populated ({message: string, ...}) instead of a separate method.

Jira: https://do-internal.atlassian.net/browse/MARSOHS-440

### DIFF
--- a/internal/agentproxy/codex/facade.go
+++ b/internal/agentproxy/codex/facade.go
@@ -45,8 +45,8 @@ type Facade struct {
 
 	// Sessions is the harness bridge: turn/start calls SendInput on it, and
 	// the background streaming goroutine reads StreamSession's SSE events to
-	// translate into codex notifications. Required once turn/start is used;
-	// M0/M1 methods (initialize, thread/start, etc.) don't touch it.
+	// translate into codex notifications. The bootstrap methods (initialize,
+	// thread/start, etc.) don't touch it.
 	Sessions do.HostedAgentsService
 
 	// notifier is set once per connection by the bridge (see SetNotifier) and
@@ -93,9 +93,12 @@ func (f *Facade) SetNotifier(n agentproxy.Notifier) {
 // TUI over WebSocket: {"result":{"userAgent":...,"codexHome":...,
 // "platformFamily":...,"platformOs":...}}. codexHome/platformFamily/platformOs
 // describe codex's own local runtime environment in a real app-server; since
-// v1 proxies to a hosted sandbox rather than running codex locally, these are
-// synthetic placeholders until the harness exposes the sandbox's real values
-// (M1+), not a reflection of anything the sandbox actually reports today.
+// this facade proxies to a hosted sandbox rather than running codex locally,
+// these are synthetic placeholders, not a reflection of anything the sandbox
+// actually reports.
+//
+// TODO(M3-M5): replace with the sandbox's real values once the harness
+// exposes them.
 type initializeResult struct {
 	UserAgent      string `json:"userAgent"`
 	CodexHome      string `json:"codexHome"`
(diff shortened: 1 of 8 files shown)


internal/agentproxy/codex/facade.go
```diff
@@ -0,0 +1,705 @@
...
+// doc comment for why there must be exactly one StreamSession call, ever, no
+// matter how many turns run one after another or overlap.
+func (f *Facade) trackTurn(ctx context.Context, runID string) {
+	f.mu.Lock()
+	if f.turns == nil {
+		f.turns = make(map[string]*turnState)
+	}
+	f.turns[runID] = &turnState{itemID: runID + "-msg"}
+	startLoop := !f.streamStarted
+	f.streamStarted = true
```
SSharma-10: I think there's a small edge case here. If `StreamSession` fails to open (or `runEventLoop` exits because of EOF/notify write failure), `streamStarted` never gets reset.

After that, any new turn/start request will still return inProgress, but no notifications are actually sent, so Codex ends up hanging

Should we reset `streamStarted` when the event loop exits, or only set it after `StreamSession` has been established successfully?

internal/agentproxy/codex/facade.go
```diff
@@ -0,0 +1,705 @@
...
+			}
+			text.WriteString(item.Text)
+		}
+		resp, err := f.Sessions.SendInput(f.SessionID, &godo.HostedAgentSendInputRequest{Text: text.String()})
+		if err != nil {
+			return nil, &agentproxy.RPCError{Code: -32000, Message: "SendInput failed: " + err.Error()}
+		}
+		// The harness's run id doubles as this facade's turn id — already a
+		// unique per-turn identifier, no separate id scheme needed.
+		f.trackTurn(ctx, resp.RunID)
```
SSharma-10: Looks like `SendInput` is called before the SSE stream is up (trackTurn → async runEventLoop → StreamSession)

On a live harness, this could race with early events like `run.started` or a fast completion. We don't see it in tests because the test harness only starts emitting after /stream is connected

Would it make sense to establish the stream before calling `SendInput`?

internal/agentproxy/proxy.go
```diff
@@ -0,0 +1,209 @@
...
+		connCtx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		notifier := &wsNotifier{conn: conn}
+		if na, ok := facade.(NotifierAware); ok {
+			na.SetNotifier(notifier)
+		}
+
+		log.Printf("agentproxy: client connected from %s", r.RemoteAddr)
+		handleConn(connCtx, conn, facade, notifier)
```
SSharma-10, line 146: One thing I noticed: the same Facade is reused across WebSocket connections, but only the notifier gets replaced. Combined with the sticky streamStarted state, disconnecting and reconnecting (codex --remote) never restarts the SSE loop

Would it make more sense to create a new Facade per connection, or reset the stream/turn state when the client disconnects?

internal/agentproxy/codex/facade.go
julia-ye: Went with first option of resetting on event loop exit. Event loop seems more likely to cause problems. 

Fix in facade.go:

- `runEventLoop` has `defer f.stopEventLoop() `as its first line, so the reset happens on every exit path
- Added `stopEventLoop` to clear `streamStarted` under `f.mu `so the next turn/start reopens a fresh stream, and fails out any turns still in `f.turns `at that moment (turn/completed with status: "failed")
- Added TestFacade_TurnStart_ReopensStreamAfterPreviousLoopEnded as regression coverage. Verified it times out on the old code and passes with the fix. 
One subtlety in the test: a turn started in the narrow window between turn/completed and the loop actually hitting EOF can still race stopEventLoop, so the test polls the (unexported, same-package) streamStarted field to deterministically wait for the loop to fully stop before asserting recovery, rather than assuming it's instant.

internal/agentproxy/codex/facade.go
julia-ye: Did a two part fix: 

Part 1 (what you suggested): `turn/start` now calls `ensureEventLoop(ctx)` before `SendInput`. `ensureEventLoop` opens `StreamSession `synchronously and only spawns the reader `goroutine` once that succeeds, so the SSE reader is always attached before `SendInput` can trigger the harness to start emitting anything.

Part 2 (found while fixing part 1): turned up an second race. Since `trackTurn` (which registers runID in f.turns) can only run after `SendInput` returns, there's now a window where the event-loop `goroutine` can see events for a run before it's registered as tracked. Fixed by having the loop's lookup (`lookupTurn`) retry briefly (up to 200ms, polling every 2ms) before treating an unrecognized run id as genuinely stale. In practice this resolves in microseconds, since it's just waiting for trackTurn's map write on the other goroutine.

Also split trackTurn into ensureEventLoop (stream lifecycle) + trackTurn (turn bookkeeping only), and runEventLoop now takes the already-opened stream instead of opening it itself. Verified with -race, 3x repeated (-count=3) plus 15 manual repeats, all green.

internal/agentproxy/proxy.go
julia-ye: Went with a new Facade per connection.

- `Serve/ServeListener` now take `newFacade func() Facade` instead of a single instance, called once per accepted connection. Every connection (including a reconnect) gets a fresh `turns/streamStarted/notifier`. Updated the `commands/agents.go` call site to pass a constructor closure.

- Added TestWire_ReconnectGetsFreshFacadeState (connect → turn → disconnect → reconnect → turn, over a real ServeListener). Passes with -race, 5x repeated.

SSharma-10:
LGTM

SSharma-10 (approved):

PR #1853, opened 2026-05-28
Open Harness Server subcommands

SSharma-10:

PR #1898, opened 2026-07-30
agents: resolve ${VAR} env references in --spec manifests locally

adilhafeez:
## Summary

- `doctl agents start --spec` (and `agents triggers create/update --spec`) now resolve `${VAR}` references in the manifest from the local environment before uploading, so specs like `OPENAI_API_KEY: ${OPENAI_API_KEY}` work instead of shipping the literal string into the sandbox.
- Only the strict braced form `${NAME}` expands; bare `$VAR`, `$(...)`, and other shell-isms embedded in manifests (skills instructions, prompts) pass through untouched. `$${VAR}` escapes to a literal `${VAR}`.
- Referencing a variable that is not set locally is a hard error listing the missing names, rather than a silent empty substitution.
- Help text for `agents start` and the trigger `--spec` flags documents the behavior, including that fresh-trigger manifests are stored expanded server-side.

Based on #1853 (`feat/agents-subcommands`), which this PR targets as its base.

## Test plan

- [x] New `TestExpandManifestEnv` table tests: set/empty/unset variables, missing-name error de-dup, `$${VAR}` escape, bare-dollar passthrough
- [x] New `TestReadManifest/expands_env_references` integration case
- [x] `go test ./commands/` passes (except pre-existing keychain-dependent `TestRegistryLogin`/`TestRegistryLogout` failures, unrelated)
- [x] `go vet ./commands/` clean


SSharma-10 (approved):
LGTM

PR #1899, opened 2026-07-31
agents: read session events from the data-plane /events endpoint (stack on feat/agents-subcommands)

logwolvy:
Puts `doctl agents attach` and `doctl agents logs` back on the OHP data plane's `/v2/agents/sessions/{id}/events`, on a branch of its own so these builds can be tested against the data plane without moving either shared PR branch.

Pairs with digitalocean/godo#1072. Merge that first — this PR pins a pseudo-version of its head commit, which should be swapped for a tagged godo release before this lands.

### Why this is a restore rather than new work

The doctl-side changes were never reverted outright; they were dismantled piecemeal on `feat/agents-subcommands` (#1853) while the endpoint was unavailable:

- `9532be7d` re-pinned godo to the released `v1.202.0-beta.1`, which dropped `/events` out of `vendor/` entirely.
- `0438544f` moved the agentproxy test harness back to the control plane's `/stream`, switched the reconnect test's resume cursor back to a `replay_from` query parameter, and replaced godo's exported `stream.state` symbols with locally-declared copies so the package would still compile against that pin.

The help-text changes from the original commits survived and are untouched here.

### What this PR does

**`agents: pin godo to the data-plane /events build`** — repoints `go.mod` at godo#1072 and re-vendors. The only diff under `vendor/` is the 89 lines `9532be7d` had removed from `hosted_agents.go`; no transitive dependency moved.

**`agents: read live and replay streams from the data-plane /events endpoint`** — undoes the three accommodations:

- `commands/agents.go` uses godo's `HostedAgentEventKindStreamState` / `HostedAgentStreamState` instead of local copies, so the wire contract lives in one place.
- The reconnect test reads the resume cursor from `Last-Event-ID`, which is where the live lane carries it. `replay_from` stays the cursor for replay-only reads, which are a separate lane with pagination rather than resume semantics.
- The agentproxy harness serves `/events` and opens every stream with a `stream.state` frame, so the codex facade tests exercise a stream shaped like the real one. `/stream` is deliberately left unregistered: no agentproxy caller makes a replay-only read, so a request landing there is a bug worth failing on.

### Testing

`go build ./...`, `go vet ./commands/... ./internal/...` and `go test ./commands/... ./internal/...` all pass, including every `agents` test and the eight `TestStreamWithReconnect_*` cases. The `doctl` binary builds and runs.

Two pre-existing failures on `feat/agents-subcommands` are unchanged by this PR, and both reproduce on the base branch: `TestRegistryLogout` fails on local keychain access, and `integration/database_create_restore_from_cluster.go` does not compile under `go build ./...` (it is a non-`_test.go` file referencing test-only symbols).


ElanHasson (approved):

SSharma-10:
doctl side looks good, create/get/attach work and /events returns stream.state 
but attach still stays on `waiting for the agent` with no output events and agents logs is empty. This looks like a backend issue

SSharma-10 (approved):
LGTM

PR #1903, opened 2026-08-06
agents: page session history backwards for bounded replay

mishraishika24-pixel:
A replay request no longer returns a session's whole transcript: the server caps it at its newest window and reports whether older events remain, so a client that wants full history has to walk backwards for it a page at a time.

do.LoadSessionHistory does that walk and stitches the pages back into chronological order, since they arrive newest-group first. Both readers of stored history go through it: `agents logs` and the codex proxy's --replay. Neither can print as it reads anymore -- the oldest events arrive from the last request -- so history is buffered before it renders.

`agents logs` gains --tail to stop the walk early, --page-size to size each request, and --before to read a single page and report the cursor for the one behind it. Its help text no longer claims a long session may replay only its recent activity; paging handles the per-request cap, and what remains unrecoverable is retention.

The vendored godo edits (before/limit query params, and an SSEReader comment hook so the has_more trailer is readable) mirror the change being upstreamed to digitalocean/godo and are in-tree until godo is re-vendored.


SSharma-10:
Can you re-vendor from OHS_endpoints and remove the hand-patched `vendor/github.com/digitalocean/godo` changes?

commands/agents.go
```diff
@@ -1129,51 +1139,96 @@ func hitlOutcomeFor(s string) (godo.HostedAgentHITLOutcome, error) {
...
+	}
 	stylingEnabled = detectStyling()
 
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	stream, err := c.HostedAgents().StreamSession(ctx, sessionID, &godo.HostedAgentSessionStreamOptions{
-		ReplayOnly: true,
+	if before != "" {
+		return runAgentsLogsPage(ctx, c, sessionID, before, pageSize)
```
SSharma-10, line 1178: nit: if both `--before` and `--tail` are set, this path wins and `--tail` is ignored. Might be worth rejecting that combination or calling it out in the flag help

PR #1904, opened 2026-08-10
MARSOHS-867 fix JSON output contract violations on triggers verbs

sdharavath:
## Summary

- **MARSOHS-867 (`create` / `rotate-secret`):** The webhook-secret banner and URL were written to stdout unconditionally. With `-o json` this non-JSON text precedes the object and breaks any JSON parser at character 0. Fix: route the banner to stderr when `Output == "json"`. `rotate-secret` additionally now emits `{"webhook_secret":"..."}` to stdout so programmatic callers get machine-readable output.

- **MARSOHS-868 (`get-execution`):** Same root cause — the run-output text block (`e.OutputText` + truncation notice + blank line) was written to stdout before `c.Display(...)` regardless of output format. Fix: suppress the block from stdout when `-o json` (the structured JSON object already carries the data).

- **MARSOHS-869 (list displayers):** All four `JSON()` methods in `displayers/agent_triggers.go` had a `len == 1` special-case that returned a bare object instead of a one-element array. List endpoints must always return an array. Fix: remove the special-case from `HostedAgentTrigger`, `HostedAgentTriggerExecution`, `HostedAgentWebhookProvider`, and `HostedAgentReusableSession`.

The pattern used for the banner redirect already existed in the same file (`printNextPageToken` checks `Output == "json"` and writes to `os.Stderr`); it just wasn't applied to the new banner-printing code.

## Test plan

- [ ] `doctl agents triggers create --kind webhook ... -o json | python3 -m json.tool` — parses cleanly; banner appears on stderr
- [ ] `doctl agents triggers rotate-secret <id> -o json | python3 -m json.tool` — emits `{"webhook_secret":"..."}` to stdout; banner on stderr
- [ ] `doctl agents triggers get-execution <tid> <eid> -o json | python3 -m json.tool` — parses cleanly; run output suppressed from stdout
- [ ] `doctl agents triggers list -o json` with exactly 1 trigger — returns `[{...}]` not `{...}`
- [ ] `doctl agents triggers list-executions <id> -o json` with exactly 1 execution — returns `[{...}]`

## Proof of testing

- [x] Manually tested locally
- [ ] Unit tested
- [ ] Integration tested

### DIFF
--- a/commands/agent_triggers.go
+++ b/commands/agent_triggers.go
@@ -215,7 +215,7 @@ func RunAgentTriggersCreate(c *CmdConfig) error {
 	if result.Trigger == nil {
 		return nil
 	}
-	return c.Display(&displayers.HostedAgentTrigger{Triggers: []do.HostedAgentTrigger{*result.Trigger}})
+	return c.Display(&displayers.HostedAgentTrigger{Triggers: []do.HostedAgentTrigger{*result.Trigger}, Single: true})
 }
 
 // RunAgentTriggersGet fetches one trigger.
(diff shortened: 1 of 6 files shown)


SSharma-10 (approved):
LGTM

PR #1906, opened 2026-08-10
MARSOHS-867 include webhook_secret in JSON output for triggers create

sdharavath:
## What

Fixes a data-loss gap in the `doctl agents triggers create -o json` output contract.

The previous fix (merged in #1904) correctly routed the human-readable secret banner to stderr in JSON mode, but the JSON payload itself never contained `webhook_secret`. Any consumer parsing `-o json` output would silently lose the one-time secret.

## Why

As Joe Keegan noted in the ticket thread: *"just know for the doctl json output, its likely not as easy as just removing the initial plain text that is output, as that data may not be in the following JSON output. This is especially the case for when you create a trigger and the secret is only included in the plain text output."*

## How

In JSON mode, `RunAgentTriggersCreate` now bypasses the displayer and encodes an inline struct that embeds all `HostedAgentTrigger` fields **plus** `webhook_secret` at the top level, producing a single flat JSON object:

```json
{
  "trigger_id": "tr_abc123",
  "name": "gh-ci",
  "kind": "webhook",
  ...
  "webhook_secret": "sec_once"
}
```

The human-readable banner (secret notice) is still sent to stderr in JSON mode, preserving the existing UX for interactive use.

Text mode is unchanged.

## Test

- `TestAgentTriggersCreateWebhook_JSONMode` now asserts `webhook_secret` and `trigger_id` are present in the parsed JSON object (not just that the output is syntactically valid JSON).
- All existing trigger tests pass.

## Proof of testing

- [x] `go test ./commands/ -run TestAgentTriggers` — all 13 tests pass

### DIFF
--- a/commands/agent_triggers.go
+++ b/commands/agent_triggers.go
@@ -202,14 +202,32 @@ func RunAgentTriggersCreate(c *CmdConfig) error {
 	if err != nil {
 		return err
 	}
-	if result.WebhookSecret != "" {
-		noticeOut := c.Out
-		if Output == "json" {
-			noticeOut = os.Stderr
+	if Output == "json" {
+		// In JSON mode the one-time secret must be in the machine-readable output,
+		// not only in a human-readable banner that gets routed to stderr.
+		// Embed it alongside all trigger fields in a single flat JSON object so
+		// consumers can always parse the secret without screen-scraping.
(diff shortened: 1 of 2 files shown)


SSharma-10 (approved):
LGTM

PR #1907, opened 2026-08-12
MARSOHS-867 suppress webhook secret banner on stderr in -o json mode

sdharavath:
## Summary

Follow-up to #1904 / #1906. [MARSOHS-867](https://do-internal.atlassian.net/browse/MARSOHS-867) was reopened because the human-readable "Webhook secret (shown once...)" banner was still visible when running `doctl agents triggers create|rotate-secret -o json` interactively. That was stdout/stderr interleaving on a shared TTY, not a contract violation (the banner was already on stderr, stdout was pure JSON) — but since the secret already travels inside the JSON payload, echoing it a second time to stderr in `-o json` mode is redundant and risks duplicating a secret into logs that capture stderr.

- `RunAgentTriggersCreate` / `RunAgentTriggersRotateSecret`: drop the stderr banner entirely when `Output == "json"`. Text mode (default) is unchanged — banner still prints on stdout for interactive users.
- Tests: `TestAgentTriggersCreateWebhook_JSONMode` / `TestAgentTriggersRotateSecret_JSONMode` now swap the real `os.Stderr` for a pipe (`captureProcessStderr`) and assert it's empty, so this can't regress silently again — asserting against the injectable `CmdConfig.Out` alone wasn't catching this class of bug.

## Test plan

- [x] `go build ./commands/...` — clean
- [x] `gofmt -l commands/agent_triggers.go commands/agent_triggers_test.go` — clean
- [x] `go test ./commands/... -run Trigger -v` — all pass, including the two tightened JSON-mode tests
- [x] Built the real `doctl` binary (`go build -mod=vendor -o doctl-dev ./cmd/doctl`) and drove it as a subprocess against a local mock DO API server (`httptest.Server`), capturing stdout/stderr on separate streams (not `CombinedOutput`) to reproduce the exact reported repro:
  - `agents triggers create --kind webhook ... -o json` → stdout: 698 bytes pure JSON incl. `webhook_secret`; stderr: 0 bytes
  - `agents triggers rotate-secret <id> -o json` → stdout: 86 bytes pure JSON incl. `webhook_secret`; stderr: 0 bytes
  - `agents triggers rotate-secret <id>` (text mode, regression check) → banner still prints on stdout as before

Full output posted on the Jira ticket.

[MARSOHS-867]: https://do-internal.atlassian.net/browse/MARSOHS-867?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commands/agent_triggers.go
+++ b/commands/agent_triggers.go
@@ -203,13 +203,9 @@ func RunAgentTriggersCreate(c *CmdConfig) error {
 		return err
 	}
 	if Output == "json" {
-		// In JSON mode the one-time secret must be in the machine-readable output,
-		// not only in a human-readable banner that gets routed to stderr.
-		// Embed it alongside all trigger fields in a single flat JSON object so
-		// consumers can always parse the secret without screen-scraping.
-		if result.WebhookSecret != "" {
-			fmt.Fprint(os.Stderr, displayers.FormatWebhookSecretNotice(result.WebhookSecret))
-		}
+		// JSON mode emits nothing but the document: the one-time secret travels
+		// in the payload, so re-printing it as a human banner would only
+		// duplicate a secret onto a second stream that callers may capture.
 		type jsonCreateResult struct {
 			*godo.HostedAgentTrigger
 			WebhookSecret string `json:"webhook_secret,omitempty"`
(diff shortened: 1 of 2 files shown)


SSharma-10 (approved):
LGTM

PR #1908, opened 2026-08-12
MARSOHS-887 add regression tests for agents list JSON array shape

sdharavath:
## Summary

[MARSOHS-887](https://do-internal.atlassian.net/browse/MARSOHS-887) (filed by Joe) reports that `doctl agents list -o json` returns a bare JSON object instead of a one-element array when exactly one session exists — the exact same bug class as [MARSOHS-869](https://do-internal.atlassian.net/browse/MARSOHS-869) on `agents triggers list`, which was fixed via the `Single` bool pattern in `displayers/agent_triggers.go`.

**Investigation finding:** the underlying fix for `agents list` was already shipped, bundled into #1904 (see the second-to-last commit on that PR, "fix same JSON list-shape bug in HostedAgentSession displayer") — `RunAgentsList` already omits `Single` (defaults `false` → always an array), and `RunAgentsGet`/`RunAgentsCreate` already set `Single: true` (bare object). Verified live with the real binary against a mock API server at 0/1/2 rows — all three shapes are correct.

So the code fix predates this ticket; what was missing was **regression coverage** — there was no test pinning the array-vs-object contract for `HostedAgentSession`, unlike the trigger displayers which got dedicated shape tests in the same PR. This PR closes that gap:

- Add `commands/displayers/hosted_agents_test.go`: `HostedAgentSession.JSON()` at 0/1/2 rows (always array) and `Single: true` (bare object).
- Normalize `HostedAgentWorkspaceUpload.JSON()` (`doctl agents upload`) to the same `Single` bool pattern. It still had the bare `len(h.Uploads) == 1` unwrap with no row-count guard — harmless today since upload is single-file-only (`Uploads` always has exactly one element), but it's the identical latent trap that caused MARSOHS-869/887, so fixing it proactively before a future batch-upload feature could resurrect the bug.

## Test plan

- [x] `go build ./commands/...` — clean
- [x] `gofmt -l` on all touched files — clean
- [x] `go test ./commands/... -run Agent -v` — all pass, including the 6 new displayer shape tests
- [x] Built the real binary (`go build -mod=vendor -o doctl-dev ./cmd/doctl`) and drove it against a local mock API server:
  - `agents list -o json` at 0/1/2 sessions → array of length 0/1/2 in all three cases
  - `agents show <id> -o json` → still a bare object (Single=true unaffected)

Full output posted on the Jira ticket.

[MARSOHS-887]: https://do-internal.atlassian.net/browse/MARSOHS-887?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MARSOHS-869]: https://do-internal.atlassian.net/browse/MARSOHS-869?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commands/agents.go
+++ b/commands/agents.go
@@ -990,10 +990,13 @@ func workspaceTransferUpload(c *CmdConfig, sessionID, workspacePath string, f *o
 	if written == 0 {
 		written = size
 	}
-	return c.Display(&displayers.HostedAgentWorkspaceUpload{Uploads: []*godo.HostedAgentWorkspaceUploadResponse{{
-		Path:         workspacePath,
-		BytesWritten: written,
-	}}})
+	return c.Display(&displayers.HostedAgentWorkspaceUpload{
+		Uploads: []*godo.HostedAgentWorkspaceUploadResponse{{
+			Path:         workspacePath,
+			BytesWritten: written,
+		}},
+		Single: true,
+	})
 }
 
 // workspacePartUploadURLs mints presigned PUT URLs for one or more 1-based parts.
--- a/commands/displayers/hosted_agents.go
+++ b/commands/displayers/hosted_agents.go
@@ -82,14 +82,20 @@ func (h *HostedAgentSession) KV() []map[string]any {
 }
 
 // HostedAgentWorkspaceUpload renders the result of `doctl agents upload`.
+// Upload is single-file-only today (Uploads always has exactly one element),
+// so JSON() defaults to list semantics (always an array) like every other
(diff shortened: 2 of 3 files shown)


SSharma-10 (approved):
LGTM

PR #1911, opened 2026-08-13
agents: add `auth` command for provider OAuth connect flow

jkosanam:
## Summary

Adds `doctl agents auth <provider>` to drive the team-scoped harness-api provider connect flow (Barbican connectlinks), so hosted agent sessions can perform authenticated git operations (e.g. `git clone`/`push` against private GitHub repos).

- Starts (or resumes) the connect flow, prints/opens the authorization URL, then polls until the team's authorization handle is persisted server-side.
- Team-scoped credential; the OAuth handle never leaves the control plane — doctl only observes connect status. Tokens are exchanged server-side at session time.
- Flags: `--no-browser` (print the URL instead of opening a browser), `--no-wait` (print the URL and exit without polling).

Backed by two new `do`-layer calls issued against the shared godo client, matching the deployed contract:

```
POST /v2/agents/auth/{provider}
GET  /v2/agents/auth/{provider}/poll?poll_url=...
```
Response `status` is `pending` or `success`.

## Notes for reviewers

- The wire types (`HostedAgentProviderAuth*`) and the `NewRequest`/`Do` calls live in the `do/` package rather than godo, because these endpoints are not in the vendored godo yet and this avoids editing vendored code. If we prefer the usual convention (wire types in godo, `do/` just wraps), we can land those in godo first and re-point — happy to do that as a follow-up.
- `RunAgentsAuth` confirmations (`Notice: ... connected successfully`, `Re-run ...`) go to **stderr** via doctl's shared `notice()`, consistent with the rest of the agents commands.

## Test plan

- [x] `go build ./do/... ./commands/...`, `go vet`, `gofmt` clean
- [x] Unit tests: `TestRunAgentsAuth` (already-connected, `--no-wait`, poll-until-authorized, missing-provider) and `TestAgentsCommand`
- [x] Manual end-to-end against a local stub of the two endpoints: start → poll(pending → success), `--no-wait`, and connection-refused when the backend is down
- [ ] Live staging run (`doctl agents auth github` against staging gateway)


SSharma-10 (approved):
LGTM

PR #1912, opened 2026-08-13
MARSOHS-883 verify and polish spec.skills doctl authoring flow

amandhaw:
Add test coverage proving multi-line `spec.skills ` instructions survive the --name/legacy-openai manifest round-trips, and that harness-api's `HARNESS_SKILLS` size-cap rejection surfaces as a readable CLI message rather than a raw JSON/HTTP dump. Add a `spec.skills` example to the agents start --help text.


SSharma-10 (approved):
LGTM

PR #1913, opened 2026-08-18
agents: support the flat agents.yaml manifest format

adilhafeez:
## Summary

MARSOHS manifest-simplification, doctl milestone (M6). The platform now accepts a flat `agents.yaml` format (top-level `agent:` key, snake_case, no `apiVersion`/`kind`/`metadata`/`spec` envelope) alongside the deprecated `v1alpha1` envelope. This PR makes doctl's client-side manifest handling dual-format:

- `--name` injection is format-aware: flat manifests get a top-level `name`, legacy envelopes keep `metadata.name` (previously a flat manifest would have been corrupted with an injected `metadata` block the server rejects).
- codex-agentapi (OpenAI sandbox provider) orchestration reads the create-session body from the flat top-level `config` block, falling back to legacy `spec.runtime.config` / `spec.openai`; adapter detection reads flat `agent` or legacy `spec.runtime.adapter`.
- `doctl agents start` prints a deprecation notice when the manifest uses the legacy envelope (detection mirrors the server's routing rule: presence of a top-level `apiVersion` key).
- `agents start` help text and examples are rewritten flat-primary (minimal manifest `agent: opencode`; skills example uses flat `skills` with `allowed_tools`), with a short legacy-deprecation note.

`${VAR}` expansion is unchanged. The server still owns full manifest parsing and validation.

## Merge gate

Per the workstream landing order, this must not ship in a doctl release before the harness-api flat parser is **deployed** to production (flat manifests emitted by the new help/examples must be accepted). Targeting `feat/agents-subcommands` keeps it behind that gate.

## Testing

- New unit tests: format detection, flat/legacy `--name` injection, flat codex-agentapi create-body extraction and session prep (including the wrong-agent error), full `agents start` flow with a flat manifest.
- `go vet ./commands/...` and `go test ./commands/` pass (pre-existing local keychain failures in `TestRegistryLogin`/`TestRegistryLogout` unrelated, fail on the base branch too).

### DIFF
--- a/commands/agents.go
+++ b/commands/agents.go
@@ -240,26 +240,31 @@ Commands that act on a single session accept either the session ID or its name. 
 		"Start a new agent session",
 		`Creates a new agent session from an agent manifest file and prints its session id and status.
 
-The `+"`"+`--spec`+"`"+` flag is required and accepts a YAML manifest matching the `+"`"+`agents.digitalocean.com/v1alpha1`+"`"+` schema. `+"`"+`${VAR}`+"`"+` references in the manifest are resolved from your local environment before upload; referencing an unset variable is an error, and `+"`"+`$${VAR}`+"`"+` escapes to a literal `+"`"+`${VAR}`+"`"+`. The expanded manifest is then sent to the server, which owns parsing and validation.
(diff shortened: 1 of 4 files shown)


SSharma-10 (approved):
LGTM

PR #1915, opened 2026-08-19
agents: add port-forward command (WebSocket tunnel POC)

adilhafeez:
## Summary

POC of `doctl agents port-forward` (MARSOHS port-forward workstream):

- `doctl agents port-forward <session> [<local-port>:]<remote-port> ...` opens one local TCP listener per pair; each accepted connection dials its own WebSocket to `GET /v2/agents/sessions/{id}/port-forward/{port}` and pumps binary frames as an opaque byte stream (no multiplexing in v1 — one WS per TCP connection).
- `--address` controls the local bind (default `127.0.0.1`; a warning is printed for non-loopback binds). `0:<remote>` lets the OS pick the local port.
- Client-side validation: remote ports 1024–65535; listen failures tear down cleanly.
- Server WS close codes distinguish clean guest close (1000) from session-end / dial-failure (4xxx); a per-connection failure never kills the listener.
- Honors `--api-url` / `DIGITALOCEAN_API_URL` (http(s) → ws(s)), auth via the standard bearer token.

The server-side endpoint and the edge WebSocket route have both landed and are deployed; the tunnel was exercised against production while debugging the two items below.

Two things gorilla's self-dialed connection quietly excluded from the rest of the agents surface, both fixed here:

- **Refusals were undiagnosable.** A rejected upgrade printed only `websocket: bad handshake`, discarding the body. The status alone can't separate the API declining the guest dial from a proxy in front of it failing to reach the API, so diagnosing one meant reproducing it by hand with curl. The body is now read once and reused for both the error and the trace, preferring the API's `{"id","message"}` shape and falling back to raw text collapsed to one line and capped. `--trace` missed this path for the same reason — the recorder wraps godo's `RoundTripper` — so the handshake is now logged with the recorder's framing and the bearer token masked.
- **`X-Device-UUID` was not sent** (thanks @SSharma-10). The server mounts its device middleware on the whole `/v2/agents` subrouter and its non-storing endpoints rely on that middleware's log line for attribution, so the tunnel was the one agents call logged with an empty device id — and it is also the one that opens a raw TCP channel into the sandbox. `--trace` masks it the way the server does (prefix + suffix), so a trace line and a server log line for the same request stay correlatable without either recording the full hardware id.

Also fixes `TestAgentsCommand`, which has been failing on this branch since the command was registered without adding `port-forward` to its expected subcommand list.

## Test plan

- [x] End-to-end against a local harness-api dev stack: forwarded local port → HTTP server inside a codex session sandbox, bidirectional traffic verified with curl
- [x] End-to-end against production, which is what surfaced the two diagnostics gaps above
- [x] Out-of-range remote port rejected client-side; dead remote port surfaces per-connection error without killing the listener
- [x] Unit tests for pair parsing, URL construction, rejection-message parsing, header redaction, and `tunnelHeader`
- [x] Two tests drive a real gorilla dial against an `httptest` server rather than the parsing in isolation: one proves the server's message survives the handshake instead of being replaced by `bad handshake`, the other proves `X-Device-UUID` actually arrives

**Note on the empty check list:** `ci.yml` and `typos.yml` both filter `pull_request` to base branches `main` and `feature/**`. This PR targets `feat/agents-subcommands`, which matches neither, so no Go CI runs here — that applies to every PR into this feature branch, not just this one. `go test ./commands/` was run locally instead: the only failure is `TestRegistryLogout`, which fails identically on a clean `feat/agents-subcommands` worktree (it wants a native keychain).

### DIFF
--- a/commands/agent_port_forward.go
+++ b/commands/agent_port_forward.go
@@ -18,6 +18,7 @@ import (
 	"syscall"
 
 	"github.com/digitalocean/doctl"
+	"github.com/digitalocean/doctl/internal/deviceid"
 	"github.com/gorilla/websocket"
 	"github.com/spf13/viper"
 )
@@ -192,7 +193,7 @@ func (t *wsTracer) handshake(wsURL string, header http.Header, resp *http.Respon
 	t.logger.Println("<-", strconv.Quote(line))
 }
 
-// formatHeader renders headers on one line with the bearer token masked;
+// formatHeader renders headers on one line with sensitive values masked;
 // trace output gets pasted into tickets.
 func formatHeader(h http.Header) string {
 	keys := make([]string, 0, len(h))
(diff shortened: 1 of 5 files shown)


syedhashmi (approved):

amandhaw (approved):

commands/agent_port_forward.go
```diff
@@ -0,0 +1,363 @@
...
+		return err
+	}
+	if address != "127.0.0.1" && address != "localhost" && address != "::1" {
+		warn("binding %s exposes the tunnel to anyone who can reach this address", address)
+	}
+
+	header := http.Header{}
+	if token := c.getContextAccessToken(); token != "" {
+		header.Set("Authorization", "Bearer "+token)
+	}
```
SSharma-10: One question here- websocket connection sets `Authorization`, but not `X-Device-UUID,` while the other /v2/agents requests get it through the deviceid transport.
 Is that intentional for port-forwarding, or should we also add deviceid.Get() to the upgrade headers?

SSharma-10 (approved):

commands/agent_port_forward.go
adilhafeez: Not intentional — good catch, and it turns out to matter more than consistency.

harness-api mounts `device.Middleware` on the `/v2/agents` subrouter that the port-forward route hangs off, and the middleware's comment says non-storing endpoints rely on its log line for device attribution. Port-forward is non-storing, so every tunnel was being logged with an empty device id — on the one call that opens a raw TCP channel into the sandbox.

The cause is the same one that kept `--trace` off this path: the deviceid transport wraps godo's `RoundTripper`, and gorilla dials its own connection.

Fixed in b335f56b — `deviceid.Get()` now goes onto the upgrade headers. Two related bits:

- `--trace` redacts it the way the server's `device.Redact` does (`564D***5678`), so a trace line and the server's log line for the same request stay correlatable without either recording the full hardware id.
- Header construction moved into `tunnelHeader(token, deviceID)` with both values injected, because `deviceid.Get` caches in a `sync.Once` and a test can't otherwise control it. Covered by `TestTunnelHeader` plus a real gorilla dial in `TestBridgeLocalConnSendsDeviceHeader` that asserts the header actually arrives at the server.

PR #1919, opened 2026-08-21
feat(agents): stream reasoning live and keep run status visible in attach

Spherrrical:
## Summary

Follow-up UX pass on `doctl agent`/`doctl ohr` on top of `feat/agents-subcommands`, focused on making `attach` sessions legible in real time and tidying up display/validation along the way.

### Live run status & reasoning streaming (the core of this PR)

- Stream reasoning tokens live (dim italic, separate from the buffered final answer) instead of dropping them — for both the SPI protocol (`run.token_delta` with `is_reasoning=true`) and Codex's dedicated reasoning delta/done events.
- Keep a sticky "Run in progress" indicator visible for the *whole* run — tool calls, lifecycle events, not just the gap before the first token — so a multi-second tool call no longer looks identical to a hung session.
- Give accurate "message queued" feedback (distinct from the generic "Waiting for the agent...") when input is sent while a run is still in flight.
- Resolve HITL approvals in place ("Approval required" → "✓ Approved" / "✗ Rejected" / "⏸ Deferred" on the same line) instead of printing a disconnected receipt.

### Attach REPL input handling

- Fix a paste bug where pasting sufficiently long text corrupted the terminal with dozens of stale, truncated duplicate lines (the prompt was repainting on every pasted byte instead of once at the end).
- Add Option/Alt+Enter to insert a newline, so a message can be composed across multiple lines and sent as one input.
- Add `/pause`, `/resume`, `/upload`, `/download`, and `/exit` to the attach REPL, plus Tab-completion for slash commands. Make `/upload` and `/download`'s second argument optional, defaulting the local/workspace path to the other side's basename.
- Redesign the attach banner and detach notice to be left-aligned inside the green box, and rewrite `/help` as plain left-aligned text (no border) using `tabwriter` for column alignment.
- Animate session-creation waits with a live spinner + ticking elapsed time on a real terminal (replacing a static "Creating hosted session…" line that gave no feedback for the duration of the blocking API call). Piped/non-TTY output keeps the old one-line-per-step behavior.

### Display & validation cleanup

- Normalize hosted-agent kind/harness display names (`opencode` -> `OpenCode`, `claude-code` -> `Claude Code`, `codex` -> `Codex`, `cursor cli` -> `Cursor CLI`) so list/show/run output reads consistently.
- Hide session ID and repo by default; show both under `-v`/`--verbose`. Show `CreatedAt` as relative time ("5m ago") by default, full timestamp when verbose.
- Redesign agent error cards: drop the "Reason"/bullet labels and print the reason and next-step command directly, one line each.
- Validate that `ANTHROPIC_API_KEY` is set before creating a `claude-code` session (mirrors the existing `OPENAI_API_KEY`/codex check), so a missing key fails fast locally instead of erroring deep into hosted provisioning.

## Test plan

- [x] `go build ./commands/...`
- [x] `gofmt -l` / `go vet ./commands/...` clean
- [x] `go test ./commands/... ./do/...` (agents-related tests all pass; only pre-existing, unrelated `TestRegistryLogin`/`TestRegistryLogout` failures from macOS Keychain access in the sandbox)
- [x] Manually verified in a real terminal: live spinner animates and clears correctly during session creation
- [x] Manually verified: running `--harness claude-code` without `ANTHROPIC_API_KEY` fails fast with a clear local error, before any network call
- [x] Manually verified in `doctl agent attach`: reasoning streams live and distinctly from the final answer, the "Run in progress" indicator stays up through tool calls without corrupting the terminal, HITL approvals update in place, large pastes no longer duplicate lines, and Option/Alt+Enter composes multi-line messages

### DIFF
--- a/commands/agents.go
+++ b/commands/agents.go
@@ -36,6 +36,7 @@ import (
 	"strings"
 	"sync"
 	"syscall"
+	"text/tabwriter"
 	"time"
 
 	"github.com/charmbracelet/glamour"
@@ -471,6 +472,7 @@ func RunAgentsStart(c *CmdConfig) error {
 	prog := (*creationProgress)(nil)
 	if Output != "json" {
 		prog = newCreationProgress(c.Out)
+		defer prog.stop()
 		prog.header("Launching agent session")
 	}
 
@@ -579,6 +581,7 @@ func finishAgentsStartSession(c *CmdConfig, sess *do.HostedAgentSession, prog *c
 
 	if prog == nil {
 		prog = newCreationProgress(c.Out)
+		defer prog.stop()
 	}
 	sess, err := waitForSessionReady(ctx, c.HostedAgents(), sessionID, prog)
 	if err != nil {
@@ -1158,6 +1161,13 @@ func RunAgentsUpload(c *CmdConfig) error {
 		return err
 	}
 
+	return runWorkspaceUpload(c, sessionID, localFile, workspacePath, isArchive)
+}
+
+// runWorkspaceUpload validates and streams a local file (or tar archive) into
+// a session's workspace sandbox. Shared by `doctl agents upload` and the
+// interactive attach `/upload` command.
+func runWorkspaceUpload(c *CmdConfig, sessionID, localFile, workspacePath string, isArchive bool) error {
 	info, err := os.Stat(localFile)
 	if err != nil {
 		if os.IsNotExist(err) {
@@ -1445,7 +1455,14 @@ func RunAgentsDownload(c *CmdConfig) error {
 		return err
 	}
 
-	written, err := workspaceTransferDownload(c.HostedAgents(), sessionID, workspacePath, saveTo, asArchive)
+	return runWorkspaceDownload(c, c.HostedAgents(), sessionID, workspacePath, saveTo, asArchive)
+}
+
+// runWorkspaceDownload fetches a file (or tar archive) from a session
+// workspace sandbox. Shared by `doctl agents download` and the interactive
+// attach `/download` command.
+func runWorkspaceDownload(c *CmdConfig, svc do.HostedAgentsService, sessionID, workspacePath, saveTo string, asArchive bool) error {
+	written, err := workspaceTransferDownload(svc, sessionID, workspacePath, saveTo, asArchive)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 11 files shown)


commands/agents_run.go
```diff
@@ -310,12 +343,23 @@ func buildHarnessManifest(harness, repo, prompt, name string) ([]byte, error) {
...
 		doc.Config = defaultCodexRunConfig(prompt)
+	case agent == claudeCodeAgentName:
+		// Claude Code needs its own inference key just like Codex needs
+		// OPENAI_API_KEY. Referencing it here (rather than baking in a
+		// literal) routes through expandManifestEnvCollect/ensureEnvVar so a
+		// missing key is caught — and prompted for on a TTY — before doctl
+		// ever calls CreateSessionFromManifest, instead of failing deep into
+		// a hosted session that was never going to work.
+		doc.Env = map[string]string{
+			"ANTHROPIC_API_KEY": "${" + anthropicAPIKeyEnv + "}",
```
SSharma-10, line 363: this only fails fast when the var is unset/empty, a bogus value still creates the session

commands/agents.go
```diff
@@ -3792,6 +4188,20 @@ func handleAttachEscapeSequence(b byte, state *attachState) bool {
 			state.display.redraw()
 		}
 		return true
+	case bytes.Equal(seq, []byte{0x1b, 0x0d}), bytes.Equal(seq, []byte{0x1b, 0x0a}):
+		// Option/Alt+Enter: most terminals (iTerm2 and Terminal.app's default
```
SSharma-10, line 4198: Option/Alt+Enter stores a real newline, but the prompt flattens \n to a space [which ids expected] but it is easy to think the shortcut did nothing until submit

commands/agents_run.go
Spherrrical: Fixed in cf2e8b6a: added `prepareClaudeCodeStart`, which mirrors the existing OPENAI_API_KEY/codex path — after resolving the key it calls `GET /v1/models` on Anthropic's API before doctl ever creates the session, so a bogus (but non-empty) key now fails fast locally with the actual 401/403 from Anthropic instead of only surfacing once already deep inside the hosted sandbox. See `commands/agents_anthropic.go`.

commands/agents.go
Spherrrical: Fixed in cf2e8b6a: `displayInputBuffer` now renders the embedded newline as a visible " ↵ " marker instead of collapsing it to a plain space, so it's immediately obvious the line break landed. The underlying `lineBuf` still carries a real `\n` restored on submit — just the flattened single-row display changed.

SSharma-10 (approved):

PR #1929, opened 2026-08-26
fix(agents): keep warm-up input visible and queue every message

adilhafeez:
Follow-up to MARSOHS-972. Attaching to a session that is still booting mishandled input in three ways.

## What was wrong

**The caret never moved while typing.** `warmupPaintLocked` computed the correct caret column and then discarded it by ending with `\x1b8` (DECRC), restoring the cursor saved at the top of that same paint. Warm-up deliberately suppresses echo and repaints instead, so that paint was the only thing that could advance the caret — it stayed pinned just after `> ` no matter how much you typed.

**Submitted messages vanished.** On Enter the buffer is cleared first, then `echoAttachSubmitNewline` skipped the commit to scrollback whenever the banner was visible. The typed text was gone with no record of what had been queued.

**Only the first message was ever sent.** A second Enter hit the `inputAlreadyQueued` guard and was dropped outright — not sent, not shown, not returned to the input line:

```
> hello
Message already queued — waiting for agent to start   <- "hello again" silently lost
```

The guard was also unnecessary. It is only set after a *successful* send, which means the guest run was already accepting input; from that point OHR queues turns in order on a 64-slot channel and answers a full queue with a clean "agent is busy" error.

**Stale spinner rows.** `Write` ignored `warmupStatusLines`, so any plain output during warm-up landed on the prompt row and pushed the prompt down without the banner's row accounting knowing. Every subsequent frame then painted its spinner one row lower, stranding the previous one:

```
⠋ Agent is warming up… please wait
⠙ Agent is warming up… please wait
```

## What changed

- The banner renders top-down from a known anchor via `appendWarmupBlock` and stops at the computed caret — no DECSC/DECRC to override it. Row count is derived from `warmupBlockLinesLocked()`, so accounting and painting cannot disagree.
- Scrollback writes and submitted lines go above the pinned banner (`writeAboveWarmupLocked`); clearing the banner deletes its rows (DL) rather than blanking them in place, which used to leave empty lines behind.
- Every message typed during warm-up is sent and committed to scrollback in order. `markInputQueued` counts instead of latching, and the banner reports the running total.
- A failed send calls `restoreInput`, putting the text back on the input line with the caret at its end, on both the hosted and OpenAI attach paths.

Result for the reported case:

```
> hello
> hello again
⠋ Agent is warming up… please wait
2 messages queued until agent is ready
>
```

No client-side cap was added: the backend stays the single authority on capacity, so the count stays honest.

## Test plan

- [x] `go test ./commands/` — no new failures (`TestRegistryLogin`/`TestRegistryLogout` fail identically on a clean checkout of the base branch; unrelated and environment-dependent)
- [x] New `commands/agents_warmup_render_test.go` replays the real escape-sequence stream through a small terminal model and asserts the resulting screen — the only way to express "exactly one spinner row" or "caret in column 7". Covers caret tracking (end and mid-line), one spinner row across events, committed lines above the banner, the two-message case above, and no blank rows after teardown.
- [x] Each new test was confirmed non-vacuous by reverting the corresponding fix and watching it fail in the reported way.
- [x] `TestWarmupState_blocksDuplicateSubmit` replaced by `TestWarmupState_queuesEverySubmittedMessage` and `TestWarmupState_failedSendRestoresInput`.
- [x] `gofmt` / `go vet` clean.
- [x] Caret tracking and the visible queued message were confirmed in a live `doctl agents attach` session against a warming-up sandbox.
- [ ] Multi-message queueing is covered by tests but has not yet been exercised against a live session — worth a reviewer sanity check.

### DIFF
--- a/commands/agents.go
+++ b/commands/agents.go
@@ -2445,10 +2445,11 @@ func printAttachSendAck(out io.Writer, warmup *warmupState, thinking *thinkingSt
 	fmt.Fprintln(out, colorize("… waiting for the agent", colMuted))
 }
 
-func echoAttachSubmitNewline(display *promptDisplay, warmup *warmupState, visual string) {
-	if warmup != nil && warmup.isBannerVisible() {
-		return
-	}
+// echoAttachSubmitNewline commits the submitted line to scrollback. During
+// warm-up it lands above the banner rather than being dropped: the buffer is
+// already cleared by then, so skipping it erased every record of what the user
+// queued while they waited.
+func echoAttachSubmitNewline(display *promptDisplay, visual string) {
 	display.finishInputLine(visual)
 }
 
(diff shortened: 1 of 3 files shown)


SSharma-10 (approved):
LGTM

PR #1930, opened 2026-08-26
feat(agents): accept the manifest as an argument, default to ./agents.yaml

adilhafeez:
## Summary

Naming a manifest required a flag for the most common invocation. The thing everyone types first — `doctl agents start agents.yaml` — was silently ignored and fell through to `one of --harness, --spec, or --config-id is required`.

`start`, `run` and `validate` now resolve the manifest most-explicit-first: `--spec` beats a positional path, which beats an `agents.yaml` discovered in the working directory. All four of these are now equivalent:

```bash
doctl agents start                          # uses ./agents.yaml
doctl agents start agents.yaml
doctl agents start -f agents.yaml
doctl agents start --spec agents.yaml       # unchanged
```

`--spec` stays canonical in help output because it is doctl's existing word for a declarative file (`apps create/update/propose`, `dedicated-inference create/update`). `-f` and `--file` are added as aliases for the kubectl / compose / planoai muscle memory — `--file` is a pflag normalize alias rather than a second flag, so there is no duplicate to keep in sync.

## Behavior notes

- **Implicit sources are never silent.** A discovered file is announced (`Notice: using agents.yaml`), suppressed under `-o json` so machine output stays parseable.
- **Discovery only runs when no other source was selected**, so `--harness opencode` in a directory holding an `agents.yaml` stays unambiguous.
- **Pairing a positional path with `--harness` / `--config-id` is an error**, not a silent winner; so is passing both `--spec` and a positional.
- `validate` loses `requiredOpt()` on `--spec`, since requiring the flag would defeat the point.
- `-` (stdin) and `${VAR}` expansion are unchanged.

## Test plan

- [x] `go build`, `go vet`, `gofmt` clean
- [x] `go test ./commands/...` — passes except pre-existing `TestRegistryLogin` / `TestRegistryLogout`, which fail identically on an unmodified tree (local docker credential helper / keychain)
- [x] New unit tests: `TestNamedManifestPath` (flag, positional, both-given, >1 positional, neither), `TestDiscoverManifestFile` (`.yaml`, `.yml`, ignores a *directory* of that name, absent), `TestRunAgentsStart_DiscoversAgentsYAML`, `TestRunAgentsStart_RejectsManifestArgWithHarness`
- [x] Updated the two tests asserting the old required-flag message; both now `t.Chdir` to a temp dir so a stray `agents.yaml` cannot make them flaky
- [x] Manually exercised all five spellings plus `--file=x` and stdin against a built binary, and confirmed `start` still reaches session creation via the `--file` alias

### DIFF
--- a/commands/agents.go
+++ b/commands/agents.go
@@ -57,6 +57,7 @@ import (
 	"github.com/muesli/termenv"
 	"github.com/pkg/browser"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 	"golang.org/x/term"
 	yaml "gopkg.in/yaml.v2"
 )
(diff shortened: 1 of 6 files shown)


commands/agents.go
```diff
@@ -911,6 +925,81 @@ func readManifest(stdin io.Reader, path string) ([]byte, error) {
...
+// conventional name used throughout the docs.
+var manifestFileNames = []string{"agents.yaml", "agents.yml"}
+
+// namedManifestPath returns the manifest the user named explicitly, either
+// with --spec (also spelled -f / --file) or as a positional path. An empty
+// result means they named none, which leaves the caller free to fall back to
+// discoverManifestFile — but only once it knows no other source (--harness,
+// --config-id) was selected.
+func namedManifestPath(c *CmdConfig) (string, error) {
+	spec, err := c.Doit.GetString(c.NS, doctl.ArgAgentSpec)
```
SSharma-10: This can fail if --spec is still marked required in the user’s config. Then doctl agents start won’t pick up agents.yaml.

SSharma-10: Updated

PR #1931, opened 2026-08-26
feat(agents): land in the chat after start, with --detach to opt out

adilhafeez:
## Summary

Starting an agent and then running a second command to talk to it inverts the common case — the reason to start a session is to use it. `doctl agents start` now opens the chat once the session is ready.

```bash
doctl agents start --spec agents.yaml            # ready, then you are in the chat
doctl agents start --spec agents.yaml --detach   # ready summary only (previous behavior)
```

This also closes the `start` / `run` gap: the two verbs were previously distinguished only by whether they attached, and `run --no-attach` was already identical to `start`. They now share one decision helper and one Codex-proxy tip instead of two copies, and `run`'s help says it is equivalent.

## Detach rules

Detach is explicit via `--detach` / `-d`, and **implied** by two conditions:

| Condition | Why |
|---|---|
| `-o json` | Machine-readable output must not become a TUI |
| stdin/stdout is not a terminal | Safety rail — see below |

The non-terminal rule is the important one. `runAgentsAttachSession` opens an SSE stream and reads stdin, so without it every existing script or CI job calling `doctl agents start --spec x.yaml` would have gone from "print a summary and exit" to "block forever in a TUI nothing can drive". The best evidence it works: all eleven pre-existing `TestRunAgentsStart` tests pass **unmodified**, because `go test` runs with stdout as a pipe and therefore takes the detached path.

## Compatibility

- `--no-attach` keeps working everywhere it did, now as a hidden alias of `--detach`. No `run` caller changes.
- `run` behavior is unchanged (it already attached by default).
- The ready summary, `--prompt` send, and wait/timeout behavior are untouched on the detached path.

## Test plan

- [x] `go build`, `go vet`, `gofmt` clean
- [x] `go test ./commands/...` — passes except pre-existing `TestRegistryLogin` / `TestRegistryLogout`, which fail identically on an unmodified tree (local docker credential helper / keychain)
- [x] New `TestAttachAfterStart`: attaches by default on a terminal; `--detach`, `--no-attach`, `-o json`, and a non-terminal each opt out
- [x] All pre-existing `TestRunAgentsStart` / `TestRunAgentsRun` tests pass unmodified
- [x] Verified `-d, --detach` renders on both `start --help` and `run --help`, and that the hidden `--no-attach` is still accepted

### DIFF
--- a/args.go
+++ b/args.go
@@ -1059,8 +1059,12 @@ const (
 	// ArgAgentRepo is a Git repository URL cloned into the session workspace.
 	ArgAgentRepo = "gh-repo"
 
-	// ArgAgentNoAttach stops `agents run` after the session is ready instead of attaching.
+	// ArgAgentNoAttach is the older spelling of ArgAgentDetach, still accepted.
 	ArgAgentNoAttach = "no-attach"
+
+	// ArgAgentDetach stops `agents start` / `agents run` at the ready summary
+	// instead of opening the chat TUI.
+	ArgAgentDetach = "detach"
 
 	// ArgAgentWaitTimeout is how long `agents run` waits for readiness, in seconds.
 	ArgAgentWaitTimeout = "wait-timeout"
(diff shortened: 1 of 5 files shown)


SSharma-10 (approved):
LGTM

PR #1936, opened 2026-08-31
Fix databases resize --wait returning before resize completes

gangasingh01:
1. Fix `doctl databases resize --wait` returning success immediately while the cluster is still online with its old config. this happens because it make take few seconds for cluster status to change to RESIZING

2. Wait until the cluster is online and matches the requested --size, --num-nodes, and --storage-size-mib (when provided).

3. Skip the storage check for redis and valkey, which do not support storage sizing.

Tested locally using 

` ./build/doctl databases resize  <CLUSTER_UUID> --num-nodes 3 --size gd-4vcpu-16gb --storage-size-mib 102400 --wait --api-url=https://api.digitalocean.com  --access-token <ACCESS_TOKEN>          ` 
 
` --wait `flag is working as expected with these changes
 
 
image


anup-deka (approved):
LGTM

PR #1938, opened 2026-08-31
agents: stop the codex facade rendering reasoning as the reply (MARSOHS-1012)

nkurraDO:
Fixes [MARSOHS-1012](https://do-internal.atlassian.net/browse/MARSOHS-1012) (p2, "Cursor showing thinking process as response"). Targets `feat/agents-subcommands` because the agents surface is not on `main` yet.

## The bug

`translateEvent`'s `TokenChunk` case decoded only `text`:

```go
var payload struct {
    Text string `json:"text"`
}
```

so a chunk marked `is_reasoning` was forwarded as `item/agentMessage/delta` — the exact same channel as the answer. The client concatenates that channel into the visible reply, so the model's thinking trace rendered *as* the response.

This is the path that matters for the reported case. `raw.go` only forwards native frames when the hosted runtime is codex itself, so a **cursor** session always falls back to this canonical translator.

## Confirmed on production

A real cursor session on prod returned 25 token deltas, **all 25 carrying the field: 4 `true`, 21 `false`**. The raw SSE shows the provenance:

```
"source_event_type":"session/update:agent_thought_chunk",  "data":{"text":"This is a classic riddle.","is_reasoning":true}
"source_event_type":"session/update:agent_message_chunk",  "data":{"text":"“","is_reasoning":false}
```

The server side has been correct since [cthulhu#172482](https://github.internal.digitalocean.com/digitalocean/cthulhu/pull/172482) ("OHS: forward is_reasoning on token chunks to consumers"), whose commit `299c38e6` is what production runs today. Before it, both proto→SPI mappings dropped the flag, which is why it previously looked absent. The remaining defect was entirely client-side.

## Why drop rather than re-route

`item/agentMessage/delta` is the only text channel this facade speaks and the protocol has no reasoning item, so there is nowhere faithful to put a thought. Codex sessions don't lose anything — `raw.go` forwards their native frames, which carry reasoning as its own item. Dropping is the only honest option for the non-codex fallback until the protocol gains somewhere to put it.

The skip sits **ahead of** the `item/started` block on purpose: a turn that leads with reasoning would otherwise announce an `agentMessage` item whose only content we then refuse to send.

## Deliberately not touching `commands/agents.go`

This branch already handles reasoning there, and handles it *better* — `reasoningStreamer` renders the trace in a separate muted channel rather than discarding it. That is the right treatment where a channel exists, and it is left alone. Only the facade, which has no such channel, changes here.

## Test plan

- New `TestFacade_TurnStart_SkipsReasoning` pins that a reasoning delta produces no notification and stays out of the completed item. The reasoning chunk leads the turn, so it also pins that no `item/started` is announced on its behalf.
- **Verified the test guards the fix**: removing the skip makes it fail, with the reasoning text leaking through as a second `item/agentMessage/delta`.
- `go test ./internal/agentproxy/... ./commands/` all pass.

[MARSOHS-1012]: https://do-internal.atlassian.net/browse/MARSOHS-1012?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/agentproxy/codex/facade_test.go
+++ b/internal/agentproxy/codex/facade_test.go
@@ -348,6 +348,45 @@ func TestFacade_TurnStart_SendInputError(t *testing.T) {
 // translated into the turn/started, item/started, item/agentMessage/delta*,
 // item/completed, turn/completed notification sequence the design doc's
 // event-mapping table calls for.
+// TestFacade_TurnStart_SkipsReasoning pins MARSOHS-1012: a token chunk marked
+// is_reasoning is the model's thinking trace, not the answer, and the facade
+// has no reasoning item to route it to — so it must not surface as
+// agentMessage text. The reasoning chunk here leads the turn, which also pins
+// that no item/started is announced on its behalf: the first notification
+// after turn/started belongs to the answer.
+func TestFacade_TurnStart_SkipsReasoning(t *testing.T) {
+	f, h, rec := newTestFacade(t)
+
+	h.QueueRun("run-1",
+		agentproxytest.Event{Type: string(godo.HostedAgentEventKindRunStarted)},
+		agentproxytest.Event{Type: string(godo.HostedAgentEventKindTokenChunk), Data: json.RawMessage(`{"text":"let me think","is_reasoning":true}`)},
+		agentproxytest.Event{Type: string(godo.HostedAgentEventKindTokenChunk), Data: json.RawMessage(`{"text":"Paris","is_reasoning":false}`)},
+		agentproxytest.Event{Type: string(godo.HostedAgentEventKindRunCompleted)},
+	)
+
+	_, err := dispatch(t, f, "turn/start", turnStartParams{
+		ThreadID: testSessionID,
+		Input:    []userInputItem{{Type: "text", Text: "hi"}},
+	})
+	require.NoError(t, err)
+
+	require.Equal(t, "turn/started", rec.next(t).method)
+	require.Equal(t, "item/started", rec.next(t).method)
+
+	delta := rec.next(t)
+	require.Equal(t, "item/agentMessage/delta", delta.method)
+	assert.Equal(t, "Paris", delta.params.(agentMessageDeltaNotification).Delta,
+		"reasoning must not be forwarded as an agentMessage delta")
+
+	itemCompleted := rec.next(t)
+	require.Equal(t, "item/completed", itemCompleted.method)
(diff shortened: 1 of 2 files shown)


nkurraDO:
## Verified end to end against a live production cursor session

Built both binaries and ran the **same turn** through the codex facade twice, changing only the binary. One session, reused for both halves, so the binary is the only variable. The facade was driven over its WebSocket JSON-RPC (`initialize` → `thread/start` → `turn/start`) and every `item/agentMessage/delta` was collected — that channel is what the client renders as the reply.

Prompt: *"Think step by step before answering. A farmer has 17 sheep and all but 9 run away. Reason it through, then state the final number."*

**Before (`feat/agents-subcommands`, unpatched) — 34 deltas:**

```
This is a classic riddle. I am reasoning through the numbers.

Nine sheep remain.“All but 9 run away” means every sheep **except** 9 leaves. Those 9 stay.

**Final number: 9**
```

**After (this branch) — 25 deltas:**

```
“All but 9” means all except 9 leave, so 9 remain.

**Final number: 9**
```

The unpatched output is the bug verbatim. The thinking trace leads the reply, and you can see it collide with the actual answer mid-line — `Nine sheep remain.` is the tail of the reasoning and `“All but 9 run away” means…` is the start of the answer, concatenated with no separator because both arrived on the same channel. That is "Cursor showing thinking process as response."

For completeness, the raw SSE for that session confirms where the flag comes from:

```
"source_event_type":"session/update:agent_thought_chunk",  "data":{"text":"This is a classic riddle.","is_reasoning":true}
"source_event_type":"session/update:agent_message_chunk",  "data":{"text":"“","is_reasoning":false}
```

## One note for reviewers, unrelated to this diff

Creating the test session needed an older binary: `knownAgentAdapters` in `commands/agents_validate.go` no longer lists `cursor`, and `TestValidateAgentManifest_RemovedAdaptersRejected` pins that rejection deliberately — but the **server still accepts** `agent: cursor` and happily returns `AGENT_KIND_CURSOR_CLI`. So cursor sessions exist and stream today; they just can't be created through current doctl.

That is worth knowing here because it's the reason this fix matters rather than being dead code: cursor sessions created via the API or UI still reach the facade, take the canonical path (no codex `source_raw` to forward), and hit exactly the line this PR changes. Whether the client-side accept-list should be re-synced with the server is a separate question and I haven't touched it.

SSharma-10 (approved):

PR #1941, opened 2026-09-01
Pre-run flag validation and terminal error styling

SSharma-10:


commands/doit.go
```diff
@@ -159,10 +159,17 @@ func configHome() string {
...
+			os.Exit(-1)
 		}
-		os.Exit(-1)
+		// Route through checkErr so flag validation and other errors honor
+		// --output json and a consistent Error:/JSON shape, but keep the
+		// historical process exit status (-1 → 255) for top-level failures.
+		prev := errAction
+		errAction = func() { os.Exit(-1) }
+		defer func() { errAction = prev }()
+		checkErr(err)
```
DO-rrao: **Exit code for missing required flags changes from `1` to `255`.**

Validation now fails in `PreRunE`, so the error propagates up to `Execute()` and hits the forced `os.Exit(-1)`. Previously these errors surfaced inside `Run` via `checkErr` -> `errAction` -> `os.Exit(1)`.

```
$ doctl compute droplet create foo
next_gen_phase1_dev: exit 1
this branch:         exit 255
```

The comment says this keeps the historical process exit status, but the historical status for missing-required-flag failures was specifically `1`. `-1` was only ever used for the top-level cases that never reached `checkErr`. Anything branching on the exit code breaks.

Simplest fix is to leave `errAction` alone for `*FlagValidationError` so the default `os.Exit(1)` applies, and keep `-1` for the cases it was meant for.

commands/errors.go
```diff
@@ -71,15 +71,24 @@ func checkErr(err error) {
 
 	switch output {
 	default:
+		if fv, ok := err.(*FlagValidationError); ok {
```
DO-rrao: Minor: this is a bare type assertion, so if anything ever wraps the error you fall through to the generic branch and print `Error: ✘ Error: Missing ...` with a doubled prefix. `errors.As` avoids that, and `errors` is already imported here.

```suggestion
		var fv *FlagValidationError
		if errors.As(err, &fv) {
```

commands/flag_validation.go
```diff
@@ -0,0 +1,433 @@
...
+	}
+
+	if key := flagAnnotation(f, annoFlagViperKey); key != "" {
+		if v := strings.TrimSpace(viper.GetString(key)); !isEmptyFlagRaw(v) {
+			return v
+		}
+	}
+
+	return raw
+}
```
DO-rrao: **The viper fallback only handles scalars, so required slice flags supplied via `config.yaml` are rejected.**

`viper.GetString` returns `""` for a list value, and the pflag value is `"[]"`, so the flag reads as missing:

```yaml
# config.yaml
tag:
  apply:
    resource:
      - do:droplet:12345
```

```
$ doctl compute tag apply mytag --config config.yaml
next_gen_phase1_dev: POST /v2/tags/mytag/resources
this branch:         ✘ Error: Missing required flag for doctl compute tag apply
```

This is the same bug class as the defaults/config issue that commit d2a876a fixed for strings, just not covered for slices. Ten required `stringSlice` flags are affected:

- `compute tag apply --resource`
- `compute tag remove --resource`
- `databases firewalls replace --rule`
- `monitoring uptime update --regions`
- `network attachment create --vpc-ids`
- `nfs create --vpc-ids`
- `secrets unset --key`
- `spaces keys create --grants`
- `spaces keys update --grants`
- `vector-databases tags --tag`

The fallback needs to branch on `f.Value.Type()` and use `viper.GetStringSlice` for `stringSlice`/`stringArray`.

commands/flag_validation.go
```diff
@@ -0,0 +1,433 @@
...
+
+func isEmptyFlagRaw(raw string) bool {
+	return raw == "" || raw == "[]"
+}
```
DO-rrao, line 288: **"Non-empty string" doesn't match doctl's own required-flag semantics for numeric flags.**

`LiveConfig.GetInt` treats a required int of `0` as missing (`doit.go:393`), and `GetFloat64` does the same for `0.0`. Here `f.Value.String()` returns `"0"`, which is non-empty, so the flag passes pre-run and then fails later with the old message:

```
$ doctl compute volume-action resize <id> --region nyc1
Error: (volume-action.resize.size) command is missing required arguments   # unchanged on both builds
```

Net effect is two presentations for the same class of mistake: 157 required string flags get the new block, the 10 required int flags get the old one-liner. Affected flags:

- `compute droplet-action change-kernel --kernel-id`
- `compute droplet-action get --action-id`
- `compute droplet-action restore --image-id`
- `compute image-action get --action-id`
- `compute volume-action get --action-id`
- `compute volume-action resize --size`
- `databases pool create --size`
- `databases resize --num-nodes`
- `kubernetes cluster node-pool create --count`
- `network attachment create --connection-bandwidth-in-mbps`

Emptiness should be type-aware and mirror `LiveConfig`: zero-is-unset for int/float, `""` for string, `[]` for slices.

commands/flag_validation.go
```diff
@@ -0,0 +1,433 @@
...
+func (e *FlagValidationError) Error() string {
+	return e.format(ui.NewStyle(ui.Plain(os.Stdout, os.Stderr)))
+}
+
+// Display renders the validation error using the Next-Gen terminal design
+// system (colored error label, bold flags/commands, dim hints).
+func (e *FlagValidationError) Display() string {
+	env := ui.Detect(os.Stdout, os.Stderr)
+	return e.format(ui.NewStyle(env))
+}
```
DO-rrao, line 63: This re-detects terminal capabilities from scratch, which loses the policy `NewCmdConfig` applies at `command_config.go:111-114`:

```go
UI: ui.Detect(out, os.Stderr,
    ui.WithMachineOutput(Output != "text"),
    ui.WithInteractive(Interactive),
),
```

The point of the capability layer added in 4fb3b9f5 was to resolve this once per invocation and pass it down, so a second, weaker detection on the error path undercuts it. Suggest extracting a single `resolveEnv()` used by both `NewCmdConfig` and `checkErr`.

For what it's worth the colours themselves are correct: I checked with stderr on a pty and stdout redirected, and the ANSI correctly stays on stderr.

internal/ui/style.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+// Next-Gen terminal palette from doctl-nextgen-design.html (Kraken tokens).
+var (
+	ColorError   = lipgloss.Color("#d74623") // primary-terracotta
+	ColorSuccess = lipgloss.Color("#00c483") // primary-green-200
+	ColorWarning = lipgloss.Color("#daa549") // tentacle-500
+	ColorInfo    = lipgloss.Color("#6FE0ED") // foam-200
+	ColorMuted   = lipgloss.Color("#8090a0") // secondary / dim
+)
```
DO-rrao, line 31: This is the part of the earlier "don't redefine the UI parts, reuse the common UI" feedback that's still open. Moving off `termstyle.go` onto `ui.Env` is the right call, but the palette here is new and doesn't match anything else in the tree:

- `commands/charm/colors.go` -> `Colors.Error = #ff6188`, `Colors.Muted = 241`
- `commands/errors.go` -> `colorErr = color.RedString("Error")` (fatih/color red)
- here -> `ColorError = #d74623`

So the same binary prints two different reds for errors depending on which path produced them. `commands/charm/text` already exposes `Error`, `Muted`, `Bold` and a `✘` crossmark that do the same jobs as `ErrorLabel`/`Dim`/`Bold` here.

If the Next-Gen tokens are meant to replace the charm palette, let's do that swap in `charm/colors.go` so there's one source of truth. If they're meant to coexist during the migration, that needs to be written down somewhere, because as it stands it just reads as drift.

Two smaller things in this file while you're in here: both branches of `ErrorLabel` (L52-58) are byte-identical, so the `if s.env.ASCII` does nothing — `env.Glyphs()` already handles the fallback. And `Success`, `Warning` and `Info` (L74-87) have no callers outside `style_test.go`.

integration/image_create_test.go
```diff
@@ -74,28 +74,26 @@ var _ = suite("compute/image/create", func(t *testing.T, when spec.G, it spec.S)
...
-
 		var cases = []struct {
 			desc string
 			err  string
 			args []string
 		}{
 			{
 				"missing all",
-				fmt.Sprintf(baseErr, ".image-name"),
+				"Missing 2 required flags for doctl compute image create",
```
DO-rrao: These assertions got weaker rather than updated, here and in the ~25 similar changes across `integration/gradientai_*`.

`"missing required arguments"` -> `"required flag"` no longer checks *which* flag was reported. In this file specifically, `"missing all"` (L84) and `"missing all flags"` (L89) now assert the same `"Missing 2 required flags ..."` string, where previously they distinguished `.image-name` from `.image-url`.

Reporting the right set of flags is the whole point of this feature, so these should assert on the flag names in the output, e.g. `expect.Contains(output, "--image-url")` alongside the count. Same applies to the gradient suite.

DO-rrao (commented):
A few review notes from building this branch alongside `next_gen_phase1_dev` and diffing behaviour between the two binaries. Details inline.

DO-rrao (approved):

PR #1942, opened 2026-09-01
agents proxy: opencode facade M0-M1 — attach renders against stubs

julia-ye:
First slice of `doctl agents start-proxy --type opencode`: an unmodified `opencode` CLI attaches to the proxy as if it were a local `opencode serve` backend. This PR covers M0 (protocol capture) + M1 (attach renders against stubs); prompt/event bridging to the hosted session lands in the next PR.

## Why opencode is codex-shaped

opencode's TUI is natively a remote-capable HTTP client: `opencode attach <url>` points it at any server speaking the opencode REST + SSE surface — no client modification, same product shape as `codex --remote`. The transport differs: HTTP REST + SSE instead of JSON-RPC over WebSocket.

## M0 — capture

Ran a real `opencode serve` + `opencode attach` pair (both v1.18.25) through a logging relay. Key findings, encoded in the code:

- The TUI subscribes to `/global/event`, whose SSE frames wrap the instance-stream shape in a payload envelope (`{"payload":{id,type,properties}}`) — not the bare `{type,properties}` shape.
- First frame must be `server.connected` (the TUI's liveness signal).
- The attach burst is ~30 parallel read-only GET routes, spanning the classic surface (`/config`, `/provider`, `/agent`, `/session`, ...) and a newer wrapped `/api/*` surface.

## M1 — implementation

- `internal/agentproxy/httpproxy.go`: `ServeHTTP`/`ServeHTTPListener`, the REST+SSE sibling of the WS transport. Same loopback-only rule and shutdown pattern; `BaseContext` roots request contexts in the server ctx so a blocked SSE handler is released on shutdown (the same bug class the WS transport's connCtx comment records).
- `internal/agentproxy/opencode/facade.go`: an `http.Handler` facade answering the captured burst, streaming `server.connected` first, enforcing one event-stream consumer at a time (409), refusing `POST /session/{id}/share` (403 — share syncs history to opencode's public infra), and logging + 404ing unknown routes as the protocol-drift detector. A synthetic single provider/model catalog gives the TUI a selected model, since model choice is a session-create-time decision on the hosted side.
- `commands/agents.go`: `--type opencode` accepted; per-type listen/connect strings; `--replay` rejected for opencode (its attach fetches history itself — that endpoint is a later milestone).

## Testing

- Unit tests: the full attach-burst route inventory answers 200/valid JSON, SSE envelope shape, single-consumer conflict, share refusal, catalog self-consistency, and transport shutdown release.
- Live: against both a stub harness and the local dev stack, a real `opencode attach http://127.0.0.1:1144` renders the full empty-session TUI with **zero unhandled routes** at v1.18.25.

## Next (separate PRs)

M2 `prompt_async` → `SendInput` + event-stream bridging (first demo checkpoint), M3 message history, M4 tool calls + usage, M5 permission round-trip, M6 hardening.

### DIFF
--- a/commands/agents.go
+++ b/commands/agents.go
@@ -53,6 +53,7 @@ import (
 	"github.com/digitalocean/doctl/do"
 	"github.com/digitalocean/doctl/internal/agentproxy"
 	"github.com/digitalocean/doctl/internal/agentproxy/codex"
+	"github.com/digitalocean/doctl/internal/agentproxy/opencode"
 	"github.com/digitalocean/godo"
 	"github.com/muesli/termenv"
 	"github.com/pkg/browser"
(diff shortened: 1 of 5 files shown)


SSharma-10:
M1 verified: TUI attaches cleanly. Prompt currently fails with an unhandled POST /session, which is expected since session creation and prompt bridging are part of M2

SSharma-10 (approved):

PR #1943, opened 2026-09-02
secrets: add kvs flag for secrets get

bentranter:
Adds a new `--kvs` flag for the `doctl secrets get` command which prints the secret values as `KEY=VALUE` pairs.

For example, running:

```sh
$ doctl secrets get --region tor1 --kvs my-secret
```

will output something like:

```
KEY=b-value
OTHER_KEY=a-value
```

These will always print in alphabetical order.

### DIFF
--- a/args.go
+++ b/args.go
@@ -732,6 +732,11 @@ const (
 	ArgSecretVersion = "version"
 	// ArgSecretShow reveals secret values instead of masking them on get.
 	ArgSecretShow = "show"
+	// ArgSecretShowKV reveals secret values instead of masking them on get,
+	// formatted as:
+	//
+	//	KEY=VALUE
+	ArgSecretShowKV = "kvs"
 	// ArgSecretRaw writes a single secret value to stdout with no formatting.
 	ArgSecretRaw = "raw"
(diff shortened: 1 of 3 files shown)


MaxLivesInTheMatrix:
As I was writing my comment about alphabetical order I see you fixed it lol. But yeah this looks good. Idk why I didn't implement this the first time I went through this though...

MaxLivesInTheMatrix (approved):

sebnguyen (approved):

SSharma-10:
`--raw` is checked before `--kvs`, so this fails even though `--kvs` would ignore `--raw`:

`doctl secrets get anything --region nyc3 --kvs --raw`

It returns `--raw requires --key`

We could either short-circuit --kvs earlier or explicitly reject conflicting flags
Also, it’d be good to mention --kvs in the secrets get long help since it currently only covers --show / --raw.

Otherwise LGTM

bentranter:
Updated, thanks for the feedback @SSharma-10

SSharma-10:
> Updated, thanks for the feedback @SSharma-10

LGTM, thank you.

SSharma-10 (approved):

PR #1946, opened 2026-09-03
Standardize wait and progress handling behind one shared waiter

DO-rrao:
## Why

Every `--wait` in doctl polled the API its own way:

- some printed a trail of dots, some printed nothing at all
- several would poll forever if a resource never settled, wedging a script indefinitely
- an action that ended in `errored` was reported as a **successful** wait, so `--wait` claimed a resource was ready when the API had already given up on it
- `doctl compute droplet create --wait` waited inside `do/droplets.go` via `util.WaitForActive`, silently, unbounded, discarding its own error

## What

**One shared waiter.** `commands/wait.go` adds `waiter` + `waitOp`, and every wait path now goes through it: actions (which covers nfs, volumes, images and droplet actions), droplets, databases, vector DBs, app deployments, Kubernetes, load balancers, VPC peerings, security scans and partner network connect.

Each `waitOp` carries an `Activity`, the short present-tense phrase shown while the operation runs (`Creating database (some-id)`), and a `Subject`, the longer phrase used after the fact in the timeout and failure lines, where the message is read out of context and the extra words are what make it actionable. Where one helper serves several commands the verb comes from the caller, since only it knows whether a database is being created, forked or migrated.

Stage detail is appended only where it adds something the activity does not: the raw API status for a database or cluster (`creating`, `provisioning`), progress counts for droplet and deployment waits, where `3 of 5 active` beats five separate statuses. A count of one is not reported, and an action no longer echoes `in-progress`, which is the only status an in-flight action can report.

**Every wait is bounded.** New `--wait-timeout` (default 30m), registered with `--wait` as a pair by `AddWaitFlags` so a command cannot offer one without the other. The timeout error says the operation is unaffected and how to wait longer, because otherwise a timeout reads like a failed create.

**Progress degrades by stream instead of being suppressed.** A terminal gets one updating line. A pipe or a CI log gets a plain line per stage change — no glyphs, no escape sequences, and no repeat of a stage that has not moved (a 10s poll across a 20m provision would otherwise write a hundred identical lines), with an unchanged stage repeated on a one-minute heartbeat so a slow job does not read as a hung one. Glyphs are a terminal affordance; a build log is read by grep and by people who never saw the terminal it would have been drawn on.

Terminal:

```
⠹ Creating database (some-id) (creating) (12s)
✓ Database (some-id) is online (3m20s)
```

Pipe or CI:

```
Creating database (some-id) (0s)
Creating database (some-id) (creating) (12s)
Creating database (some-id) (configuring) (1m40s)
Database (some-id) is online (3m20s)
```

Progress goes to stderr throughout, so piping data into another program keeps working.

## Also in this diff

Two things that are separable from the wait work, called out so reviewers know they are here on purpose. Happy to split either into its own PR if you would rather review them apart:

- **Width-aware text displayer.** `commands/displayers/output.go` drops `text/tabwriter` for a renderer that fits a table to the terminal attached to *stdout*, measuring in terminal cells so styled and double-width values stay aligned. A column is never narrowed past the longest run in its values that has nowhere to break, so IDs, addresses and UUIDs keep their full width while prose and comma-separated lists give up space; half an ID is not a shorter ID, and it cannot be pasted into the next command. When those floors still leave the table wider than the terminal it is written at full width and without rules, since narrowing only the cuttable columns would cost values and still not fit. Piped output stays unconstrained, which is why `ui.Env` gains `DataWidth` separately from `Width`.
- **Diagnostics routed through `ui.Env`.** `warn`/`notice` wrote through `fatih/color` globals that decide once, at package init, from whether *stdout* is a terminal. So `doctl ... 2>log` wrote escape sequences into the log file, and `doctl ... > data` stripped colour from a terminal perfectly able to show it.

`go.mod` promotes `charmbracelet/x/ansi` and `spf13/pflag` from `// indirect` to direct. `vendor/modules.txt` already marked both `## explicit`, so go.mod was the stale file.

## Still open

No `Cancelled` stage. `waiter.wait` takes no context and sleeps with a bare `time.Sleep`, so Ctrl-C kills the process before the deferred `Stop()` runs and a terminal keeps a half-painted frame. Wiring a context through `wait` is a self-contained follow-up; `commands/apps.go` already has the `signal.NotifyContext` pattern to copy.

Defaulting `--wait` on for interactive terminals is worth doing but is deliberately not here — this PR leaves every command's default exactly where it was, so a change in when commands return is reviewable on its own. Discussion in the thread below.

`kubernetes cluster create` warns and continues when its wait fails, so a create that times out can still exit 0. That predates this PR in shape, but `--wait-timeout` gives it a new way to happen, so it should be settled before this merges.

## Test plan

- [x] `go test ./internal/ui/... ./commands/` — passes. New `spinner_test.go` covers the plain path: no escapes, no glyphs, a line per stage change, and dedupe of an unchanged stage. `wait_test.go` covers immediate first poll, poll error, timeout, and that progress never reaches stdout.
- [x] `go test ./integration/` — all wait-related tests pass. Failure set is identical to the pre-change baseline (registry login and auth init need the macOS Keychain, gradient tests need network).
- [x] Integration expectations updated to the activity lines, e.g. `Running action (10101)` and `Deploying app (...) (0 of 1 steps complete)`.
- [x] Manual check on a real terminal of `droplet create --wait`, which is what turned up the truncated table values and the over-long progress line fixed in this branch.


SSharma-10:
Manual testing of `compute droplet create --wait` on a TTY showed that the boxed table truncates values like the ID, name, and IP addresses (e.g. `5…`, `167.71.255…`). This makes the output harder to copy and use directly.

We should find a way to display these values in full. Truncating important resource details makes the output less user-friendly

SSharma-10:
the waiting line is quite long: 
`Waiting for Droplet (wait-test2) to become active (0 of 1 active) (1s)`.

Could we shorten it, something like:
`⠹ Creating droplet… (12s)` → `✓ Droplet ready (35s)`
and keep stage detail (`0 of 1 active`, `provisioning`) only when it adds signal?

SSharma-10:
IMO progress only appears with `--wait`, so interactive users have to explicitly opt into the better UX. Would it make sense to default `--wait` to enabled for interactive TTYs? Otherwise, adding more flags could make the CLI feel more cluttered and confusing

commands/wait.go
```diff
@@ -0,0 +1,216 @@
...
+	spinner := w.env.NewSpinner(op.message())
+	spinner.Start()
+	defer spinner.Stop()
+
+	deadline := time.Now().Add(timeout)
+
+	for {
+		done, detail, err := poll()
+		if err != nil {
+			spinner.Fail("Gave up waiting for %s", op.Subject)
```
SSharma-10, line 169: this path also covers actual failures (errored action, LB, app deploy), not just give-ups. `Failed while waiting for %s` (or similar) would read better. Timeout already has its own `Timed out waiting...` message.

commands/actions.go
```diff
@@ -237,23 +239,45 @@ func RunCmdActionWait(c *CmdConfig) error {
...
+		case "in-progress":
+			return false, a.Status, nil
+		case "errored":
+			return false, "", fmt.Errorf("action (%d) failed", actionID)
+		default:
+			return true, a.Status, nil
+		}
+	})
+	if err != nil {
+		return nil, err
```
SSharma-10, line 279: when an action errors, we currently return nil, err even though the action has already been fetched. Returning action, err would allow callers to surface the failed action details in the output instead of only showing the error message

commands/wait.go
```diff
@@ -0,0 +1,216 @@
...
+		if remaining <= 0 {
+			spinner.Fail("Timed out waiting for %s", op.Subject)
+			return &waitTimeoutError{subject: op.Subject, timeout: timeout}
+		}
+
+		// Poll once more right on the deadline rather than sleeping past it,
+		// so a resource that settles just in time is still noticed.
+		if remaining < interval {
+			interval = remaining
+		}
```
SSharma-10, line 197:
```suggestion
        sleep := interval
		if remaining < sleep {
			sleep = remaining
		}
```
instead of mutating interval for the final partial sleep, could we use a local sleep := interval? This keeps the polling interval unchanged for subsequent iterations and makes it clearer that we are only adjusting the current sleep to stay within the deadline

commands/wait.go
DO-rrao: Agreed, that branch fires on an errored action, load balancer, or app deploy just as often as on anything doctl chose to stop waiting for. Changed it to `Failed while waiting for %s`; the timeout branch keeps its own `Timed out waiting for %s` so the two stay distinguishable. Updated the assertion in TestWaitReturnsPollError to match.

commands/wait.go
DO-rrao: Applied. The clamp writes to a per-iteration `sleep` now, so `interval` keeps its original value. It only ever triggered on the last pass before the deadline, so nothing changes in practice today, but the loop should not depend on that holding.

commands/actions.go
DO-rrao: Done, it returns `action, err` now, so callers can show the action's final status instead of only the error string. It is still nil when the Get itself failed and nothing was ever fetched. Test updated to expect the errored action back rather than nil.

DO-rrao:
> Manual testing of `compute droplet create --wait` on a TTY showed that the boxed table truncates values like the ID, name, and IP addresses (e.g. `5…`, `167.71.255…`).

Fixed in `f3e1e09`, "Show values in full rather than fit a table by cutting them".

A column is now never narrowed past the longest run of characters in its values that has nowhere to break, so IDs, addresses and UUIDs keep their full width. Prose and the comma-separated lists we print for tags, features and volumes are series of short runs, so those columns still yield space to fit the table.

When those floors still leave the table wider than the terminal, it prints at full width without rules, since narrowing only the cuttable columns would not make it fit anyway and box rules wrap into a mess at that width. So a wide table reads as it did before we fitted anything, with every value intact and copy-pasteable, and the box is kept for tables that do fit, which is most of them once `--format` has picked the columns you wanted.

DO-rrao:
> the waiting line is quite long: `Waiting for Droplet (wait-test2) to become active (0 of 1 active) (1s)`. Could we shorten it [...] and keep stage detail only when it adds signal?

Done in `d1aa2f5`, "Say what a wait is doing rather than that it is waiting", close to the shape you sketched:

```
⠹ Creating Droplet (wait-test2) (12s)
✓ Droplet (wait-test2) is active (35s)
```

`waitOp` gained an `Activity`, the short present-tense phrase shown while the operation runs. `Subject` keeps its long form because it is read after the fact and out of context, in the timeout error and the failure line, where the extra words are what make it actionable. Where one helper serves several commands the verb comes from the caller, since only it knows whether a database is being created, forked or migrated.

Detail now appears only when it adds signal, per your point: the Droplet count shows only once there is more than one to count, and an action no longer echoes `in-progress`, which is the only status an in-flight action can report. I kept the resource name in both lines rather than a bare "Droplet ready", since with several droplets in flight or a scrollback full of them it is the only thing identifying which one finished.

DO-rrao:
> progress only appears with `--wait`, so interactive users have to explicitly opt into the better UX. Would it make sense to default `--wait` to enabled for interactive TTYs?

Not changed here, and I would like to keep it out of this PR, but I do not disagree with the direction.

This PR is deliberately mechanical: one waiter, one timeout, one progress renderer, with each command's `--wait` default left exactly where it was. Flipping a default changes when a command returns and what it prints, which is a much easier change to review and revert on its own than buried in a refactor of 40-odd call sites. There is already precedent for it defaulting on: `kubernetes cluster create` has always used `--wait=true`.

Two things worth settling before we do it. First, it is per-command rather than global: waiting on `droplet create` is seconds, on `databases create` it is many minutes, and a shell that blocks that long by default on an interactive terminal is its own kind of surprise. Second, it needs the default resolved at run time from the TTY check rather than at flag registration, so `--wait=false` still wins and non-TTY invocations are untouched; `--wait-timeout` in this PR is what makes that safe to turn on, since a defaulted wait can no longer hang forever.

Happy to take it as a follow-up. Do you want it on everywhere the wait is typically short, or on for the create commands specifically?

SSharma-10 (approved):

PR #1948, opened 2026-09-03
agents proxy: opencode M2 — prompts bridge to the hosted session, answers stream back

julia-ye:
Follows #1942 (M0/M1, merged). Second slice of the opencode facade: the first demo checkpoint — a prompt typed in a real, unmodified `opencode` TUI streams the hosted agent's answer back into the TUI. This also delivers exactly what @SSharma-10's M1 verification flagged: `POST /session` and prompt bridging.

## What's here

- **Prompt bridge**: `POST /session/{id}/prompt_async` concatenates the body's text parts into `SendInput` and tracks the returned per-turn `run_id`. The synchronous `POST /session/{id}/message` variant is accepted as fire-and-forget (logged; current TUIs use `prompt_async`). Non-text parts are logged and dropped.
- **Event loop**: the `/global/event` handler drains `StreamSession` and translates canonical events into the opencode frame sequence. Single writer by construction (the SSE handler goroutine), so no notifier mutex — a comment marks M5's HITL work as the point where that changes.
- **Session synthesis**: `POST /session` / `GET /session` / `GET /session/{id}` map every client-side session onto the one bridged hosted session.

## The translation was captured, not guessed — and the capture mattered

A first version built only from the OHR adapter's inverse mapping tables bridged the turn correctly (the hosted run executed and completed) but rendered **nothing** in the TUI. Ground truth from a real `opencode serve` turn (logging relay + `curl -N /global/event`) surfaced three load-bearing requirements, each confirmed against the real client:

1. **A part must be created before its first delta** — `message.part.updated` with empty text, then deltas. A delta for an unannounced part silently renders nothing.
2. **Assistant `message.updated` info must carry `cost` and `tokens{...}`** — without `tokens.output` the 1.18.25 TUI crashes (`undefined is not an object`). Zeros are fine; absent is fatal.
3. **The turn must close with an assistant `message.updated` carrying `time.completed`** — otherwise the TUI leaves the turn looking in-flight.

Full verified sequence: user echo (message + prompt part) → `session.status busy` → assistant message → part-create → delta trail → finalized part → completed message → `session.status idle` → `session.idle`. Reasoning tokens stream as their own part with `field:"reasoning"`.

## Testing

- Unit: the full frame sequence (order + ids + per-frame fields) against the shared `agentproxytest` harness; failure path (`run.failed` → `session.error` → idle); untracked-run skipping; prompt validation; session list lifecycle.
- Live: against the local dev stack, a real `opencode attach` + typed prompt rendered the hosted agent's streamed answer (verified by replaying the pty capture through a terminal emulator), with no unhandled routes during the turn.

## Not in this PR

History on re-attach (M3), tool-call parts and real usage/cost in `step-finish` (M4), permission round-trips (M5), stream reconnect (M6).

### DIFF
--- a/internal/agentproxy/opencode/bridge.go
+++ b/internal/agentproxy/opencode/bridge.go
@@ -1,0 +1,463 @@
+package opencode
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"log"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/digitalocean/godo"
+)
+
+func timeNowMs() int64 { return time.Now().UnixMilli() }
+
+// M2: the harness bridge. Client prompts go to SendInput; the hosted
+// session's canonical event stream comes back as opencode-native SSE frames.
+// The translation tables are the inverse of plano's opencode adapter
+// (crates/agent_adapters/src/opencode/raw.rs) — when unsure how a canonical
+// event should look in opencode terms, that adapter is the answer key.
+
+// turnState tracks one in-flight hosted run and the opencode ids synthesized
+// for it. Created when SendInput returns the run id; deleted on
+// run.completed/run.failed. Events for runs this facade isn't tracking are
+// skipped, same as the codex facade — they belong to another client's turn
+// or predate this connection (M3's history endpoint covers the latter).
+type turnState struct {
+	// userMsgID/promptText echo the user's message back on the stream when
+	// the turn starts — the real server does this (message.updated role
+	// "user" + a text part carrying the prompt), and the TUI renders the
+	// user's side of the conversation from those frames, not optimistically.
+	// userMsgID is the client-minted id from the prompt body when present.
+	userMsgID  string
+	promptText string
+	// startedSent: the turn-start frames (user echo, busy, assistant
+	// message.updated) went out. Normally set by run.started, but set lazily
+	// by the first delta too — a turn whose run.started this stream never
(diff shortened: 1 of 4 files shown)


internal/agentproxy/opencode/bridge.go
```diff
@@ -0,0 +1,542 @@
...
+		if err := ew.session("session.error", map[string]any{
+			"sessionID": sid,
+			"error": map[string]any{
+				"name": "UnknownError",
+				"data": map[string]any{"message": payload.Message},
+			},
+		}); err != nil {
+			return err
+		}
+		return f.emitIdle(sid, ew)
```
SSharma-10, line 400: opencode TUI uses time.completed to know “this message is done.” Without it, the UI can keep treating the assistant reply like it’s still running (spinner / QUEUED), even though the run already failed. Worth calling finishAssistantMessage on the failed path too

julia-ye: Done in ce005d96 — `run.failed` now closes the assistant message (`time.completed`) before `session.error`, so a failed reply no longer renders in-flight. One nuance propagated up the stack: from M4 on, `finishAssistantMessage` stamps `finish:"stop"`, which a failed turn shouldn't claim — the M4 branch adds a `failed` flag so failed turns get `time.completed` but no finish reason, matching the history reconstruction's failed-turn shape. Test updated to pin both. (Also picked up Copilot's three notes in the same commit: bounded turn tracking, log-prefix consistency, and scanner-error surfacing in the test helper.)

PR #1949, opened 2026-09-03
agents proxy: opencode M3 — history on re-attach

julia-ye:
**Stacked on #1948** (M2 — review that first; this diff is M3 only). Third slice of the opencode facade: re-attaching shows prior turns.

## What's here

- **`GET /session/{id}/message`** served from a one-shot `replay_only` `StreamSession` pass over the session's durable event history, reconstructed into the `[{info, parts}]` shape captured from a real server. opencode replays history **client-side** — `attach` fetches this endpoint (`?limit=100` observed live, honored) — so nothing is pushed on connect and codex's `--replay` flag machinery has no opencode equivalent.
- **Session list gating**: the bridged session appears in `GET /session` when it has replayable history (or was used through this proxy), which is what `opencode attach --continue` and the session picker resume through a freshly started proxy. A truly fresh session still lists empty (the M0-captured fresh-attach state).
- **`GET /session/{id}/diff` and `/todo` stubs** (empty arrays) — the TUI's post-turn and resume refresh lookups, previously 404s.

## Found while verifying live

- `run.started`'s prompt text arrives on the SSE wire as `data.agent`, not the proto's `user_input` — both keys are accepted (with the proto name preferred), so history reconstructs the user side of each turn.
- `time.completed` must come from the `run.completed` event, not the run's max event time: late stragglers (`run.log` after completion) inflated a 2-second turn into a "20m 1s" duration badge.
- History ids are minted with the id time-encoding from #1948's ordering fix, with **monotonic T allocation across turns** — two turns whose events share a millisecond would otherwise interleave in the TUI (caught by a test before it shipped).

## Testing

- Unit: two-turn reconstruction (roles, parts, parentID linkage, finish/time.completed, reasoning-as-own-part, chronological id ordering across the conversation), `limit` handling, fresh-session emptiness, failed-turn shape, history-gated session listing.
- Live: a fresh proxy against a session with six prior turns + `opencode attach --continue` rendered all six with correct prompts, answers, order, and durations.

## Not in this PR

Tool-call parts and real token/cost numbers in history and live turns (M4), permission round-trips (M5), stream reconnect (M6).


internal/agentproxy/opencode/history.go
```diff
@@ -0,0 +1,272 @@
...
+	"github.com/digitalocean/godo"
+)
+
+// M3: history. `opencode attach` replays a session by fetching
+// GET /session/{id}/message?limit=N itself (client-driven — the server never
+// pushes history), so re-attaching shows prior turns. The facade serves that
+// endpoint from a one-shot replay_only StreamSession pass over the session's
+// durable event history, reconstructed into the [{info, parts}] shape
+// captured from a real server (see the capture doc). No cache: history is
+// fetched per request, and the TUI asks once per attach.
```
SSharma-10, line 19: this header is stale after the cache commit, it still says "No cache: history is fetched per request", but history() below is cached + single-flight and invalidated on turn end. Worth updating so the file doc matches the behavior

internal/agentproxy/opencode/history.go
```diff
@@ -0,0 +1,272 @@
...
+// The cache is warmed at the first request the facade sees (the TUI's
+// /global/health preflight fires well before the attach burst) and
+// invalidated when a live turn completes (invalidateHistory).
+func (f *Facade) history(ctx context.Context) ([]historyMessage, error) {
+	f.histMu.Lock()
+	defer f.histMu.Unlock()
+	if f.histValid {
+		return f.hist, nil
+	}
+	msgs, err := f.fetchHistory(ctx)
```
SSharma-10, line 69: history() holds histMu for the whole replay (~8s). invalidateHistory() needs that same lock, so if a turn completes while history is still loading, the event loop can wait on that lock until the replay finishes

julia-ye: Fixed in 83ede4b9 — the fetch now runs outside `histMu`. Single-flight is a `histFetching` flag + done channel (waiters honor their ctx), and a `histGen` counter catches an invalidation that overlapped a fetch: the overlapped result is still returned to its caller (valid point-in-time read) but not cached, so the next read replays fresh. `invalidateHistory()` is now only ever a fast lock. Regression test holds a replay mid-stream (via the harness's `WaitForHITL` gate — `HangStreamAfterEvents` exempts replay_only streams) and asserts invalidation returns immediately and the superseded fetch doesn't populate the cache.

internal/agentproxy/opencode/history.go
julia-ye: Updated in the same commit — the header now says replays are cached for the proxy's lifetime (fetched once, single-flighted, warmed at first request, invalidated on turn completion), and the `history()` doc comment describes the outside-the-lock design.

SSharma-10 (approved):

Issue #996, opened 2021-06-22
"doctl auth switch" mangles all auth-contexts keys that are email addresses


andygeorge:
### Describe the Issue:

With one or more auth contexts that are email addresses (eg `test@example.com`), using `doctl auth switch` mangles ALL `auth-contexts` keys in the config file: it removes the TLD substring (eg `com`) from each original email address key, and makes it a "sub key" under the now-mangled email address key:

A `~/.config/doctl/config.yaml` with working contexts/API keys:
```
auth-contexts:
  test@example.com: my_secret_api_key
  test2@example.org: another_secret_api_key
```

The same `~/.config/doctl/config.yaml`, after a `doctl auth switch`:
```
auth-contexts:
  test@example:
    com: my_secret_api_key
  test2@example:
    org: another_secret_api_key
```
----
### Steps to reproduce:

1. Just showing I am doing this from scratch:
```
$ rm -rf ~/.config/doctl/config.yaml

$ doctl auth list
default (current)
```

2. Run a `doctl auth init` and verify the new context:
```
$ doctl auth init --context "test@example.com"
Please authenticate doctl for use with your DigitalOcean account. You can generate a token in the control panel at https://cloud.digitalocean.com/account/api/tokens

Enter your access token:
Validating token... OK

$ doctl auth list
test@example.com
default (current)
```

3. Run a `doctl auth switch` (with or without arguments; this will mangle ALL auth contexts that are email addresses)
```
$ doctl auth switch
Now using context [default] by default

$ doctl auth list
test@example
default (current)
```

----

I assume this is happening [somewhere around here](https://github.com/digitalocean/doctl/blob/00cddfcca309bdb8c5fb46b5399218ceae39a28b/commands/auth.go#L234), so not 100% sure it's a `doctl` thing, or upstream `gopkg.in/yaml.v2` / `spf13/viper` issues.

----

**Environment:**

 - doctl version: 1.61.0-release
 - OS: Fedora 33
 - Installation method: GitHub release for Linux

**Additional Details:**

PR #1954, opened 2026-09-08
Greet the user on a bare `doctl`, and stop auth commands rewriting the config file

DO-rrao:
Replaces the command listing that a bare `doctl` printed with a short greeting, and fixes the auth and configuration problems that surfaced while building it. The most serious of those: `auth init`, `auth switch`, and `auth remove` were writing doctl's fully resolved settings over the user's config file, so a token that existed only in the environment was persisted to disk by commands that had nothing to do with it.

### Greet the user on a bare `doctl`

`doctl` with no arguments printed the full command listing, which buries the two things a newcomer needs: whether their credentials work, and where to start. It now prints a short greeting with the running version, the active account and team, the current context, a few first commands, and an upgrade notice when one is available.

Auth is reported as four distinct states rather than a boolean, because "configured" and "working" are different claims. In particular an unreachable API is reported as `token could not be verified` rather than as a bad token, so an offline user is not sent off to rotate a perfectly good credential. Only a 401 produces `token rejected by the API`.

Both network lookups run concurrently and are individually time-bounded, so the greeting costs the slower of the two rather than their sum, and never blocks on either. A machine-readable form is available with `-o json`.

### Stop writing resolved settings back over the user's config file

`auth init`, `auth switch`, and `auth remove` each ended by serialising `viper.AllSettings()` over the config file. That is doctl's merged view of built-in defaults, the file, `DIGITALOCEAN_*` variables, and flags, and viper deliberately forgets which layer a value came from, so writing it back persisted whatever happened to be set during that invocation.

Measured on the same 5-line config with `DIGITALOCEAN_ACCESS_TOKEN` exported, running `auth switch --context work`:

| | before | after |
| --- | --- | --- |
| lines written | 2558 | 4 |
| environment token present | yes, line 10 | no |
| saved `access-token` | overwritten with the environment's | left as it was |

The third row is its own bug: because viper resolves `access-token` to the environment's value, the dump did not merely leak the CI token, it destroyed the token the user had saved.

These commands now load the YAML, apply the change they were actually asked to make, and write it back. Keys doctl does not own, including ones this version knows nothing about, are returned unchanged. `auth init` still saves the token it just validated whatever its source, because saving a token is the whole job of that command; the bug was the other two doing it as a side effect.

Reading the YAML rather than viper's copy also retires the workaround in `auth switch` whose comment admitted it "somehow magically fixes an issue" (#996). The cause was that viper lowercases keys and splits them on `.`, so a context named `test@example.com` came back as a nested map. Nothing splits it now.

**Worth calling out for the release notes:** anyone who has run doctl in CI with `DIGITALOCEAN_ACCESS_TOKEN` set may already have that token sitting in their `config.yaml`. It is worth checking, and rotating if so.

### Correct the auth command descriptions that contradict the code

`scripts/gen-yaml-docs.go` runs cobra's `GenYaml` over the whole command tree, so every `Short`, `Long`, and `Example` is published as documentation. Several were wrong.

`auth init` instructed the reader to "Provide a case-sensitive name for the context". Names are the opposite of case-sensitive: they are lowercased in `initConfig` and again in `RunAuthInit` and `RunAuthSwitch`, and contexts are matched with `strings.EqualFold`. It also described `auth switch` as re-initializing doctl, which invites the reader to expect a token to be revalidated; switching writes the `context` key and nothing else.

Two behaviors were undocumented. Adding a context does not begin using it, so a reader following the example was left on the context they started from wondering why their new token was not in use. And the page promised a prompt twice while saying nothing about `--access-token` or `DIGITALOCEAN_ACCESS_TOKEN`, either of which suppresses that prompt and replaces the saved token. The claim that `--access-token` can be added "to each command" is also true only of the `default` context; `README.md` documented that limitation but the help did not, and the help is what you read when the flag appears to do nothing.

The rest is proofreading: "requires and API token", "the API token you want use", "a just one account", and a `Short` that ended in a space.

### Say when the active token came from outside the config file

Nothing told you which token was in play. With a config holding a token for `saved@example.com` and `DIGITALOCEAN_ACCESS_TOKEN` set to one for `ci-robot@example.com`, `auth token` printed the environment's token with no indication, `auth list` was byte-identical either way, and the new greeting reported a different account *and* a different team while the `Context` row still said `default`.

The greeting now names the source when the token did not come from the config file:

```
  Account   ✔ saved@example.com                                         ← saved token, unannotated
  Account   ✔ ci-robot@example.com (from DIGITALOCEAN_ACCESS_TOKEN)
  Account   ✔ flag-user@example.com (from --access-token)
  Account   ✘ token rejected by the API (from DIGITALOCEAN_ACCESS_TOKEN)
```

The last line is the one that earns the change: a bad ambient token used to be indistinguishable from a bad saved one. A named context is never annotated, since the flag and the variable cannot reach it, and the config file is left unannotated on the same reasoning the `Team` row uses — a row that is always present stops being read. Provenance is established by comparing the resolved token against each layer in viper's order of precedence, because viper merges them and forgets which one won. `-o json` always carries `tokenSource`.

`doctl auth --help` and `README.md` now also state the rule this PR establishes: an ambient token outranks the saved one, applies only to the command you run, and is never written to the file, with `auth init` as the deliberate exception.

## Testing

Unit tests cover the new configuration seam, provenance resolution, and every greeting state. Tests gained a config file seam alongside the existing writer: without one, the rewritten assertions would read the config of whoever ran them, and the previous `TestAuthInitConfig` asserted against the dump and so passed or failed depending on whether that developer had ever run `doctl auth init`.

Each user-facing claim in the new help text and README was verified against a built binary rather than against the source.

`TestRegistryLogin` and `TestRegistryLogout` fail on a macOS workstation because they reach the real keychain. That is pre-existing and unrelated.

## Follow-ups not in this PR

- `defaultConfigFileWriter` still uses `os.Create`, which opens `0666` before the `chmod` to `0600`, and truncates before the new bytes land. Now that the file is essentially only credentials, `os.OpenFile(..., 0600)` plus a temp-file-and-rename would be worth it, the latter also making concurrent invocations safe.
- `auth init` with `DIGITALOCEAN_ACCESS_TOKEN` set still replaces a stored token silently. It is now documented, but a warning would close the last path by which a CI token lands on disk.
- When the token is invalid, the greeting still suggests `doctl auth init`. If the bad token came from the environment, `unset DIGITALOCEAN_ACCESS_TOKEN` is the actual fix, so the suggested steps could be provenance-aware.
- Comments and key order in the config file do not survive a rewrite, since the document is remarshalled with `yaml.v2`. That is not a regression — the old dump destroyed them too — but the README now tells users to hand-edit the file, and `yaml.v3`'s `yaml.Node` round-trips both.


SSharma-10:
context is out of scope here and is breaking the package build. Looks like the switch logic got added to writeConfig during the rebase. Please fix the compile issue and align this with the file-based config path

SSharma-10 (approved):
LGTM

PR #1957, opened 2026-09-09
agents: add `harness-runtime prompt` for headless single-prompt runs

Spherrrical:
Sending one prompt to a session and reading the answer back currently takes an interactive attach, which a script cannot drive. This adds a top-level verb that does it in one non-interactive call: send, wait for the run to finish, print the answer.

```
doctl harness-runtime prompt <session> <prompt>...   (alias: ask)
```

The session comes first, matching every other session command in the tree (`show`, `exec`, `upload`, …). The prompt is variadic, so an unquoted sentence still works, and `-` reads it from stdin.

## The stream split is the contract

stdout carries the agent's answer alone, so it can be captured directly:

```
ANSWER=$(doctl harness-runtime prompt my-session "What is the capital of France?")
# Paris
```

Progress, tool calls, and the closing token/cost summary go to stderr, where they stay visible on a terminal without corrupting a pipe. `--include-reasoning` also emits the model's thinking (to stderr — reasoning is not the answer and never reaches stdout), `--quiet` silences stderr, and `-o json` emits the answer, run ID, status, and usage as one object.

The exit status reports the run — 0 completed, 1 failed, 124 timed out (GNU `timeout`'s status) — so this composes in `&&` chains and `if` tests. A timeout stops waiting locally and does not stop the agent; the run continues and its output stays readable with `logs`.

## Two ordering details worth review attention

**Subscribe before send.** The event stream is opened before `SendInput`, because `DoStream` returns once the connection is live. Sending first would race the subscription: the run's opening events could land before the stream exists, and the command would then wait out its entire timeout for a terminal event it had already missed.

**Filter to this run.** Every event is matched against the run ID `SendInput` returned, so a session driven from elsewhere cannot contaminate the answer or end the wait early. Events are also deduplicated by event ID, since a reconnect replays from the cursor and the boundary event can arrive twice — which would otherwise duplicate a chunk of the answer.

Both are pinned by tests.

## Reconnect logic is shared, not copied

The reconnect loop is extracted out of `watchSessionHeadless` into `runHeadlessStream` and both callers now use it. That keeps one copy of the subtle rule that a stream which stayed up before dropping is a server idle timeout rather than a fault and must not consume the failure budget — without it, an agent that merely thinks for a few minutes gets reported as a lost stream. This is why `agents_hitl_watch.go` shrinks by 67 lines; its behaviour is unchanged.

## Headless means nobody to ask

An approval request stops the command unless `--on-hitl approve|reject|defer` says how to answer, and the error names both the flag and the request ID so it can be resolved out of band. Blocking would hang until the timeout and look identical to a slow agent; silently defaulting to approve would run a command the caller never saw.

A paused session is resumed and waited to `READY` before the prompt is sent, since sending to a session that is awake but not yet listening would drop it.

## Testing

15 new tests covering the stdout/stderr split, unquoted and stdin prompts, subscribe-before-send ordering, run-ID filtering, reasoning exclusion, both approval paths, failed/timed-out/incomplete exit codes, paused-session resume, and JSON output. `gofmt` and `go vet` are clean and the full `commands` suite shows no regressions.

`TestRegistryLogin`/`TestRegistryLogout` fail on this machine, but they fail identically on a clean checkout of the base branch and are unrelated to this change.


SSharma-10 (approved):
LGTM

PR #1960, opened 2026-09-15
fix(MARSOHS-1404): validate egress field spelling per manifest format

sanpj2292:
## Summary

- Make `validateEgressObject` format-aware: flat manifests accept only snake_case keys (`allow_ips`, `vpc_uuid`, …); legacy envelope accepts only camelCase keys (`allowIps`, `vpcUuid`, …). Wrong-format spellings now produce clear errors naming the correct key.
- Remove the client-side copy of the "no host allowlist is set" IP-only egress advisory. That wording is harness-api's `session.warnings` content only (via `computeWarnings` + `flatPathWarnings`), printed after create — fixes the duplicate warning with mismatched field names (`egress.allowIps` vs `egress.allow_ips`).

## Test plan

- [x] `go test ./commands/... -run 'TestValidateAgentManifest_Egress' -v` — all pass
- [x] `go build .` — main binary builds
- [ ] Manual: `doctl harness-runtime create -f j3-ips.yaml --name j3-ips-test` against dev harness-api — expect exactly one `Warning egress.allow_ips` (server, after create only)


commands/agents_validate.go
```diff
@@ -422,16 +422,68 @@ func validateManifestSkills(raw any, path string, out *agentManifestValidation)
...
+	flatKey   string
+	legacyKey string
+}
+
+var egressAliasFields = []egressAliasField{
+	{flatKey: "allow_hosts", legacyKey: "allow"},
+	{flatKey: "allow_ips", legacyKey: "allowIps"},
+	{flatKey: "vpc_uuid", legacyKey: "vpcUuid"},
+	{flatKey: "subnet_uuid", legacyKey: "subnetUuid"},
+}
```
sdharavath, line 446: do we have any other fields that fall in this aliases for egress ?

commands/agents_validate.go
```diff
@@ -422,16 +422,68 @@ func validateManifestSkills(raw any, path string, out *agentManifestValidation)
...
+
+// egressWrongFormatHint reports whether k is the OTHER format's spelling for
+// a known egress concept, so the error can name the correct key instead of a
+// bare "unknown field". A key matching neither format (e.g. the fictitious
+// "allowHosts") falls through to the generic unknown-field error.
+func egressWrongFormatHint(k string, legacy bool) (hint string, matched bool) {
+	for _, f := range egressAliasFields {
+		right, wrong := f.flatKey, f.legacyKey
+		wrongFormatName, thisFormatName := "legacy envelope", "flat"
```
sdharavath: Can we have this moved to the const section

commands/agents_validate.go
sanpj2292: What do you mean by other fields ?

sdharavath: I mean that we have fields like allow_hosts, ips, uuids, apart from these there are any other fields, if not then we are good. Hope this clarifies

sanpj2292: Yes, that clarifies and we’re good.

commands/agents_validate.go
sanpj2292: This is done.

sdharavath (approved):
lgtm

SSharma-10 (approved):

PR #1961, opened 2026-09-15
Structure errors as Title/Reason/Status/Next step, name exit codes

DO-rrao:
## Summary
- `checkErr` now resolves every error to a `StructuredError` (an explicit implementation, or a `genericStructuredError` synthesized from a new status-code table) and renders **Title → Reason → Status → Next step** as separate lines, mirroring the same four fields into the JSON error envelope additively (`detail` unchanged for existing consumers).
- Adds `commands/error_codes.go`: canned reason + a `run <command>` next step for common API status codes (400/401/403/404/409/422/429/5xx), so `Style.Hint` bolds the whole suggestion consistently.
- `FlagValidationError` implements the new interface for its JSON output while keeping its own richer per-flag text block untouched.
- Names the two exit codes doctl has always used (`exitGeneralError = 1`, `exitUsageError = 255`) instead of magic numbers, and routes the config-read failure through `checkErr` instead of a raw `log.Fatalln`, so it gets consistent formatting and a controlled exit code.
- Fixes a pre-existing compile bug in `auth.go`'s `writeConfig()`, which referenced an out-of-scope `context` identifier; it's now a parameter from its one call site.

## Test plan
- [x] `go build ./cmd/doctl` succeeds
- [x] `go vet ./commands/... ./internal/...` clean
- [x] `go test ./commands/...` — existing `checkErr`/`FlagValidationError` tests pass unchanged (one pre-existing, unrelated failure: `TestRenderWelcomeUsesASCIIGlyphsWhenRequired`, confirmed to fail identically without these changes)
- [x] Manually verified: 404/text, 404/JSON, missing-flags/text, missing-flags/JSON, unknown command (exit 255), success path (exit 0), corrupt config file (exit 1, styled instead of raw crash), forced-PTY color/bold rendering


commands/auth.go
```diff
@@ -344,10 +344,10 @@ func RunAuthSwitch(c *CmdConfig) error {
 	viper.Set("context", context)
 
 	notice("Now using context [%s] by default", context)
-	return writeConfig()
+	return writeConfig(context)
```
SSharma-10: We are printing the success message twice: `notice` above and `fmt.Printf` in `writeConfig` L362

SSharma-10:
Build looks good now, but the auth switch is still broken.

Locally, `go test ./commands/` is failing with `context does not exist` in:

* `TestAuthSwitchPersistsOnlyTheContext`
* `TestAuthForcesLowercase`

Looks like `RunAuthSwitch` is still checking viper while `writeConfig` checks the YAML file, so they are out of sync. Can you wire the switch through the file-based path as well?

SSharma-10:
Tried this locally:

  doctl compute droplet get 999999999

Text only shows "Not Found" + the canned reason, the actual API message only shows up in json `detail`.
Also for a 404, suggesting `doctl auth list` feels off

commands/errors.go
```diff
@@ -38,14 +53,144 @@ var (
...
+	return defaultNextStep()
+}
+
+// defaultNextStep suggests the active command's help text. Empty if no
+// command is active (e.g. errors raised during early config bootstrap).
+func defaultNextStep() string {
+	if activeCommand == nil {
+		return ""
+	}
+	return fmt.Sprintf("run %s --help", activeCommand.CommandPath())
```
SSharma-10, line 170: Defaulting every error to `→ run <cmd> --help` feels too broad. When the error already tells the user what to do, it becomes redundant or conflicting. For example:

`Error: Unable to initialize DigitalOcean API client: access token is required. (hint: run 'doctl auth init')`
`→ run doctl compute droplet get --help`

Same for confirm-aborts, which already tell the user to use `--force`.

Can we suppress the default next step when the error already includes guidance, or only show `--help` for validation/usage-style errors?

commands/error_codes.go
```diff
@@ -0,0 +1,137 @@
...
+	},
+	http.StatusNotFound: {
+		// No canned next step: a bad ID and a wrong auth context both surface
+		// as 404, and the ID typo is by far the more common of the two, so
+		// there is no single suggestion that fits most of the time.
+		Reason: "the requested resource does not exist, or not in this account/context",
+	},
+	http.StatusConflict: {
+		Reason:   "the resource is in a state that conflicts with this request",
+		NextStep: "run %s",
```
SSharma-10, line 65: `NextStep: "run %s"` for 409/5xx tells users to re-run the same command that just failed. That feels misleading for conflicts and transient server errors.

Can we use a wait/retry hint, or `--help` instead of telling users to run the same command again?

commands/errors.go
```diff
@@ -38,14 +53,144 @@ var (
...
+	}
+	return ""
+}
+
+// Reason prefers the message the API itself returned, since it is specific
+// to the request that failed; the status-code table's line is a fallback for
+// when the API had nothing more to say than the status code.
+func (e genericStructuredError) Reason() string {
+	if gerr, ok := apiError(e.err); ok && gerr.Message != "" {
+		return gerr.Message
```
SSharma-10, line 125: Using godo.ErrorResponse.Message is good, but it drops useful details from Error(), like giving up after N attempt(s).

This makes exhausted-retry and retries-disabled look identical in text. Can we keep that annotation in the Reason/Title or elsewhere in the text output, not just JSON detail?

commands/errors.go
DO-rrao: Instead of checking specific error types (which broke tests since many local errors are just fmt.Errorf), the logic now simply checks if the error reached the API. If we got an HTTP status or network error, it skips --help. If it's a local error, --help is shown.

I also fixed a few related issues I found while testing:

1. Empty next-steps now correctly print nothing instead of falling back to --help.
2. AskForConfirmDelete no longer swallows the actual error for a generic abort message.
3. Fixed commands/auth.go masking godo errors with %s instead of %w, which was causing them to lose their metadata and hints.
4. 404 errors with no canned next step no longer incorrectly fall back to --help.

commands/error_codes.go
DO-rrao: Agreed, the run %s hint for those errors was misleading. Fixed in ee86227e.

For 409s: The hint is dropped entirely since state conflicts usually require manual intervention.

For 5xxs: The hint is now suppressed once retries are exhausted (which happens by default). It will only show run <cmd> if the user explicitly disables retries (--http-retry-max 0).

Neither of these will fall back to --help anymore.

commands/errors.go
DO-rrao: The attempt count is now included in the Reason line. I pulled this directly from the gerr.Attempts struct field rather than parsing the error string, so it won't break if the error format changes. This update also carries over to the JSON output, making the reason field much more informative.

SSharma-10 (approved):

PR #1964, opened 2026-09-15
agents: add --resume-on-topoff for auto-resume after a low-balance pause

Spherrrical:
## What

harness-api's `CreateSession` grew a `resume_on_topoff` opt-in: a session that consents is resumed automatically once the team's prepayment balance is restored after a low-balance pause. doctl had no way to send it, so the option was unreachable from the CLI.

This exposes it as `--resume-on-topoff` on every command that creates a session: `create`, `launch` (create-then-attach), and `config start-session`.

## How it reaches the server

The API takes it two ways, and doctl matches each:

| Create path | How it travels |
| --- | --- |
| Manifest (YAML body) | `?resume_on_topoff=true` query parameter |
| Agent Config (JSON body) | `resume_on_topoff` body field |

It is deliberately **not** an `agents.yaml` field. Per the swagger spec, that document is immutable and shared through Agent Configs, while spending consent has to stay per-session and revocable. So the value travels beside the manifest rather than inside it.

## Decisions worth reviewing

**`--dry-run` says the flag isn't in the YAML.** The printed manifest cannot carry the consent, so piping it into `config create` — a pipeline the help text documents — would silently drop it. A note on stderr makes that explicit while stdout stays pipeable.

**Withheld consent sends nothing** rather than `resume_on_topoff=false`. The server treats the two identically, and `omitempty` keeps the request clean.

**`launch <existing-session>` rejects the flag** instead of ignoring it. The API sets this consent at create time only (there is no session-update surface yet), so accepting it while attaching would leave the user believing an existing session had been enrolled.

A granted consent is echoed back on the show card and in `-o json`.

## godo

The three type changes live in `vendor/github.com/digitalocean/godo/hosted_agents.go` (query options struct, config request body, session response), following this branch's existing practice for hosted-agents work. **These still need upstreaming to godo before this merges.**

## Testing

New `commands/agents_topoff_test.go` covers both wire forms, the wire-name tags, the `launch` rejection, and the show card.

Also verified the built binary against a stub server:

- manifest path with the flag: `POST /v2/agents/sessions?resume_on_topoff=true`, YAML body untouched
- without the flag: no query parameter at all
- config path: `{"name":...,"config_id":...,"resume_on_topoff":true}`

The agent suites pass. The only failures in `go test ./commands/...` are pre-existing macOS keychain errors in the registry login tests, unrelated to this change.


SSharma-10 (approved):
Verified after the vendor dedupe: builds and Topoff/ResumeOnTopoff tests pass. LGTM.

PR #1965, opened 2026-09-15
prepayment: add config and status commands

rodrigoea:
## Summary

Adds doctl commands for the public prepayment endpoints:

- `doctl prepayment config get`
- `doctl prepayment status get`

These wrap:

- `GET /v2/customers/my/prepayment_config`
- `GET /v2/customers/my/prepayment_status`

Requires `billing:read`. Endpoints are already live in production.

---

https://do-internal.atlassian.net/browse/BILL-16464


commands/displayers/prepayment.go
```diff
@@ -0,0 +1,125 @@
+/*
+Copyright 2018 The Doctl Authors All rights reserved.
```
DO-rrao, line 2: Nit: Created At/Updated At render as 0001-01-01T00:00:00Z when no prepayment config exists

Tested against prod on an account that's eligible: true for the prepayment gate but hasn't configured it yet (spend_limit, prepay_amount, etc. all zero). Output:

```
Spend Limit    Auto Prepay Enabled    Prepay Amount    Prepay Threshold    Balance    Blocked    Eligible    Month-to-date Balance    Created At              Updated At
0.00           false                  0.00             0.00                0.00       false      true        0.00                     0001-01-01T00:00:00Z    0001-01-01T00:00:00Z

```
Since godo.PrepaymentConfig.CreatedAt/UpdatedAt are non-pointer time.Time, an absent/null value from the API unmarshals to Go's zero time, and commands/displayers/prepayment.go unconditionally formats it:

```
x["CreatedAt"] = p.Config.CreatedAt.Format(time.RFC3339)
x["UpdatedAt"] = p.Config.UpdatedAt.Format(time.RFC3339)

```
The other zero-valued fields (0.00, false) read fine as "not configured," but the date looks like a bug to an end user. Suggest guarding on IsZero() so the column is blank instead, e.g.:

```
if !p.Config.CreatedAt.IsZero() {
    x["CreatedAt"] = p.Config.CreatedAt.Format(time.RFC3339)
}
if !p.Config.UpdatedAt.IsZero() {
    x["UpdatedAt"] = p.Config.UpdatedAt.Format(time.RFC3339)
}
```

rodrigoea: nice catch @DO-rrao -- I pushed e6083da4d7ed36224a13caa61275be2b18e562d4 with this fix and some additional tests.

DO-rrao (approved):
lgtm

PR #1966, opened 2026-09-16
Mask secrets and database credentials unless --show or CI is set

DO-rrao:
## Summary
- Promote secrets get's local `--show` flag to a global persistent flag, and make masking a first-class capability (`ui.Env.Mask`) resolved once per invocation, alongside `Style`/`Anim`/`DataTTY`.
- Mask secret values (`doctl secrets get`) and database credentials (`doctl databases user get/list/create`, `doctl databases connection get`, `doctl databases replica connection get`, including the password embedded in the connection URI) by default. `--show` reveals them.
- CI (`CI`, `GITHUB_ACTIONS`, `GITLAB_CI`, etc.) is treated as a script that needs the real value, so masking is skipped automatically there without needing `--show`.
- Fold `IsCI()` into the same capability layer that already decided color/TTY-gated layout, so CI also gets the legacy plain, uncolored, uncarded output — even on a runner whose streams would otherwise pass every terminal capability check (verified with a real pty fixture, not just a buffer).

## Test plan
- [x] `go test ./internal/ui/... ./commands/...` — all green except two pre-existing, unrelated failures already present on the base branch (`TestRegistryLogout`, `TestRenderWelcomeUsesASCIIGlyphsWhenRequired`)
- [x] New unit tests: `internal/ui/env_test.go` (`TestDetectMask`, `TestDetectCIForcesPlainOutput` using a real pty via `github.com/creack/pty`)
- [x] New command tests: `commands/secrets_test.go`, `commands/databases_test.go` (mask/show assertions, URI-masking edge cases)
- [x] Manual verification against real secrets/database clusters: default masks, `--show` reveals, `CI=true` reveals and drops color/cards, `-o json` follows the same rule
- [x] `--help` output checked: `--show` appears once as a Global Flag, not duplicated locally on `secrets get`


commands/databases.go
SSharma-10, line 920: The `create` help mentions the auto-generated password but doesn’t mention that it’s masked by default. Could we add the same “Use `--show` to reveal it”

SSharma-10, line 932: Same here: reset returns new credentials, but help doesn't say they'll be masked unless --show is passed.

internal/ui/env.go
```diff
@@ -165,6 +195,11 @@ func Detect(out, err io.Writer, opts ...Option) Env {
 		env.DataWidth = detectWidth(out)
 	}
 
+	// CI is treated as a script that needs the real value, not a screen
+	// someone is watching, so it is exempt from masking. --show overrides
+	// both directions.
+	env.Mask = !cfg.show && !IsCI()
```
SSharma-10, line 201: in CI, secrets/DB passwords are printed in full without --show, so they can end up in CI logs. If that’s the intended tradeoff for scripting, LGTM. Otherwise, we may want to require --show in CI as well

DO-rrao: Yes, this is the intended tradeoff

SSharma-10 (approved):

PR #1973, opened 2026-09-18
Add shared colorful --help layout for all doctl commands

SSharma-10:


commands/doit.go
```diff
@@ -123,6 +123,7 @@ func init() {
 	}
 
 	addCommands()
```
DO-rrao, line 125: writeHelpExample only green-highlights a line if it starts with "doctl " after trimming:

```
if strings.HasPrefix(strings.TrimSpace(trimmed), "doctl ") {
    fmt.Fprintf(b, "%s%s\n", indent, p.successBadge(strings.TrimSpace(trimmed)))
    continue
}
fmt.Fprintf(b, "%s\n", trimmed)
```

This works well for commands whose Example field is a bare invocation (e.g. compute droplet create), but many commands — e.g. [kubernetes.go:386](https://github.com/digitalocean/doctl/blob/main/commands/kubernetes.go#L386) — write Example as a full sentence with the command embedded mid-line: "The following example creates a cluster ...: doctl kubernetes cluster create ...". Since the line doesn't start with doctl , it renders as plain text with no highlight, even though it's a real runnable example.

The result is that --help output looks inconsistent depending on how each command happened to author its Example string — some get the green treatment, most don't.

commands/doit.go
```diff
@@ -123,6 +123,7 @@ func init() {
 	}
```
DO-rrao, line 123: registries delete --help command has text colored starting from example till end of output. ptal

DO-rrao (approved):

PR #1974, opened 2026-09-18
added hermes-base to template base allowList

spinits4:
We will be supporting hermes-base as a base template, but we do not want to expose it to all customers. We want only customers who know about it to be able to use it.


SSharma-10 (approved):
LGTM

